PR #9, opened 2024-07-16
Fix broken links

mattroyer:
The path was incorrect, causing a 404 when trying to visit the internal links.


AlexCheema:
thank you, congrats on the first merged PR (albeit a small one)

PR #18, opened 2024-07-17
fix: syntax error in requirements.txt

andvarfolomeev:


AlexCheema:
Thanks for spotting this!

Issue #41, opened 2024-07-19
Thunderbolt discovery for Macs


AlexCheema:
- should already be supported
- just check that it prioritises thunderbolt over WiFi
- Thanks apple for making thunderbolt usable

![IMG_0094](https://github.com/user-attachments/assets/6b025b99-4bdd-4aa5-bf1c-d6ae2e7fd720)

mintisan:
nice👍

mintisan:
How can I tell if the data is being transferred via Thunderbolt or network WiFi? [If two Macs are connected through Thunderbolt while on the same local area network]

RevREB:
I know is not as simple as this BUT a cli/config file flag to set discovery to use only certain NIC's... either blacklist or whitelist. the auto discovery is cool to enable nodes to dynamically join but this would help to force traffic to only go over the thunderbolt mesh.

Evanev7:
Done!

PR #47, opened 2024-07-20
prioritise thunderbolt over WIFI #41

itsknk:
For issue #41, I figured the solution could be quite simple since it already supports the Thunderbolt discovery and needs to be prioritized.

I've updated the `grpc_discovery.py` file to get this done.

Changes made:
- Import `netifaces` and `platform`.
- Create a `get_network_interfaces` method that prioritizes Thunderbolt interfaces over WiFi on macOS.
- Make the `__init__` method to call `get_network_interfaces` and store the result.
- Update `task_broadcast_presence` to broadcast on all available interfaces, prioritizing Thunderbolt on macOS.
- Make `task_listen_for_peers` to listen on all available interfaces.

### DIFF
--- a/exo/api/chatgpt_api.py
+++ b/exo/api/chatgpt_api.py
@@ -268,9 +268,10 @@ class ChatGPTAPI:
     callback = self.node.on_token.register(callback_id)
 
     if DEBUG >= 2: print(f"Sending prompt from ChatGPT api {request_id=} {shard=} {prompt=} {image_str=}")
-    asyncio.create_task(self.node.process_prompt(shard, prompt, image_str, request_id=request_id))
 
     try:
+      await asyncio.wait_for(self.node.process_prompt(shard, prompt, image_str, request_id=request_id), timeout=self.response_timeout)
+
       if DEBUG >= 2: print(f"Waiting for response to finish. timeout={self.response_timeout}s")
 
       if stream:
@@ -284,9 +285,9 @@ class ChatGPTAPI:
         )
         await response.prepare(request)
 
-        async def stream_result(request_id: str, tokens: List[int], is_finished: bool):
-          prev_last_tokens_len = self.prev_token_lens.get(request_id, 0)
-          self.prev_token_lens[request_id] = max(prev_last_tokens_len, len(tokens))
+        async def stream_result(_request_id: str, tokens: List[int], is_finished: bool):
+          prev_last_tokens_len = self.prev_token_lens.get(_request_id, 0)
+          self.prev_token_lens[_request_id] = max(prev_last_tokens_len, len(tokens))
           new_tokens = tokens[prev_last_tokens_len:]
           finish_reason = None
           eos_token_id = tokenizer.special_tokens_map.get("eos_token_id") if hasattr(tokenizer, "_tokenizer") and isinstance(tokenizer._tokenizer,
@@ -316,7 +317,7 @@ class ChatGPTAPI:
             if DEBUG >= 2: traceback.print_exc()
 
         def on_result(_request_id: str, tokens: List[int], is_finished: bool):
-          self.stream_tasks[request_id] = asyncio.create_task(stream_result(request_id, tokens, is_finished))
+          if _request_id == request_id: self.stream_tasks[_request_id] = asyncio.create_task(stream_result(_request_id, tokens, is_finished))
 
           return _request_id == request_id and is_finished
 
@@ -345,6 +346,9 @@ class ChatGPTAPI:
         return web.json_response(generate_completion(chat_request, tokenizer, prompt, request_id, tokens, stream, finish_reason, "chat.completion"))
     except asyncio.TimeoutError:
       return web.json_response({"detail": "Response generation timed out"}, status=408)
+    except Exception as e:
+      if DEBUG >= 2: traceback.print_exc()
+      return web.json_response({"detail": f"Error processing prompt (see logs with DEBUG>=2): {str(e)}"}, status=500)
     finally:
       deregistered_callback = self.node.on_token.deregister(callback_id)
       if DEBUG >= 2: print(f"Deregister {callback_id=} {deregistered_callback=}")


AlexCheema:
This is awesome! First proper PR, congrats!

Code looks great, only one very minor thing for consistency across the codebase, instead of `platform` use `psutil`. It's the same thing but just more consistent across the codebase.

itsknk:
Will change.

AlexCheema:
Also don't we need to add `netifaces` as a dependency in `setup.py`?

itsknk:
My bad! We do! I just installed it locally and forgot about setup.py! Will update the whole thing and give you a better one!

exo/networking/grpc/grpc_discovery.py
```diff
@@ -34,6 +35,16 @@ def __init__(self, node_id: str, node_port: int, listen_port: int, broadcast_por
         self.broadcast_task = None
         self.listen_task = None
         self.cleanup_task = None
+        self.interfaces = self.get_network_interfaces()
+
+    def get_network_interfaces(self):
+        interfaces = netifaces.interfaces()
+        if psutil.system() == "Darwin":  # macOS
```
AlexCheema, line 42: Make this `if psutil.MACOS:` and remove the comment

AlexCheema:
Just added an [integration test for discovery](https://github.com/exo-explore/exo/actions/runs/10018442725/job/27693811593). Already coming in useful, we'll see if it still passes after your changes.

exo/networking/grpc/grpc_discovery.py
itsknk: Beat me to it!

itsknk:
I guess setting up netifaces dependency failed?!

exo/networking/grpc/grpc_discovery.py
```diff
@@ -114,7 +124,15 @@ async def on_listen_message(self, data, addr):
             self.known_peers[peer_id] = (self.known_peers[peer_id][0], self.known_peers[peer_id][1], time.time())
 
     async def task_listen_for_peers(self):
-        await asyncio.get_event_loop().create_datagram_endpoint(lambda: ListenProtocol(self.on_listen_message), local_addr=('0.0.0.0', self.listen_port))
+        for interface in self.interfaces:
```
AlexCheema, line 127: Does listen need to be interface-by-interface? Can we listen on all interfaces?

exo/networking/grpc/grpc_discovery.py
```diff
@@ -34,6 +35,16 @@ def __init__(self, node_id: str, node_port: int, listen_port: int, broadcast_por
         self.broadcast_task = None
         self.listen_task = None
         self.cleanup_task = None
+        self.interfaces = self.get_network_interfaces()
```
AlexCheema, line 38: If we can remove`self.interfaces` in `task_listen_for_peers` (https://github.com/exo-explore/exo/pull/47/files#r1685535090) then we should move this out of the constructor to `task_broadcas_presence` and have it be error handled (in each loop, we can check if `self.interfaces` is already set and if not, set it)

itsknk: I'll get on with this.

itsknk: - The `task_listen_for_peers` method now listens on all interfaces using '0.0.0.0'.
- The `self.interfaces` attribute is removed from the constructor.
- The `task_broadcast_presence` method now updates the list of interfaces in each iteration, ensuring it always uses the most up-to-date list of network interfaces.

AlexCheema:
I just quickly tested this myself locally. For me, the interface name does not begin with `thunderbolt`.

```python
# before thunderbolt connected
>>> netifaces.interfaces()
['lo0', 'gif0', 'stf0', 'anpi0', 'anpi1', 'anpi2', 'en4', 'en5', 'en6', 'en1', 'en2', 'en3', 'bridge0', 'ap1', 'en0', 'awdl0', 'llw0', 'utun0', 'utun1', 'utun2', 'utun3', 'utun4', 'utun5', 'utun6', 'utun7', 'utun8', 'utun9']
# after thunderbolt connected
>>> netifaces.interfaces()
['lo0', 'gif0', 'stf0', 'anpi0', 'anpi1', 'anpi2', 'en4', 'en5', 'en6', 'en1', 'en2', 'en3', 'bridge0', 'ap1', 'en0', 'awdl0', 'llw0', 'utun0', 'utun1', 'utun2', 'utun3', 'utun4', 'utun5', 'utun6', 'utun7', 'utun8', 'utun9', 'en11', 'en16']

**Comment by AlexCheema** [comment]
Eventually we should support automatically upgrading to the best interface available (in this case thunderbolt) - see #51. We can implement this later though, for now this is great.

**Comment by itsknk** [comment]
So, Thunderbolt interfaces on macOS are also named with the 'en' prefix, just like other network interfaces?! If yes, then does prioritizing all' en' interfaces make sense instead of trying to identify Thunderbolt interfaces specifically? 

Something like:
```python
def get_network_interfaces(self):
    interfaces = netifaces.interfaces()
    if psutil.MACOS:
        en_interfaces = [iface for iface in interfaces if iface.startswith('en')]
        other_interfaces = [iface for iface in interfaces if not iface.startswith('en')]
        return en_interfaces + other_interfaces  # Prioritize all 'en' interfaces
    else:
        return interfaces

```

itsknk:
> Eventually we should support automatically upgrading to the best interface available (in this case thunderbolt) - see #51. We can implement this later though, for now this is great.

This makes better sense, @AlexCheema.

AlexCheema:
Just to clarify, this still needs to be fixed / tested: https://github.com/exo-explore/exo/pull/47#issuecomment-2241334740

itsknk:
Update:
- Look for Thunderbolt in both the hardware and type fields, as Thunderbolt interfaces might be identified in either.
- Identify WiFi interfaces by checking if the type is 'wi-fi' (case-insensitive).
- Categories all other interfaces as 'other'.

@AlexCheema, I wrote a test script and it returned:

Screenshot 2024-07-20 at 11 42 16 PM

Please review this.

exo/networking/grpc/grpc_discovery.py
```diff
@@ -35,6 +37,42 @@ def __init__(self, node_id: str, node_port: int, listen_port: int, broadcast_por
...
+                else:
+                    other_interfaces.append(interface_name)
+
+            if self.DEBUG >= 2:
+                print(f"Thunderbolt interfaces: {thunderbolt_interfaces}")
+                print(f"WiFi interfaces: {wifi_interfaces}")
+                print(f"Other interfaces: {other_interfaces}")
+
+            # Prioritize Thunderbolt, then WiFi, then others
+            return thunderbolt_interfaces + wifi_interfaces + other_interfaces
```
AlexCheema, line 72: Beautiful code. love this

AlexCheema:
Have you got a thunderbolt cable to test this with? I don't think this works in its current state. This is what `system_profiler SPNetworkDataType -json` gives me for my MacBook Pro's Thunderbolt interface:

```
{
      "_name": "Thunderbolt Bridge",
      "Ethernet": {
        "MediaOptions": [],
        "MediaSubType": "autoselect"
      },
      "hardware": "Ethernet",
      "interface": "bridge0",
      "ip_address": [
        "169.254.13.209"
      ],
      "IPv4": {
        "AdditionalRoutes": [
          {
            "DestinationAddress": "169.254.13.209",
            "SubnetMask": "255.255.255.255"
          }
        ],
        "Addresses": [
          "169.254.13.209"
        ],
        "ConfigMethod": "DHCP",
        "ConfirmedInterfaceName": "bridge0",
        "InterfaceName": "bridge0",
        "SubnetMasks": [
          "255.255.0.0"
        ]
      },
      "IPv6": {
        "ConfigMethod": "Automatic"
      },
      "Proxies": {
        "ExceptionsList": [
          "*.local",
          "169.254/16"
        ],
        "FTPPassive": "yes"
      },
      "spnetwork_service_order": 4,
      "type": "Ethernet"
    }
```

AlexCheema:
I had another look and generally I don't think this is doing what we want it to do.
I'm not even sure the behaviour is different to before the changes. We're broadcasting on all interfaces (was already the case as far as I can tell) and listening on all interfaces (was already the case).

I think what we need to do here:

- In the broadcast message, add a `priority` field which indicates the device's preference to receive connections on this particular interface
- the listener should keep track of the current `priority` for each peer (can be in the same dict as `known_peers`)
- if the listener receives a broadcast with a higher `priority` than it currently has for a connected peer, it should disconnect on the old interface and reconnect on the new interface

itsknk:
Did:

- A `priority` field is added to the broadcast message.
- Modified `known_peers` to include the priority info.
- Implemented reconnection logic in `on_listen_message` when a higher priority interface is detected.
-  `get_network_interfaces` returns interfaces with their priorities.
- A `reconnect_peer` method is added to handle the reconnection.

@AlexCheema, wrote a small test script that simulates receiving a broadcast from a peer. And to see whether it keeps track of priority and change when a better one comes. Here is the output from the test:

Screenshot 2024-07-21 at 4 09 55 PM

AlexCheema:
> Did:
> 
> * A `priority` field is added to the broadcast message.
> * Modified `known_peers` to include the priority info.
> * Implemented reconnection logic in `on_listen_message` when a higher priority interface is detected.
> * `get_network_interfaces` returns interfaces with their priorities.
> * A `reconnect_peer` method is added to handle the reconnection.
> 
> @AlexCheema, wrote a small test script that simulates receiving a broadcast from a peer. And to see whether it keeps track of priority and change when a better one comes. Here is the output from the test:
> 
> Screenshot 2024-07-21 at 4 09 55 PM

@itsknk can you commit the test too?

itsknk:
> > Did:
> > 
> > * A `priority` field is added to the broadcast message.
> > * Modified `known_peers` to include the priority info.
> > * Implemented reconnection logic in `on_listen_message` when a higher priority interface is detected.
> > * `get_network_interfaces` returns interfaces with their priorities.
> > * A `reconnect_peer` method is added to handle the reconnection.
> > 
> > @AlexCheema, wrote a small test script that simulates receiving a broadcast from a peer. And to see whether it keeps track of priority and change when a better one comes. Here is the output from the test:
> > Screenshot 2024-07-21 at 4 09 55 PM
> 
> @itsknk can you commit the test too? and add some assertions so it's a proper test would be great

Done @AlexCheema

exo/networking/grpc/grpc_discovery.py
```diff
@@ -30,11 +32,42 @@ def __init__(self, node_id: str, node_port: int, listen_port: int, broadcast_por
         self.listen_port = listen_port
         self.broadcast_port = broadcast_port if broadcast_port is not None else listen_port
         self.broadcast_interval = broadcast_interval
-        self.known_peers: Dict[str, Tuple[GRPCPeerHandle, float, float]] = {}
+        self.known_peers: Dict[str, Tuple[GRPCPeerHandle, float, float, int]] = {}  # Added priority
```
AlexCheema, line 35: Why is there a comment here?

exo/networking/grpc/grpc_discovery.py
```diff
@@ -30,11 +32,42 @@ def __init__(self, node_id: str, node_port: int, listen_port: int, broadcast_por
...
+            interfaces = []
+            for interface in network_data.get('SPNetworkDataType', []):
+                interface_name = interface.get('interface')
+                interface_type = interface.get('type', '').lower()
+                hardware = interface.get('hardware', '').lower()
+
+                if not interface_name:
+                    continue
+
+                priority = self.get_interface_priority(interface_name, interface_type, hardware)
```
AlexCheema, line 54: This is a pure function. Move it outside of the class

exo/networking/grpc/test_network_interfaces.py
```diff
@@ -0,0 +1,91 @@
+import subprocess
+import json
+import psutil
+import time
+
```
AlexCheema, line 5: Once you move this off the class, get rid of the duplicate definition here

exo/networking/grpc/grpc_discovery.py
```diff
@@ -30,11 +32,42 @@ def __init__(self, node_id: str, node_port: int, listen_port: int, broadcast_por
         self.listen_port = listen_port
         self.broadcast_port = broadcast_port if broadcast_port is not None else listen_port
         self.broadcast_interval = broadcast_interval
-        self.known_peers: Dict[str, Tuple[GRPCPeerHandle, float, float]] = {}
+        self.known_peers: Dict[str, Tuple[GRPCPeerHandle, float, float, int]] = {}  # Added priority
         self.broadcast_task = None
         self.listen_task = None
         self.cleanup_task = None
 
+    def get_network_interfaces(self):
```
AlexCheema, line 40: Same here, this is a pure function so move it outside of the class

exo/networking/grpc/grpc_discovery.py
```diff
@@ -2,13 +2,47 @@
...
+import psutil
+import subprocess
 from typing import List, Dict, Callable, Tuple, Coroutine
 from ..discovery import Discovery
 from ..peer_handle import PeerHandle
 from .grpc_peer_handle import GRPCPeerHandle
 from exo.topology.device_capabilities import DeviceCapabilities, device_capabilities, UNKNOWN_DEVICE_CAPABILITIES
 from exo import DEBUG_DISCOVERY
 
+def get_interface_priority(interface_name, interface_type='', hardware=''):
```
AlexCheema, line 15: types please

exo/networking/grpc/grpc_discovery.py
```diff
@@ -2,13 +2,47 @@
...
+    if 'thunderbolt' in hardware or 'thunderbolt' in interface_type:
+        return 3
+    elif interface_type == 'wi-fi' or 'airport' in hardware:
+        return 2
+    elif interface_name.startswith('en'):
+        return 1
+    else:
+        return 0
+
+def get_network_interfaces():
```
AlexCheema, line 25: return type

exo/networking/grpc/grpc_discovery.py
```diff
@@ -73,32 +106,38 @@ async def discover_peers(self, wait_for_peers: int = 0) -> List[PeerHandle]:
...
+                        sock.setsockopt(socket.SOL_SOCKET, socket.SO_BROADCAST, 1)
+                        transport.sendto(message, ('<broadcast>', self.broadcast_port))
+                        transport.close()
+                    except Exception as e:
+                        if DEBUG_DISCOVERY >= 2: print(f"Error broadcasting on interface {interface}: {e}")
             except Exception as e:
-                print(f"Error in broadcast presence: {e}")
-                import traceback
-                print(traceback.format_exc())
+                if DEBUG_DISCOVERY >= 2: print(f"Error updating or accessing interfaces: {e}")
```
AlexCheema, line 138: why did the message change?

exo/networking/grpc/grpc_discovery.py
```diff
@@ -107,26 +146,47 @@ async def on_listen_message(self, data, addr):
...
             device_capabilities = DeviceCapabilities(**message['device_capabilities'])
-            if peer_id not in self.known_peers:
-                self.known_peers[peer_id] = (GRPCPeerHandle(peer_id, f"{peer_host}:{peer_port}", device_capabilities), time.time(), time.time())
-                if DEBUG_DISCOVERY >= 2: print(f"Discovered new peer {peer_id} at {peer_host}:{peer_port}")
-            self.known_peers[peer_id] = (self.known_peers[peer_id][0], self.known_peers[peer_id][1], time.time())
+            
+            if peer_id in self.known_peers:
+                _, _, _, current_priority = self.known_peers[peer_id]
+                if new_priority > current_priority:
+                    await self.reconnect_peer(peer_id, peer_host, peer_port, new_priority, device_capabilities)
```
AlexCheema, line 155: discovery is not responsible for connecting to peers, it just discovers them. connecting should happen in the `standard_node.py`

AlexCheema: this requires a bit of a refactor to do properly, I'm focusing on linux fixes right now but if you have any questions async, leave a comment on this PR and tag me

itsknk:
Hey @AlexCheema, yeah it does look a little complex. After some understanding, I thought of some steps. Please do verify these, and let me know if I'm on the right track.
- Let `GRPCDiscovery` class to handle discovering peers and maintaining their information, without any connection logic.
- The `StandardNode` class will handle the connection logic, including priority-based reconnections.
- The `update_peers` method in `StandardNode` can check for new peers or higher priority connections and manage the connections accordingly.
- The periodic topology collection remains, ensuring the node regularly updates its peer connections based on the latest discovery information.

AlexCheema:
@itsknk

First of all, thanks a lot for these contributions - experimentation helps a lot to inform future decisions even if things don't get merged. I end up throwing away most the code I write.

There have been quite a few changes since this PR (health checks, broadcast on multiple interfaces, priorities).

There's now a `priority` that is sent in broadcast messages. Connections for a given node_id are upgraded to the highest priority connection that is healthy.

I think this PR might need to be closed now, but I'm happy to send you a retrospective $100 bounty for your work and it helped inform some of the decisions made in the repo.

If you like, we are currently setting `priority` to 1 for every kind of interface. If we have a good way to assign meaningful priorities (higher is better) e.g. Thunderbolt > WiFi that would be great.

PR #49, opened 2024-07-20
Flag to disable Viz TUI

apotl:
In lieu of proper logging, and as a quicker workaround than doing so, I found the Viz TUI bothersome when trying to catch errors. I think it's also important to be able to disable the TUI in the context of running exo headless and capturing output. This PR ensures passing `--disable-tui` through the CLI disables Viz rendering.


AlexCheema:
LGTM, great work thank you!

PR #77, opened 2024-07-25
Update device_capabilities.py (first PR ever... hope I did okay lol)

Cloud1590:
Added flops for most modern NVIDIA and AMD GPUs.


Cloud1590:
reverted change to main.py
device_capabilities.py should be okay (I hope)

arch-user-france1:
In `return DeviceCapabilities(model=f"Linux Box ({gpu_name})", chip=gpu_name, memory=gpu_memory_info.total // 2**20, flops=CHIP_FLOPS.g`, the GPU's name is passed from `pynvml.nvmlDeviceGetName(handle)`. This name is then shown on the EXO CLI. The reason your flops are not inherited properly might be a disparity between your naming 'Nvidia GeForce RTX 4080' and pynvml's return value. I do not have an Nvidia GPU readily available, but you should be able to fix this problem by looking at the EXO CLI's output after 'Linux Box'. Happy hacking!

AlexCheema:
Thanks for filling all these in, great work!

tqolf:
maybe caseless compare for device name or device id ought to be considered? my device named with "NVIDIA GeForce RTX 4090" but not "Nvidia GeForce RTX 4090", and it worked after changes.

PR #87, opened 2024-07-26
refactor(mlx): model sharding and add deepseek v2 support

mzbac:
This will make porting other models easier and allow for reusing mlx_lm's `load_model` instead of copying and pasting it.

### DIFF
--- a/exo/api/chatgpt_api.py
+++ b/exo/api/chatgpt_api.py
@@ -54,7 +54,7 @@ shard_mappings = {
       model_id="mlx-community/Mistral-Large-Instruct-2407-4bit", start_layer=0, end_layer=0, n_layers=88
     ),
   },
-  # deepseek v2
+  ### deepseek v2
   "deepseek-coder-v2-lite": {
     "MLXDynamicShardInferenceEngine": Shard(
       model_id="mlx-community/DeepSeek-Coder-V2-Lite-Instruct-4bit-mlx", start_layer=0, end_layer=0, n_layers=27
@@ -67,6 +67,9 @@ class Message:
   def __init__(self, role: str, content: str):
     self.role = role
     self.content = content
+  def __init__(self, role: str, content: str):
+    self.role = role
+    self.content = content
 
 
 class ChatCompletionRequest:
@@ -74,9 +77,21 @@ class ChatCompletionRequest:
     self.model = model
     self.messages = messages
     self.temperature = temperature
+  def __init__(self, model: str, messages: List[Message], temperature: float):
+    self.model = model
+    self.messages = messages
+    self.temperature = temperature
 
 
 def resolve_tinygrad_tokenizer(model_id: str):
+  if model_id == "llama3-8b-sfr":
+    return AutoTokenizer.from_pretrained("TriAiExperiments/SFR-Iterative-DPO-LLaMA-3-8B-R")
+  elif model_id == "llama3-70b-sfr":
+    return AutoTokenizer.from_pretrained("TriAiExperiments/SFR-Iterative-DPO-LLaMA-3-8B-R")
+  else:
+    raise ValueError(
+      f"tinygrad doesnt currently support arbitrary model downloading. unsupported model: {model_id}"
+    )
   if model_id == "llama3-8b-sfr":
     return AutoTokenizer.from_pretrained("TriAiExperiments/SFR-Iterative-DPO-LLaMA-3-8B-R")
   elif model_id == "llama3-70b-sfr":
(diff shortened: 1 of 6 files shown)


exo/api/chatgpt_api.py
```diff
@@ -31,65 +31,84 @@
...
-        if DEBUG >= 2: print(traceback.format_exc())
-        if DEBUG >= 2: print(f"Failed again to load tokenizer for {model_id}. Falling back to mlx tokenizer")
+        if DEBUG >= 2:
+            print(traceback.format_exc())
+        if DEBUG >= 2:
+            print(f"Failed again to load tokenizer for {
+                  model_id}. Falling back to mlx tokenizer")
 
-    if DEBUG >= 2: print(f"Trying mlx tokenizer for {model_id}")
+    if DEBUG >= 2:
```
AlexCheema, line 96: Just a style thing, can you put these back on the same line

exo/api/chatgpt_api.py
```diff
@@ -119,31 +138,37 @@ def generate_completion(
     choice = completion["choices"][0]
     if object_type.startswith("chat.completion"):
         key_name = "delta" if stream else "message"
-        choice[key_name] = {"role": "assistant", "content": tokenizer.decode(tokens)}
+        choice[key_name] = {"role": "assistant",
```
AlexCheema, line 141: It looks like your linter is being aggressive with line lengths. We probably need a linter in general but lets leave that to a different PR.

AlexCheema:
Your linter has been quite aggressive on line lengths so a lot of lines have changed without any actual changes, just expanding to 2 lines. Could you change these back so it's easier to see what changed?

exo/inference/mlx/sharded_utils.py
```diff
@@ -107,17 +105,14 @@ def load_model_shard(
         raise FileNotFoundError(f"No safetensors found in {model_path}")
 
     weights = {}
-    all_weights_keys = set()
     for wf in weight_files:
-        weights_dict = mx.load(wf)
-        all_weights_keys.update(weights_dict.keys())
-        weights.update({k: v for k, v in weights_dict.items() if not k.startswith("model.layers.") or shard.start_layer <= int(k.split('.')[2]) <= shard.end_layer})
+        weights.update(mx.load(wf))
```
AlexCheema, line 109: Isn't this going to load the entire model into memory? We need to make sure only the layers for this shard are loaded into memory.

exo/inference/mlx/sharded_utils.py
```diff
@@ -128,18 +123,7 @@ def load_model_shard(
...
-            layer_num = int(k.split('.')[2])
-            if shard.start_layer <= layer_num <= shard.end_layer:
-                new_key = f"model.layers.{layer_num - shard.start_layer}." + '.'.join(k.split('.')[3:])
-                filtered_weights[new_key] = v
-        else:
-            filtered_weights[k] = v
-    weights = filtered_weights
-
-    model.load_weights(list(weights.items()), strict=False)
+    model.load_weights(list(weights.items()))
```
AlexCheema, line 126: Again, the part you removed is important to ensure that only the weights for this shard are loaded into memory.

AlexCheema:
Great job with adding deepseek v2 support. Tested this and worked great!

Please see my comments, just some small things. The most important thing is making sure we don't load the entire model into memory, as this would defeat the point of exo.

exo/inference/mlx/sharded_utils.py
mzbac: I don't know the exact details of `mx.load` implementation, but it should be lazy evaluation. That's why when you use mlx_lm.convert, you can convert a large model (>700GB) without requiring that amount of system memory. From my quick test, I don't feel it loads the entire weight into memory. Perhaps @awni could help confirm this.

exo/api/chatgpt_api.py
mzbac: @AlexCheema, would you share your linter or editconfig with me so I can set it up on my machine?

exo/inference/mlx/sharded_utils.py
mzbac: @AlexCheema based on the clarification from the MLX team, the above changes should be fine..detials -> https://github.com/ml-explore/mlx/discussions/1292

AlexCheema: Awesome - thanks for getting clarification on that

AlexCheema:
You can run `python3 format.py ./exo` to format with black, preserving the one-line `if DEBUG` lines.

exo/api/chatgpt_api.py
AlexCheema: Yup good point, see https://github.com/exo-explore/exo/pull/87#issuecomment-2254289002

exo/inference/mlx/sharded_utils.py
AlexCheema: It's okay as it's lazy loaded.

AlexCheema:
Looks good to me. Just needs main merged in and conflicts resolved

Issue #1, opened 2024-07-15
[BOUNTY - $200] Batched Requests


AlexCheema:
**Motivation:** Batching multiple inference requests together can speed up inference. Batching can even be leveraged with single-input settings for speedups with e.g. staged speculative decoding.

**What:** Currently, exo handles inference requests separately. This bounty is for batching inferences together, so that multiple inputs can be passed through model shards together in a single pass.

**Reward:** $200 Bounty paid out with USDC on Ethereum, email alex@exolabs.net

abdussamettrkr:
I want to work on this issue

abdussamettrkr:
Working on this

AlexCheema:
@abdussamettrkr let me know if you need any help / run into any issues. here to help

githubbin765:
Is this still available? Are there also other bounties? Are they open to anyone?

wnma3mz:
I volunteered to implement an EXO-like project that implemented batch requests.

https://github.com/wnma3mz/tLLM

Omar8345:
@AlexCheema Is this still available?

Bennethxyz:
Opened an initial PR adding batched sampling to reduce per-token overhead when multiple requests are active: https://github.com/exo-explore/exo/pull/891

This is a safe incremental step toward full forward-pass batching across shards. Happy to iterate on batched forward (combining per-request caches) next.

Omar8345:
@AlexCheema I want to actually work on this issue but please confirm if it is still available

rishi-jat:
@AlexCheema  I’m interested in working on this bounty. Is it still active, and may I begin?

rishi-jat:
??

Evanev7:
We're closing bounties, thank you to all who contributed, and apologies that we left this unmanaged for so long.

PR #108, opened 2024-07-31
batched inference

varshith15:
- [x] batched inference llama unit testcase
- [x] batched inference llava unit testcase
- [x] take list as input for prompt, image_str, request_id
- [x] fix callbacks order
- [x] online batching in chatgpt api
- [x] top_p!=1 for batched patch
- [x] early stopping in the batch if the stop has eos occurred before, currently only check the last token
- [x] online batching conditions update (timeout, max_batch etc)

### DIFF
--- a/exo/helpers.py
+++ b/exo/helpers.py
@@ -8,6 +8,7 @@ import psutil
 import uuid
 import netifaces
 from pathlib import Path
+import tempfile
 
 DEBUG = int(os.getenv("DEBUG", default="0"))
 DEBUG_DISCOVERY = int(os.getenv("DEBUG_DISCOVERY", default="0"))
@@ -34,7 +35,7 @@ def get_system_info():
 
 
 def find_available_port(host: str = "", min_port: int = 49152, max_port: int = 65535) -> int:
-  used_ports_file = os.path.join(os.path.dirname(os.path.abspath(__file__)), ".exo_used_ports")
+  used_ports_file = os.path.join(tempfile.gettempdir(), "exo_used_ports")
 
   def read_used_ports():
     if os.path.exists(used_ports_file):


varshith15:
hey @AlexCheema 
to start with, i am assuming that inflight batching isnt needed right away
we can just batch requests in the api section and do inference and split them back in the api section thoughts?

AlexCheema:
> hey @AlexCheema to start with, i am assuming that inflight batching isnt needed right away (have to revamp a lot the api to make inflight happen) we can just batch requests in the api section and do inference and split them back in the api section thoughts?

Sounds good. Yes, no need for inflight batching - we can add that in a subsequent PR.
Lets do this!

Related issue: https://github.com/exo-explore/exo/issues/1

varshith15:
Hey @AlexCheema 
I was just working on the batching of requests at the endpoint part but it just seems a little gimmck-y to merge them and separate them after processing because there are a bunch of moving parts like stream functionality etc

My current idea is to basically combine the requests ("req1_req2_req3..") and then check if req_id is present in the broadcasted result id and split it based on the position of the id. 

do you have a better idea? what did you have in your mind?

varshith15:
ive thought of a better way, just need to update functions to take list of req_ids, prompts and img_strs and broadcast back the specific results based on the index of the req in the back - that way dont have to change a lot of code in the chatgpt api

varshith15:
hey @AlexCheema, batched inference with online batching works now. There are few more small patches required, ill push them in a bit. Could you please review the idea meanwhile.

varshith15:
@AlexCheema it's done! PRM

AlexCheema:
Thanks @varshith15 I will take a proper look at some point next week.

AlexCheema:
I haven't had a proper look yet, but I'm wondering what the behaviour is when a request is being processed and another request comes in?

varshith15:
currently the requests just keep getting processed one after the other, by processes i mean after the await process_prompt is done but maybe a semaphore is needed

AlexCheema:
> currently the requests just keep getting processed one after the other, by processes i mean after the await process_prompt is done but maybe a semaphore is needed

I think this might cause issues currently since the kv_cache is shared. Can you test this out to see if it works? It doesn't work on main but maybe with your changes it works now.

varshith15:
I've tested it out in a bunch of different scenarios with sending the reqs to the service, it works as expected

varshith15:
@AlexCheema you are right, there a bug when theres a request running already and a new request comes, the older one is not able to return, checking it

varshith15:
@AlexCheema could you expand on why concurrent processing doesn't work as the kv_csche is shared?

AlexCheema:
> @AlexCheema could you expand on why concurrent processing doesn't work as the kv_csche is shared?
> 
> I haven't debugged yet, will do it over the weekend

I pushed something for MLX that uses a LRU cache with a bunch of KV caches that should fix it. Can you merge the latest exo and fix conflicts then I will take a look through the PR.

varshith15:
@AlexCheema still doesnt work as expected, i tried sending two reqs with a slight delay between them, one gives the expected answer and the other gives some unicode
checking this

varshith15:
i think we should have a semaphore and have the max_caches value to be equal to semaphore
with the LRU cache fix the previous error is fixed now, it works as expected

exo/helpers.py
```diff
@@ -93,22 +94,30 @@ def terminal_link(uri, label=None):
...
+
+      await asyncio.wait_for(wait_for_valid_result(), timeout)
+      return self.result.popleft()
 
   def on_next(self, callback: Callable[..., None]) -> None:
     self.observers.append(callback)
 
   def set(self, *args: T) -> None:
-    self.result = args
+    self.result.append(args)
```
varshith15, line 117: @AlexCheema  the concurrent execution issue is due to this, not due to the LRU

varshith15:
@AlexCheema  this is done now, please review

Issue #104, opened 2024-07-30
Parallel model downloading


AlexCheema:
Even with https://github.com/exo-explore/exo/issues/16, we would only download model shards one at a time sequentially.
We can do this in parallel across all devices.

Issue #99, opened 2024-07-30
Bittorrent to distribute models amongst nodes


austinbv:
I am opening an issue for discussion as well as a place holder for work where we use bittorrent to distribute large models amongst nodes.

Implementation Path
- Network should check to see if the model is downloaded on any of the existing nodes.
- If the model doesn't exist on any node one node should register itself as the `Seed node`
- Seed node should download the model from hugging face and distribute the model via bittorrent to other nodes
- As nodes get parts of the seed they should distribute their parts to other nodes
- If the model exists on one or more nodes those nodes should advertise as seeds
- New nodes to a network will leech from existing nodes rather than the internet

✅ Pros
- Allows for networks to minimize model downloads saving hundreds of gigs if not TB of data use
- Speeds up distribution of models in a network - after the seed node has fully downloaded the node all nodes will finish leeching the model around the same time
- Models are guaranteed to be the same b/c of bittorrent native hash checking

AlexCheema:
I really like this idea. @RayFernando1337 suggested something similar when we were running into issues with slow download for Llama 3.1 405b.

Lets use this as a design discussion.

stephanj:
Possible option libtorrent with [Python bindings](https://www.libtorrent.org/python_binding.html) 

- Robust, feature-rich C++ library with Python bindings
- Widely used and well-maintained
- Supports both BitTorrent and WebTorrent protocols
- Python package: python-libtorrent

Pros:

- High performance and scalability
- Extensive feature set
- Active development and community support

Cons:

- C++ core might introduce complexity in deployment
- Potential issues with Python version compatibility

Other option: [qBittorrent-API](https://pypi.org/project/qbittorrent-api/)

⚠️ Claude Sonnet 3.5 code refactoring recommendations: 

To design expanded node capabilities for BitTorrent roles in the Exo system, we'll need to enhance the existing node structure to incorporate BitTorrent functionality. Here's a proposed design approach:

1. BitTorrent Manager Class

First, let's create a new class to manage BitTorrent operations:

```python
import libtorrent as lt
from exo.inference.shard import Shard

class BitTorrentManager:
    def __init__(self, save_path: str):
        self.session = lt.session()
        self.save_path = save_path
        self.torrents = {}

    async def add_torrent(self, torrent_file: str, shard: Shard):
        info = lt.torrent_info(torrent_file)
        handle = self.session.add_torrent({'ti': info, 'save_path': self.save_path})
        self.torrents[shard] = handle

[...]

Issue #80, opened 2024-07-25
Can I manually download model files if my server cannot access huggingface.co?


artistlu:
I'm encountering an issue when trying to use the Exor project. It appears that my server is unable to access the `huggingface.co` domain, which is preventing me from downloading the required model files.

Is there a way for me to manually download the model files and then place them in the appropriate directory on my server? If so, could you please provide the steps or the specific directory path where I should put the downloaded model files?

Alternatively, do you have any other suggestions on how I can work around this issue and successfully use the Exo on my server?

```
Trying AutoTokenizer for llama3-8b-sfr
Traceback (most recent call last):
  File "/root/miniconda3/envs/exo/lib/python3.12/site-packages/urllib3/connection.py", line 196, in _new_conn
    sock = connection.create_connection(
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/root/miniconda3/envs/exo/lib/python3.12/site-packages/urllib3/util/connection.py", line 85, in 
create_connection
    raise err
  File "/root/miniconda3/envs/exo/lib/python3.12/site-packages/urllib3/util/connection.py", line 73, in 
create_connection
    sock.connect(sa)
TimeoutError: timed out

The above exception was the direct cause of the following exception:

Traceback (most recent call last):
  File "/root/miniconda3/envs/exo/lib/python3.12/site-packages/urllib3/connectionpool.py", line 789, in urlopen
    response = self._make_request(
               ^^^^^^^^^^^^^^^^^^^
  File "/root/miniconda3/envs/exo/lib/python3.12/site-packages/urllib3/connectionpool.py", line 490, in 
_make_request
    raise new_e
  File "/root/miniconda3/envs/exo/lib/python3.12/site-packages/urllib3/connectionpool.py", line 466, in 
_make_request
    self._validate_conn(conn)
  File "/root/miniconda3/envs/exo/lib/python3.12/site-packages/urllib3/connectionpool.py", line 1095, in 
_validate_conn
    conn.connect()
  File "/root/miniconda3/envs/exo/lib/python3.12/site-packages/urllib3/connection.py", line 615, in connect
    self.sock = sock = self._new_conn()
                       ^^^^^^^^^^^^^^^^
  File "/root/miniconda3/envs/exo/lib/python3.12/site-packages/urllib3/connection.py", line 205, in _new_conn
    raise ConnectTimeoutError(
[...]

Issue #79, opened 2024-07-25
Allow LLM copy from the local network instead of downloading from Hugging Face


stephanj:
Implementing a feature to allow LLM copying from the local network instead of downloading from Hugging Face is an excellent way to optimize your setup, especially for multi-node environments. Here's a detailed approach to implement this feature:

1. Local File Server Setup:
   - Set up a lightweight file server (e.g., using Python's `http.server` or a more robust solution like nginx) on one of the nodes or a dedicated machine in your local network.
   - This server will host the model files once they're downloaded.

2. Model Registry:
   - Create a simple "model registry" service that keeps track of which models are available locally.
   - This could be a simple key-value store or a database with model names/versions and their local network locations.

3. Download and Share Process:
   - Modify the model loading process to follow these steps:

```python
import requests
import os
from pathlib import Path

def get_model(model_name: str, local_cache_dir: Path):
    registry_url = "http://registry-server:8000/models"
    local_file_server = "http://file-server:8080/models"

    # Check if model is in local registry
    response = requests.get(f"{registry_url}/{model_name}")
    if response.status_code == 200:
        # Model is available locally
        model_url = f"{local_file_server}/{model_name}"
    else:
        # Model not available locally, use Hugging Face
        model_url = f"https://huggingface.co/{model_name}/resolve/main/model.safetensors"

    # Download the model
    response = requests.get(model_url, stream=True)
    model_path = local_cache_dir / f"{model_name}.safetensors"
    with open(model_path, 'wb') as f:
        for chunk in response.iter_content(chunk_size=8192):
            f.write(chunk)

    # If downloaded from Hugging Face, update registry and copy to file server
    if "huggingface.co" in model_url:
        requests.post(f"{registry_url}/{model_name}", json={"path": str(model_path)})
        copy_to_file_server(model_path, model_name)

    return model_path

def copy_to_file_server(model_path: Path, model_name: str):
    # Implementation depends on your file server setup
    # This could be a file copy, an HTTP POST, or any other method to transfer the file
    pass
```

4. Registry Server:
   - Implement a simple HTTP server that maintains the registry:

```python
from flask import Flask, request, jsonify

app = Flask(__name__)
model_registry = {}

@app.route('/models/<model_name>', methods=['GET'])
def check_model(model_name):
    if model_name in model_registry:
        return jsonify({"available": True, "path": model_registry[model_name]})
    return jsonify({"available": False}), 404

@app.route('/models/<model_name>', methods=['POST'])
[...]

PR #124, opened 2024-08-05
Refactor model download

AlexCheema:
- Unify downloading, so all inference engines can use the same downloading logic
- Async downloads
- More robust, retryable, pausable/resumable downloads
- Progress for downloads #60
- Exo controls downloads now so we can swap this out for different implementations, for example P2P (see #79 #99) or alternative (see #80)
- Makes it easier to implement #104
- Makes it easier to implement #16

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -17,11 +17,11 @@ commands:
             source env/bin/activate
 
             # Start first instance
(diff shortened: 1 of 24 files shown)


AlexCheema:
Tests green, lets get this beast merged

Issue #119, opened 2024-08-05
Add support for Linux distributions and Docker containers


rekpero:
## Current Situation
Currently, the project lacks official support for Linux distributions and does not provide an option to run within a Docker container.

## Feature Request
1. Add support for running the project on Linux distributions.
2. Provide a Docker container option for easy deployment.

## Rationale
These additions would significantly enhance the project's accessibility and usefulness:

1. **Linux Support**:
   - Enables users with Linux systems or servers to run the project without building from source.
   - Allows for clustering different systems to create inference endpoints, broadening the project's applicability.

2. **Docker Container**:
   - Simplifies deployment across various environments.
   - Facilitates easy setup and management of the project.

## Use Cases
1. **Diverse System Support**: Many users have Linux systems or servers and want to cluster different systems for creating inference endpoints. Official Linux support would make this process smoother.

2. **Small Device Aggregation**: At Spheron, we're working on aggregating smaller devices (e.g., laptops with GPUs like 3080s, 3090s) to be part of a network. Running this project in a containerized environment would showcase how users can create clusters with numerous smaller devices for AI inference tasks.

## Benefits
- Increased adoption due to broader system compatibility.
- Easier deployment and management for users and organizations.
- Enhanced scalability for distributed computing scenarios.

We believe these additions would greatly benefit the community and expand the project's potential use cases. @AlexCheema Let us know if you need any further information or clarification.

PR #125, opened 2024-08-06
feat(docker): add dockerfile to build

dan-online:


Axodouble:
Noticed this pull request was closed, is there a particular reason why?
As there doesn't seem to be a lot of information, and I noticed that #119 requests the docker support.

ISJ-439:
Yes this would be a crucial addition for enterprise deployment. The ability to reduce dependency issues will also be desirable.

Scot-Survivor:
https://github.com/exo-explore/exo/pull/173/

Re-opened, was originally opened by mistake, hence why this was closed.

PR #183, opened 2024-08-27
Add Llama.cpp Support

### LINKED ISSUE #167: [BOUNTY - $500] Llama.cpp inference engine
- it should automatically detect the best device to run on
- We should require 0 manual configuration from the user, by default llama.cpp for example requires specifying the device

bayedieng:
This PR adds support for Llama.cpp and closes #167.


AlexCheema:
Hey @bayedieng just checking in. Anything I can help to move this along?

bayedieng:
Hey @AlexCheema I indeed was initially having trouble understanding the codebase however it's clearer now (Inheritance can be confusing). I wrote a basic sharded inference engine class and will proceed with the implementation.

My plan is to largely follow the implementation of the pytorch and tinygrad inference engines implementation with the only exception being skipping the tokenizer part of the problem. The Llama CPP API tokenizer is tied to the ```Llama``` class being instantiated. Also, the Tokenizer being defined in the other implementations don't seem to be tokenizing inputs but rather applies a chat template in the ```handle_chat_completions``` function of the ChatGPT API. I will be implementing it manually later in the call stack.

I will let you know if I have any further questions and am looking to have atleast a one working ai model being inferenced later today.

exo/models.py
```diff
@@ -5,6 +5,8 @@
   "llama-3.1-8b": {
     "MLXDynamicShardInferenceEngine": Shard(model_id="mlx-community/Meta-Llama-3.1-8B-Instruct-4bit", start_layer=0, end_layer=0, n_layers=32),
     "TinygradDynamicShardInferenceEngine": Shard(model_id="mlabonne/Meta-Llama-3.1-8B-Instruct-abliterated", start_layer=0, end_layer=0, n_layers=32),
+    "LlamaCppDynamicShardEngine": Shard(model_id="lmstudio-community/Meta-Llama-3.1-8B-Instruct-GGUF", start_layer=0, end_layer=0, n_layers=32)
```
bayedieng, line 8: Hey @AlexCheema, I'm having an issue here actually getting this model here to download. It is a valid huggingface model and I get no exception involving the download portion however, there is no progress being displayed. After some print debugging the ```download_repo_files``` just seems to exit right on this line https://github.com/exo-explore/exo/blob/87e08f89f1abd6711ee7686ac0bce02bbfb457d6/exo/download/hf/hf_helpers.py#L282

Again with no exception. Inferencing should be fairly straight forward after this is solved.

AlexCheema:
How is this going @bayedieng? Anything I can help with?

bayedieng:
I started a branch using ggml given that the llama cpp api doesn't expose the weights and thus wouldn't be able to be sharded. What would then be the requirements of having this PR done then? Each model type (vision or pure text LLM) would have to be implemented separately as ggml does not have an auto model generation similar to pytorch.

AlexCheema:
> I started a branch using ggml given that the llama cpp api doesn't expose the weights and thus wouldn't be able to be sharded. What would then be the requirements of having this PR done then? Each model type (vision or pure text LLM) would have to be implemented separately as ggml does not have an auto model generation similar to pytorch.

Let's start with Llama, which would support Llama, Phi and Mistral model weights since they're all based on Llama.

AlexCheema:
How's this going @bayedieng? Anything I can help with?

bayedieng:
Was waiting on confirmation for the bounty requirements. I should have a working llama implementation within the working week.

AlexCheema:
> Was waiting on confirmation for the bounty requirements. I should have a working llama implementation within the working week.

Checking in again.
You can use the PR for `TorchInferenceEngine` as a reference: https://github.com/exo-explore/exo/pull/139

bayedieng:
Thanks, I'm quite used to the Exo codebase at this point however, I've been struggling quite a bit with the GGML API as it's quite low level. it is largely undocumented and essentially any error you may have with regards to it will lead to a crash with very little information as to why.  

This will likely take much more time than I had initially anticipated. With that being said I think a simpler solution to support llama cpp would be to use candle as it also supports the GGUF format for weights that llama cpp uses, and also has python bindings with an implementation of quantized [llama](https://github.com/huggingface/candle/blob/main/candle-pyo3/py_src/candle/models/llama.py) which would essentially be the same I would have implemented using GGML but with a much simpler API.

I fully understand if you'd still want to go forward with llama cpp as it may not fill your needs but using candle would be much [simpler](https://github.com/abetlen/ggml-python#usage).

bayedieng:
Made a few simpler changes in another branch that is in conflict with this one will be closing and opening a new one.

PR #203, opened 2024-09-05
Non blocking

AlexCheema:
- The issue was that MLX and Tinygrad block and don't have async API's
- Exo uses asyncio for everything and so stuff was getting blocked when MLX and Tinygrad would run. e.g. when they load al large model into memory, a lot of I/O would time out.
- Now uses ThreadPoolExecutor with 1 thread to ensure that MLX and Tinygrad operations always run on their own dedicated thread in a way that is asyncio compatible. Seems to work well, and even got some inference speedups


AlexCheema:
LGTM 💯 🥇 👍

PR #209, opened 2024-09-08
move `.exo_used_ports` to `/tmp`

GaetanLepage:
Instead of writing the `.exo_used_ports` file in the package directory, have it in `/tmp`.


AlexCheema:
looks good, thank you!

PR #210, opened 2024-09-08
Improve setup.py for proper installation

GaetanLepage:
I am working towards packaging `exo` to [nixpkgs](https://github.com/NixOS/nixpkgs/).
This required a few patches to the source code.

In this PR, there are two of them, which I think don't hurt to have.
Currently, `exo` is not expected to be installed as a proper python library (using `pip install`) but rather ran directly from the source directory.
However, to package it to nixpkgs, we need to have it behave as a proper package.
The good news is that there is not too much to do.

- **move main.py script to package dir**
    Right now, the `main.py` script is omitted and not copied to the `lib/python3.12/site-packages/exo/` install folder.
    By moving the source file under the `exo/` subfolder, it gets properly included.
- **add entry_point exo to run main.py**
  This allows being able to run `exo` from the command line and get the application to start.
- **add missing __init__.py files**
  Two `__init__.py` files were missing in some subdirectories which lead to them not being copied to the install folder.
- **move tinychat inside exo package**
  Same, ensure that the `tinychat` webapp is installed too.


Teebor-Choka:
Have you managed to nix package your fork, @GaetanLepage ?

GaetanLepage:
Yes, I have made it work (almost, my GPU doesn't support bfloat16 so I could not have my prompt answered).
Those changes should be enough.

Teebor-Choka:
> Yes, I have made it work (almost, my GPU doesn't support bfloat16 so I could not have my prompt answered). Those changes should be enough.

Do you have a repo for the nix package?

GaetanLepage:
Yes, I have opened a draft PR here: https://github.com/NixOS/nixpkgs/pull/345352

GaetanLepage:
@AlexCheema would you like me to do/test anything else about this PR ?

exo/main.py
AlexCheema, line 1: The instructions in the README would need to be updated to reflect this change. Can we keep it in the root? Or if not can we have a `main.py in the root that simply runs `exo/main.py`?

AlexCheema:
This looks good. Conflicts need to be fixed and README needs to be updated.
You can also add the info about running the nix package in the README!

exo/main.py
Teebor-Choka: Once installed through `pip install .`, the command `exo` is used to replace `python3 main.py` altogether.

AlexCheema: that sounds good. need to make this this works reliably on all platforms and update the README with the updated instructions.

GaetanLepage:
> You can also add the info about running the nix package in the README!

I will wait for it to be successfully packaged before advertising it here. But this is a great idea :)

Teebor-Choka (approved):

AlexCheema:
> > You can also add the info about running the nix package in the README!
> 
> I will wait for it to be successfully packaged before advertising it here. But this is a great idea :)

Awesome. Great work on this, I like this better than `python main.py`.

I think we need to inform the community about this change once it's successfully packaged.

GaetanLepage:
> Awesome. Great work on this, I like this better than `python main.py`.

Thanks for reviewing and merging this patch. I'm happy to somehow contribute to exo !



> I think we need to inform the community about this change once it's successfully packaged.

Sure, maybe doing a post in your discord could be useful.

Issue #219, opened 2024-09-16
Update README


vectrocomputers:
The README says to using the command:

`python3 main.py`

But the correct command is actually:

`CUDA=1 python3 main.py`

nicholasyfu1:
I can fix this

nicholasyfu1:
https://github.com/exo-explore/exo/pull/224

AlexCheema:
I'm not sure about this. In tinygrad we use the default device which should be CUDA on an NVIDIA GPU instance.

<img width="668" alt="Screenshot 2024-09-20 at 17 17 23" src="https://github.com/user-attachments/assets/78ea8915-b6d9-4788-a24e-4bd497ccfe3a">

AlexCheema:
Closing as I don't think this is necessary. Typically this means you didn't install CUDA.

PR #224, opened 2024-09-19
Update README.md

nicholasyfu1:
In response to issue #219


AlexCheema:
Closing as I don't think this is necessary. Typically this means you didn't install CUDA.

PR #226, opened 2024-09-21
fix allow patterns

bayedieng:
A bitwise or is done in-case the weight map for the model does not exist, this returns an error. the ```default_patterns``` set should be converted to a list than extended with ```shard_specific_patterns```


AlexCheema:
"this returns an error" - what is the error?
This works fine for me.

bayedieng:
It occurs when the model doesn't have a weight map. ```return list(default_patterns | shard_specific_patterns)``` is invalid python code.

bayedieng:
Also, I'm not sure whether it was intented, but an allowed pattern is never returned unless there is no weight map. It returns to ```None``` when there's one.

AlexCheema:
Okay well, we still need to de-duplicate the end-result. Can you change the return line to `return list(set(allowed_patterns))`

bayedieng:
Done.

AlexCheema:
I've fixed this in a way that's a bit simpler. Take a look, if you're happy I'll merge.

Issue #168, opened 2024-08-22
Thunderbolt Connection Causes LLM Requests to Freeze When One Mac Is Offline and the Other Is Connected to a Network


mcollins123:
I am encountering an issue when using exo with a setup involving two Macs connected via Thunderbolt. The problem occurs under the following conditions:

**Setup:**

- Mac Studio (connected to the internet via Ethernet)

- M2 MacBook Air (connected to the Mac Studio via Thunderbolt, but offline)

**Behavior:**

- If both Macs are completely offline (i.e., not connected to any network), exo works as expected over Thunderbolt. However, if the Mac Studio is connected to the network (via Ethernet), and the MacBook Air is offline, the LLM request freezes.

- I suspect that exo might be attempting to communicate with the offline MacBook Air over the network instead of using the Thunderbolt connection. This leads to the request freezing because the MacBook Air is not connected to the network.

**Steps to Reproduce:**

- Connect a Mac Studio to the network using Ethernet.

- Connect an M2 MacBook Air to the Mac Studio via Thunderbolt. Ensure the MacBook Air is offline (not connected to any network).

- Run exo on the Mac Studio with the MacBook Air as part of the cluster.

- Send a request to the LLM.

- Observe that the request freezes.

**Expected Behavior:**

exo should prioritize the Thunderbolt connection to communicate with the offline MacBook Air and not attempt to use the network interface. The LLM request should be processed without freezing.

**Actual Behavior:**

exo appears to freeze the LLM request, potentially because it is trying to communicate over the network instead of using the Thunderbolt connection.

**Additional Information:**

exo version: Latest from main as of August 22, 2024

vanakema:
@mcollins123 I fixed this issue in my PR. Awaiting review from the team, but in the meantime you can pull my fork to try it out. In my tests, it's very resilient to disabling and enabling WiFi or Ethernet. Only situation that made it fail was when both machines had wifi on, I started a long inference request, and then killed the wifi on one of the machines mid request (but even though that in progress request failed, subsequent requests worked just fine)

vanakema:
@mcollins123 the fix is merged in. If it fixes your issue, please consider closing!

Evanev7:
Should be fixed in 1.0 - feel free to reopen if it is still an issue.

Issue #51, opened 2024-07-20
Switch grpc peer connections over to the best interface available


AlexCheema:
- This PR implements prioritising thunderbolt interfaces for grpc peer connections: https://github.com/exo-explore/exo/pull/47
- Right now if a peer is already connected on WiFi, and then you connect thunderbolt, the connection will **not** upgrade to Thunderbolt. It would require a full restart of the node
- We should switch over to the best interface available (for now the best is thunderbolt)

Evanev7:
Should be fixed in 1.0!

PR #227, opened 2024-09-22
Fix issue where offline node cannot detect online node over thunderbolt

vanakema:
First, I just wanna say how awesome this project is, and how much potential I think it has.

With the subtle fanboyism out of the way, let's dive into this PR.

## What this does
This fixes an issue where an "offline" node (meaning a node not connected to any network besides the p2p thunderbolt bridge) cannot detect an "online" node (meaning a node connected to wifi or eth as well as to the p2p thunderbolt bridge) over thunderbolt. Details about this issue can be seen at #168.

## Why did this problem happen
This is due to the online node not broadcasting over the Thunderbolt bridge when non-bridge (meaning non-thunderbolt) alternatives exist (ie. wifi or eth). As noted in #168, for some reason when both nodes are "offline", they can both detect each other just fine, so broadcasting on 0.0.0.0 does work when no other alternatives exist.

## Why did you even try to do things half offline?
I did it because I wanted to force it to use Thunderbolt as the interconnect since I was not getting good performance, but I also wanted to be able to access the internet on my primary machine. Sure enough, when WiFi or Ethernet is available, inference is significantly slower even though a Thunderbolt bridge is actively available.

## Some extra info and thoughts
Something #168 didn't explicitly call out is that the "online" node can see the "offline" node, but the "offline" node CANNOT see the "online" node (and this is visible in the TUI). This makes the "online" node think everything is ready to go in terms of serving inference, but if you actually make an inference request, it will just freeze. This may be a sign that a bigger change is necessary here to make sure each node can mutually communicate with the other nodes, with better, more "intelligent" logging that does its best to explain where in the chain communication failed.

### Unrelated changes I also made
I also extracted the broadcasting logic into a DatagramProtocol to follow similar pattern as the listener, as well as running lint (but only on this file as the project is littered with linting errors haha)

### DIFF
--- a/exo/networking/udp_discovery.py
+++ b/exo/networking/udp_discovery.py
@@ -7,7 +7,7 @@ from typing import List, Dict, Callable, Tuple, Coroutine
 from .discovery import Discovery
 from .peer_handle import PeerHandle
 from exo.topology.device_capabilities import DeviceCapabilities, device_capabilities, UNKNOWN_DEVICE_CAPABILITIES
-from exo.helpers import DEBUG, DEBUG_DISCOVERY
+from exo.helpers import DEBUG, DEBUG_DISCOVERY, get_all_ip_addresses
 
 class ListenProtocol(asyncio.DatagramProtocol):
   def __init__(self, on_message: Callable[[bytes, Tuple[str, int]], Coroutine]):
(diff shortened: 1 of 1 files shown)


AlexCheema:
Thank you @vanakema this looks great! Glad you're enjoying exo.

Great work on this and making it easy to understand what you did and why. At some point I'd like to prioritise the fastest interconnect, i.e. thunderbolt over WiFi. I'll merge this now and we can fix that later issue here: https://github.com/exo-explore/exo/issues/51

PR #228, opened 2024-09-22
Update README.md

Yazington:
Adding Linux support (tested and getting 7.5 tokens/s for 4090 with llama 3.1 8B)
Screenshot from 2024-09-22 00-12-38


AlexCheema:
Awesome addition, thanks a lot!

PR #229, opened 2024-09-22
Tailscale

### LINKED ISSUE #170: Tailscale Discovery Module
- there’s a Tailscale Python lib
- We can automate a lot of the lookups etc that users would have to do themselves with manual configuration

AlexCheema:
Fixes #170


AlexCheema:
lgtm, tailscale ftw

tne-ops:
OK I see this but how do we use it. I looked the the discovery modules and set TAILSCALE_API_KEY and TAILSCALE_TAILNET, but how doe thi module get called?

PR #236, opened 2024-09-26
Add error toast in tinychat

drew-royster:
* error toast appears for 5 seconds displaying `detail` from the response body
* `this.generating` in `finally()` so that it doesn't get stuck on error

image

not the cleanest but I figure it's better than needing to check console logs 

never used alpine before so happy to change if I am doing anything wrong


AlexCheema:
Awesome QoL upgrade!

Issue #254, opened 2024-10-01
System requirements 


shad-ct:
Please add the system requirements too. That'd be very helpful.

larson-carter:
@shad-ct 

Did this PR solve what you are looking for: https://github.com/exo-explore/exo/pull/255

As seen in Alex's comment "The important thing is having enough VRAM across all your devices to fit the entire model in memory."

I guess in the future we can just pull stats every now and then, via a simply python script then update the readme if that is what you propose?

AlexCheema:
Just updated the README since exo only requires that you have enough memory across all your devices and is designed to run on anything that has a Python runtime.

AlexCheema:
Was there anything else you had in mind here @shad-ct

PR #255, opened 2024-10-01
Hardware Requirements

DevEmilio96:
Added hardware requirements to readme.md 
Issue: #254


AlexCheema:
This is good to have, I'm just not sure if this is fully accurate.
I've run exo on 8GB Mac Minis for example.
I've also seen others run on exo on raspberry pi's.
Where we are heading towards is being able to run this on anything. The important thing is having enough VRAM across all your devices to fit the entire model in memory.

DevEmilio96:
Exactly, that's why I mentioned in the notes that the hardware requirements are indicative, as they depend on various factors. If you'd like, you can specify the exact corrections to make, or you can accept the PR and modify it directly. This way, I can proceed with the other PRs without having to create new branches.

AlexCheema:
Merged for now and will make corrections.

Thanks a lot for your contribution!

Issue #250, opened 2024-09-30
[BOUNTY - $100] Display simple download status in web UI


AlexCheema:
- A user's first experience with exo is to run it, open the web ui, then send a message
- Often users don't know that exo has to download the models and they just see the first request hang for a long time
- It would be better if we transparently show the download progress in the web UI

EzraEllette:
I'm up for the "challenge." I'll add a download progress toast that's similar chromium's download progress display.

iyiolacak:
Hey @EzraEllette, I’m interested in working on this too. Are you actively working on it? If not, I can take a crack at it.

EzraEllette:
> Hey @EzraEllette, I’m interested in working on this too. Are you actively working on it? I'd love to collaborate if you need any assistance.

Hi, I am actively working on it if you want to pair that's fine with me.

EzraEllette:
initial with mock data:

https://github.com/user-attachments/assets/0729649f-df78-4164-98e8-4343ca4ee404

Changes to make:
- Show download complete toast on complete
- Show model name from repo_id
- handle multiple nodes
    - Show progress either for each node, or as a whole. Will initially show one bar for all progress

EzraEllette:
@AlexCheema

EzraEllette:
Current:

https://github.com/user-attachments/assets/99735645-0643-429a-8bf2-a81e084dc503


https://github.com/user-attachments/assets/c5565d5e-55d8-4a9a-82b4-e155a6ed80fe

Next Steps:
 - Clean up
 - post final demo

EzraEllette:
@AlexCheema The download status PR from Emilio doesn't actually handle downloads from multiple nodes and the progress popup css is broken. Here is a video demonstrating that the progress from other nodes isn't available.

https://github.com/user-attachments/assets/31ebbc2b-a02d-45d1-8a41-e20bbdc7ac5a

EzraEllette:
https://github.com/user-attachments/assets/30b780e0-792e-44e1-bbcb-a91b1241c8f7

EzraEllette:
https://github.com/user-attachments/assets/31660530-08f7-44a2-90da-e3d3242b23e5

iyiolacak:
@AlexCheema Is this still open? I was working on it and I just saw a download status in the web UI after I fetch the up to date repo.

EzraEllette:
@iyiolacak I think it's safe to assume he's happy with the rugged implementation that was merged into the main branch.

[...]

PR #258, opened 2024-10-01
Display simple download status in web UI

DevEmilio96:
Non-infinite pooling, it is activated only when the user sends a message, and stops either when the page is refreshed or when the download finishes

Issue : #250


AlexCheema:
Great job with this. I've added a retrospective $100 bounty for this as I think it deserves it. Stuff like this is really important for UX.

One annoying thing about this behaviour is the first time you use exo when the model is not downloaded, the request times out and you see nothing.
What would make this really nice from a UX perspective is to automatically retry the request once the download completes.

DevEmilio96:
Thank you for the bounty, but it’s not necessary – I appreciate the gesture though!

Just a quick reminder, we have a call scheduled for Thursday at 4:00 PM. I’ll also make sure to add the other modification you requested regarding the behavior of the model download.

AlexCheema:
Just ping me when you've resolved conflicts + fixed the retry.
This is a really valuable QoL PR I'd like to get merged for all the new users that are trying out exo.

DevEmilio96:
Hi @AlexCheema, I’ve resolved all the conflicts, and everything should be good to go now. I’ll handle the "fixed the retry" modification in a separate commit, so you can go ahead and merge the changes related to this PR.

PR #268, opened 2024-10-02
better ring topology image

AlexCheema:


AlexCheema:
Current image is fine.

PR #278, opened 2024-10-03
add a priority to broadcast messages 

AlexCheema:
add a priority to broadcast messages where the broadcaster can indicate how to prioritise that particular interface. for now all priorities are set to 1 but in the future this will be based on network latency, bandwidth, jitter, etc.. e.g. Thunderbolt prioritised over


AlexCheema:
lgtm

Issue #275, opened 2024-10-02
download progress & error messages


DevEmilio96:
Hi @AlexCheema,

I wanted to inform you about two issues:

-The download progress fetch now loops infinitely even when there's no download in progress.
-When there's insufficient memory to start a model, the error message is no longer being displayed.

If I can help you, please let me know.

merge : #272

AlexCheema:
- What's the problem with periodically fetching the download progress when it's not in progress?
- Not sure what you mean by this. Do any errors show? Is this the only error that doesn't show? What do you mean by insufficient memory?

DevEmilio96:
Regarding the first point, I just wanted to inform you about the infinite loop because I thought it might be something you'd prefer to avoid, even though it's not currently causing any issues.

As for the second point, when there isn't enough memory for the model to calculate a response, exo used to display an "out of memory" error, but now that error only appears in the console and not in tinychat. I haven't checked if this issue affects other error messages as well.

Let me know if you need further details.

AlexCheema:
"infinite loops" are everywhere in exo. It's completely fine and the desired behaviour. Some things we want to run periodically, there's nothing wrong with that.

Regarding the second issue, that's a good catch - do you know how to fix this? Would be a valuable contribution to handle this edge case.

DevEmilio96:
- Ok, got it !

- Yes I'll solve it, now I'll investigate the problem.

sunyitao:
<img width="867" alt="截屏2024-12-29 17 34 47" src="https://github.com/user-attachments/assets/7aaccd1e-222c-4caa-9744-b07d726517ee" />

Evanev7:
Should be fixed in 1.0!

PR #280, opened 2024-10-03
Error Message - Out of memory

DevEmilio96:
When there isn't enough memory for the model to calculate a response, exo used to display an "out of memory" error, but now that error only appears in the console and not in tinychat.

Issue: #275

### DIFF
--- a/exo/api/chatgpt_api.py
+++ b/exo/api/chatgpt_api.py
@@ -268,7 +268,13 @@ class ChatGPTAPI:
     callback = self.node.on_token.register(callback_id)
 
     if DEBUG >= 2: print(f"Sending prompt from ChatGPT api {request_id=} {shard=} {prompt=} {image_str=}")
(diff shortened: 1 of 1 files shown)


AlexCheema:
Oh I see what you mean now.

Isn't the `asyncio.gather` redundant here?
The point with this change is to "fire off" the inference without waiting for it to finish.
What we could do is to wrap the call in another async function which we then attach a callback to.

AlexCheema:
You're right about this silencing errors that should be shown to the user. I think the best solution might be to wrap the call in `asyncio.wait_for`

AlexCheema:
Good catch @DevEmilio96 I've fixed it here: https://github.com/exo-explore/exo/pull/282
Will close this one now.

AlexCheema:
Plese let me know if this fixes the issues you were running into.

PR #282, opened 2024-10-03
Unsilence errors

AlexCheema:


DevEmilio96:
I can now see the error messages, and I was about to push a change myself. If I use DEBUG=2, I can see the correct error message in the console as well.

PR #289, opened 2024-10-05
Use official nvidia-ml-py instead of pynvml

GaetanLepage:
From [pynvml's README](https://github.com/gpuopenanalytics/pynvml?tab=readme-ov-file):

> This package includes unofficial NVML bindings and utilities.
> This package is intended for demonstration purposes only. There
> is no guarantee for long-term maintenence or support.
>
> The official NVML bindings are published by NVIDIA under a different
> `nvidia-ml-py` project (see: https://pypi.org/project/nvidia-ml-py/).
>
> Future releases of this project will not include NVML bindings
> directly, but will instead depend on `nvidia-ml-py`.
>
> Please migrate to the official package for long-term support.
> You can still import and use a familar `pynvml` module after
> migrating to the official package.

[nvidia-ml-py](https://pypi.org/project/nvidia-ml-py/) should offer the same API.

cc @AlexCheema


AlexCheema:
Good catch - thank you!

Issue #267, opened 2024-10-02
 source install.sh   


MiguelSanzBr:
Installing build dependencies ... done
  Checking if build backend supports build_editable ... done
  Getting requirements to build editable ... done
  Preparing editable metadata (pyproject.toml) ... done
Collecting tinygrad@ git+https://github.com/tinygrad/tinygrad.git@232edcfd4f8b388807c64fb1817a7668ce27cbad
  Cloning https://github.com/tinygrad/tinygrad.git (to revision 232edcfd4f8b388807c64fb1817a7668ce27cbad) to /tmp/pip-install-qr_3zk_z/tinygrad_0ca3d675f951436db7369c46f5021624
  Running command git clone --filter=blob:none --quiet https://github.com/tinygrad/tinygrad.git /tmp/pip-install-qr_3zk_z/tinygrad_0ca3d675f951436db7369c46f5021624
  Running command git rev-parse -q --verify 'sha^232edcfd4f8b388807c64fb1817a7668ce27cbad'
  Running command git fetch -q https://github.com/tinygrad/tinygrad.git 232edcfd4f8b388807c64fb1817a7668ce27cbad
  Running command git checkout -q 232edcfd4f8b388807c64fb1817a7668ce27cbad
  Resolved https://github.com/tinygrad/tinygrad.git to commit 232edcfd4f8b388807c64fb1817a7668ce27cbad
  Preparing metadata (setup.py) ... done
Collecting tqdm==4.66.4
  Using cached tqdm-4.66.4-py3-none-any.whl (78 kB)
Collecting grpcio==1.64.1
  Using cached grpcio-1.64.1-cp310-cp310-manylinux_2_17_x86_64.manylinux2014_x86_64.whl (5.6 MB)
Collecting requests==2.32.3
  Using cached requests-2.32.3-py3-none-any.whl (64 kB)
Collecting netifaces==0.11.0
  Using cached netifaces-0.11.0.tar.gz (30 kB)
  Preparing metadata (setup.py) ... done
Collecting safetensors==0.4.3
  Using cached safetensors-0.4.3-cp310-cp310-manylinux_2_17_x86_64.manylinux2014_x86_64.whl (1.2 MB)
Collecting tenacity==9.0.0
  Using cached tenacity-9.0.0-py3-none-any.whl (28 kB)
Collecting prometheus-client==0.20.0
  Downloading prometheus_client-0.20.0-py3-none-any.whl (54 kB)
[...]

Issue #269, opened 2024-10-02
Many things to improve or fix


darkanubis0100:
- 3,000 and 1 bugs...

- Using CLANG is impossible because it directly brings up any infinity of bugs, under which version is that supposed to work without bugs? Because I'm using 18.1.3 on Ubuntu 24.04 and it's impossible to even use EXO without GPU because of CLANG bugs (the same one that has been shown many times in the repo issues).

- If CLANG already didn't work for some reason, with CUDA it's even worse, it doesn't support WSL or I have to use a physical Linux to run EXO? Because CUDA fails me with a weird error but if I disable bfloat16, directly the error is a “Segmentation Fault” after it asks me to install “llvmlite”.

- Detailed instructions and requirements in the repository? That doesn't exist here, I have to do magic trying to find out what programs or modules are missing to install and it still fails. Where is the complete list of dependencies? They don't even indicate that the “build-essential” is required.

- How am I supposed to run it on Android? I tried using Termux with an Ubuntu and ran into several errors, most notably Tailscale....

AlexCheema:
Thanks for being patient with us. exo is still highly experimental and there are indeed a lot of bugs that need to be fixed.

I think the general point here is the tinygrad inference engine isn't stable enough (no fault of tinygrad, it's a new library). We have other inference engines coming (PyTorch and llama.cpp). PyTorch is almost ready to merge: #139. I'm hoping since PyTorch is more mature and people are more familiar with it, it will be a lot more stable.

I will prioritise better instructions and requirements. The idea is that there shouldn't need to be much since it should "just work".

I've run on Android successfully with termux before we introduced tailscale. Is the tailscale dependency breaking it now?

darkanubis0100:
> Thanks for being patient with us. exo is still highly experimental and there are indeed a lot of bugs that need to be fixed.
> 
> I think the general point here is the tinygrad inference engine isn't stable enough (no fault of tinygrad, it's a new library that we've integrated into a highly experimental project). We have other inference engines coming (PyTorch and llama.cpp). PyTorch is almost ready to merge: #139. I'm hoping since PyTorch is more mature and people are more familiar with it, it will be a lot more stable.
> 
> I will prioritise better instructions and requirements. The idea is that there shouldn't need to be much since it should "just work".
> 
> I've run on Android successfully with termux before we introduced tailscale. Is the tailscale dependency breaking it now?

[...]

PR #292, opened 2024-10-05
replace tailscale.devices with good old http, removing the need for tailscale dependency

AlexCheema:
replace tailscale.devices with good old http, removing the need for tailscale dependency


AlexCheema:
fixes #267. also fixes the tailscale dependency issue mentioned in #269

Issue #243, opened 2024-09-28
Dynamic TFLOPS calculation


AlexCheema:
- A lot of people are confused when exo shows 0 TFLOPS because their device has not been explicitly added to the lookup.
- This is just a visual bug, but it causes a lot of confusion.

AlexCheema:
Added $200 retrospective bounty

Evanev7:
We're closing bounties, thank you to all who contributed, and apologies that we left this unmanaged for so long.

PR #297, opened 2024-10-06
Dynamic TFLOPS Calculation (In response to #243)

snico432:
I attempted to implement a dynamic TFLOPS calculation as a fallback in case the device is not found in the lookup table. I know that PyTorch is not yet a dependency for the project but I saw some discussion in #139 that it will be soon. Please forgive me if this PR is incorrectly formatted, it's my first time attempting to contribute to an open source project, but I was motivated to because I find this project really cool! Also, please let me know if there are any changes I should make to the code.


AlexCheema:
Yes!! I was hoping someone would do this. Great work! I haven't taken a proper look / tested yet, but I love to idea of dynamically calculating FLOPS based on a quick benchmark.

I added a $200 retrospective bounty for this which will be paid once merged

AlexCheema:
One suggestion: I don't want to force torch as a dependency. If we could lean on the existing `InferenceEngine` infrastructure we have, that would be great. Perhaps each `InferenceEngine` can implement a `benchmark` function?

AlexCheema:
Screenshot 2024-10-07 at 01 05 01

Getting this output now after 

```
exo --inference-engine pytorch --run-model llama-3.1-8b
```

snico432:
> Screenshot 2024-10-07 at 01 05 01
> Getting this output now after
> 
> ```
> exo --inference-engine pytorch --run-model llama-3.1-8b
> ```

I'm assuming this comment was meant for #139 ?

AlexCheema:
> > Screenshot 2024-10-07 at 01 05 01
> > Getting this output now after
> > ```
> > exo --inference-engine pytorch --run-model llama-3.1-8b
> > ```
> 
> I'm assuming this comment was meant for #139 ?

you're right. sorry ignore that :)

AlexCheema:
Just ping me when you want me to review the PR again.

snico432:
@AlexCheema I added support for benchmarking on the MLX inference engine. Right now it only benchmarks f32 and f16 calculations because mlx doesn't support matrix multiplication for int8. Not sure how I should proceed with that. Please let me know if this is what you're looking, or if you'd like me to make some changes. I'll move on to Tinygrad afterwards.

exo/topology/mlx_benchmark.py
AlexCheema, line 1: why is this in a separate file in `topology`. can we move it into `inference/mlx`?

AlexCheema, line 1: why is this in a separate file in topology. can we move it into inference/mlx?

exo/inference/inference_engine.py
```diff
@@ -15,6 +15,10 @@ async def infer_prompt(self, request_id: str, shard: Shard, prompt: str, image_s
   async def infer_tensor(self, request_id: str, shard: Shard, input_data: np.ndarray, inference_state: Optional[str] = None) -> Tuple[np.ndarray, str, bool]:
     pass
 
+  @abstractmethod
+  def benchmark_tflops(self):
```
AlexCheema, line 19: This is going to need to be async otherwise it will block the main thread

exo/main.py
```diff
@@ -74,6 +75,7 @@
   for chatgpt_api_endpoint in chatgpt_api_endpoints:
     print(f" - {terminal_link(chatgpt_api_endpoint)}")
 
+device_capabilities = device_capabilities(inference_engine)
```
AlexCheema, line 78: I'd prefer a design where `device_capabilities` is asynchronous and called from the discovery module. That way we don't block exo from getting started and the discovery module can fetch the device capabilities when it needs to. Also, if it fails for whatever reason it can be retried. For now lets cache the device capabilities in the discovery module forever but eventually we'd want to "refresh" it by running the benchmark again

exo/topology/mlx_benchmark.py
snico432: Sorry, left it there when I was testing. Fixed in my latest commit.

exo/inference/inference_engine.py
snico432: Fixed!

exo/main.py
snico432: I think I've now got it cached correctly in the discovery modules (this is my first time working with the async library so I'm not 100% sure)

snico432:
@AlexCheema I created a benchmark for tinygrad, cleaned up the mlx benchmark and attempted to implement your requests from your last review. I also removed the dict from device_capabilites and made all the TFLOPS get calculated dynamically.

AlexCheema (changes requested):
I merged latest main and made some changes here and there.

I think this is mostly good. Computing DeviceCapabilities is very slow, so I'd prefer if we only compute this once, ideally lazily (i.e. it doesn't slow down startup).

If you can fix that, it should be good to merge.

snico432:
@AlexCheema DeviceCapabilites are now lazily computed. PTAL

AlexCheema:
Please fix merge conflicts.

snico432:
@AlexCheema resolved merge conflicts.

snico432:
@AlexCheema PTAL

snico432:
@AlexCheema PTAL

JakeHillion:

PR #326, opened 2024-10-10
also initialize embed_tokens if last layer and tie_word_embeddings true

AlexCheema:


AlexCheema:
This fixes llama 3.2 1b when running over multiple devices.

Issue #302, opened 2024-10-07
[BOUNTY - $1500] Package exo as installable(s)


AlexCheema:
### Background

Right now exo requires installing python, installing packages with `pip` and then running the `exo` command. This is too difficult for people who are non-technical. A lot of the people who want to use exo are non-technical.

### Proposed Solution

To make exo easier to use for non-technical people, I'd like to simplify and make a single installable that users can install. This can be one installable or multiple installables for each target platform (e.g. Linux, MacOS, Windows). This needs to be as frictionless as possible, installing all prerequisites for the user from a fresh installation of their OS.

### Deliverables:

- A single installable or multiple installables (one per platform) that can be installed with one click on each target platform (Linux, MacOS, Windows)
- The installable(s) need to handle edge cases like what version of MacOS you're running on or whether you have CUDA installed. Everything needs to happen in one installable.
- A circleci CI/CD pipeline that creates the installable(s) automatically on commit

SethBurkart123:
I'll have a look into doing this. If it was packaged as an application.. we also need to figure out how it should act, does there need to be interfaces designed? do we keep it minimal similar to ollama, if so how will the user deal with logs? 

For now i'll just look at making it easily installable and runnable with everything packaged together nicely into a simple executable but that could be a point of future discussion.

tomy128:
> I also had same thoughts about the installation process



> Quote reply
> Reference in n


maybe only an executable file is required? like .exe on windows. Cause the exo is a B/S application? 
for that，this issue aim to build a installable app maybe with pyinstaller/py2app(or others) with  all prerequisites exo required ?
👀

Bashorun97:
Is this still up since there's already a PR in progress?

larson-carter:
I think that this is a still an issue that would be beneficial. 

I have some ideas, but would anyone like to take this on with me?

AlexCheema:
I've assigned this to @SethBurkart123 as they seem to be furthest along with making this work.

AlexCheema:
just to be fully transparent, there seem to be two almost complete PR's for this: #354 and #334.

in the interest of fairness and rewarding good work, the one that gets merged will receive the $1500 bounty. the other will receive $750.

#332 will also receive $200 for experimenting on the circleci part.

luckyai490:
rer

AlexCheema:
This is complete.

PR #332, opened 2024-10-11
feat: add CircleCI job to create installables and script for standalone executable

PandelisZ:
NGL I did not run this or check much about what its doing. But it looks like its on the right tracks.

I was curious if Genie could tackle it after seeing this tweet: https://x.com/Edi_Danalache/status/1844613560565104985

Maybe someone can take it over the line: Tackles: #302

😄

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -186,6 +186,22 @@ jobs:
       - checkout
       - run: system_profiler SPHardwareDataType
 
+  create_installables:
+    docker:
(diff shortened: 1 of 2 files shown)


AlexCheema:
This looks good, however this is the "easy" part.
The tricky part is packaging exo as an installable in the first place and ensuring it works across all platforms.
I'll keep this open as it might be useful later when someone gets this working.

PR #337, opened 2024-10-12
Tinygrad Inference Router

reddyn12:
Add logic to be able to support building different model types in tinygrad inference

### DIFF
--- a/exo/inference/tinygrad/inference.py
+++ b/exo/inference/tinygrad/inference.py
@@ -27,7 +27,7 @@ MODEL_PARAMS = {
 }
 
 
-def build_transformer(model_path: Path, shard: Shard, model_size="8B", device=None):
+def build_llama(model_path: Path, shard: Shard, model_size="8B", device=None):
   # build model
   linear = nn.Linear
   with Context(THREEFRY=0):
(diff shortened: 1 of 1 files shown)


reddyn12:
I can move the ```build_llama``` into ```llama.py``` and rename the function to ```build_model```. This will make the ```if/else and raise``` not necessary. lmk how I should approach this

exo/inference/tinygrad/inference.py
```diff
@@ -27,7 +27,7 @@
 }
 
 
-def build_transformer(model_path: Path, shard: Shard, model_size="8B", device=None):
+def build_llama(model_path: Path, shard: Shard, model_size="8B", device=None):
```
dtnewman, line 30: why change this to `build_llama`?

reddyn12: Every model will have its own build function. So build_llava, build_llama, build_mamba, etc...

dtnewman: ah, i missed that you changed it below. These AsyncIO calls are not quite as reader friendly 😄

reddyn12:
@AlexCheema is this proposal fine, or is there a better way to do this?

AlexCheema:
Ideally we do this like #139 where we support models automatically without needing separate `build_{model}` functions.

reddyn12:
Ok, I'll close this pr, do the LLaVa tinygrad implementation pr with this kind of logic, and then put up another pr that refactors the inference logic to used the stateful_shards.

PR #338, opened 2024-10-12
Fix TFLOPS on 4060 Ti - cabelo@opensuse.org

cabelo:
Before PR
after

After PR
before


exo/topology/device_capabilities.py
```diff
@@ -71,7 +71,7 @@ def to_dict(self):
   "NVIDIA GEFORCE RTX 4070 TI": DeviceFlops(fp32=39.43*TFLOPS, fp16=78.86*TFLOPS, int8=157.72*TFLOPS),
   "NVIDIA GEFORCE RTX 4070 SUPER": DeviceFlops(fp32=30.0*TFLOPS, fp16=60.0*TFLOPS, int8=120.0*TFLOPS),
   "NVIDIA GEFORCE RTX 4070": DeviceFlops(fp32=29.0*TFLOPS, fp16=58.0*TFLOPS, int8=116.0*TFLOPS),
-  "NVIDIA GEFORCE RTX 4060 TI 16GB": DeviceFlops(fp32=22.0*TFLOPS, fp16=44.0*TFLOPS, int8=88.0*TFLOPS),
```
AlexCheema, line 74: can we keep this one too? then the guy who added this for their device will complain and we'll end up in an infinite loop 😆

AlexCheema (changes requested):

exo/topology/device_capabilities.py
cabelo: I understand, doing...

cabelo: @AlexCheema Thanks for the explanation, and sorry for the confusion. Done!

PR #342, opened 2024-10-14
benchmark for different inference engines

AlexCheema:


exo/main.py
```diff
@@ -25,10 +25,11 @@
 from exo.orchestration.node import Node
 from exo.models import model_base_shards
 from exo.viz.topology_viz import TopologyViz
+import yappi
```
dtnewman, line 28: this needs to be added to setup.py, no?

exo/main.py
```diff
@@ -48,9 +49,11 @@
 parser.add_argument("--inference-engine", type=str, default=None, help="Inference engine to use (mlx, tinygrad, or dummy)")
 parser.add_argument("--disable-tui", action=argparse.BooleanOptionalAction, help="Disable TUI")
 parser.add_argument("--run-model", type=str, help="Specify a model to run directly")
+parser.add_argument("--bench", action=argparse.BooleanOptionalAction, help="Benchmark the model")
```
dtnewman, line 52: I find this a bit confusing. If I'm a user looking to benchmark the model, I look here and I see two ways of going about it (either replace `run` with `bench` or pass in a `--bench` arg). Which one am I supposed to choose? Looking at the code below, it seems like the practical difference is whether you initialize the API (which doesn't happen if I include `--bench` but does happen with `run->bench`, but this isn't clear from the code up top here.

AlexCheema:
Outdated.

Issue #343, opened 2024-10-14
Incorrect FLOPS Display with GPU Names Including VRAM Size


Sean-fn:
FLOPS display fails when the GPU name includes VRAM size (e.g., `NVIDIA RTX A2000 12GB`), leading to incorrect formatting.

PR #344, opened 2024-10-14
Fix gpu capabilities display

Sean-fn:
This PR addresses #343 , the FLOPS display issue by splitting and deleting the VRAM size info from the GPU name.
In my case, the device name is `NVIDIA RTX A2000 12GB`, so it didn't show the info correctly.

**Before**
CleanShot 2024-10-14 at 23 24 59@2x

**After**
CleanShot 2024-10-14 at 23 25 21@2x

Also updated the info for RTX 2080 TI.


AlexCheema:
That makes sense.
Thank you.

PR #339, opened 2024-10-13
Display download progress for all nodes in tinychat

Sean-fn:
This PR addresses issue #312 by:

1. Showing download progress for all nodes instead of just the first one.
2. Updating the UI to display per-node information.
3. Enhancing error handling for failed downloads.

The gif below is a demo of the expected result using a fake generated data.
CleanShot

PR #349, opened 2024-10-15
Fix null download progress bug

Sean-fn:
This PR addresses a bug introduced in my previous PR #339 . 
It ensures that the download status is set to null once all downloads have finished, allowing the status indicator to disappear as expected.

Changes:
- Set download status to null after all downloads are complete


AlexCheema:
Thank you!

PR #354, opened 2024-10-16
Package exo as installable

josh1593:


AlexCheema:
For reference, since we talked on X. Here are the issues I encountered:

- link to `exo-logo.icns` is wrong
- python3.11 is hardcoded
- get this error (not sure how to fix -- we do want static linking, not dynamic linking): `FATAL: Automatic detection of static libpython failed. Homebrew Python is unexpectedly broken. Disable with '--static-libpython=no' if you don't want to install it.`

AlexCheema:
Still getting this error:

```
(.venv) $ python3 build/build_exo.py 
Nuitka-Options: Used command line options: exo/main.py --company-name=exolabs --output-dir=dist --follow-imports --standalone --output-filename=exo
Nuitka-Options: --macos-app-name=exo --macos-app-mode=gui --macos-app-version=0.0.1 --product-name=exo --macos-create-app-bundle
Nuitka-Options: --macos-app-icon=docs/exo-logo.icns --include-distribution-meta=pygments --include-distribution-meta=mlx --include-module=mlx._reprlib_fix
Nuitka-Options: --include-module=mlx._os_warning --include-data-files=./.venv/lib/python3.12/site-packages/mlx/lib/mlx.metallib=mlx/lib/mlx.metallib
Nuitka-Options: --include-data-dir=exo/tinychat=tinychat --include-module=exo.inference.mlx.models.llama
Nuitka-Options: --include-module=exo.inference.mlx.models.deepseek_v2 --include-module=exo.inference.mlx.models.base
Nuitka-Options: --include-module=exo.inference.mlx.models.llava --include-module=exo.inference.mlx.models.qwen2
FATAL: Automatic detection of static libpython failed. Homebrew Python is unexpectedly broken. Disable with '--static-libpython=no' if you don't want to install it.
An error occurred: Command '['python3', '-m', 'nuitka', 'exo/main.py', '--company-name=exolabs', '--output-dir=dist', '--follow-imports', '--standalone', '--output-filename=exo', '--macos-app-name=exo', '--macos-app-mode=gui', '--macos-app-version=0.0.1', '--product-name=exo', '--macos-create-app-bundle', '--macos-app-icon=docs/exo-logo.icns', '--include-distribution-meta=pygments', '--include-distribution-meta=mlx', '--include-module=mlx._reprlib_fix', '--include-module=mlx._os_warning', '--include-data-files=./.venv/lib/python3.12/site-packages/mlx/lib/mlx.metallib=mlx/lib/mlx.metallib', '--include-data-dir=exo/tinychat=tinychat', '--include-module=exo.inference.mlx.models.llama', '--include-module=exo.inference.mlx.models.deepseek_v2', '--include-module=exo.inference.mlx.models.base', '--include-module=exo.inference.mlx.models.llava', '--include-module=exo.inference.mlx.models.qwen2']' returned non-zero exit status 1.
```

josh1593:
https://github.com/Nuitka/Nuitka/issues/1353

exo/inference/tinygrad/models/llama.py
```diff
@@ -1,7 +1,10 @@
 from typing import Tuple, Union, Optional, Dict, Any
 from tinygrad import Tensor, Variable, TinyJit, dtypes, nn, Device
 from tinygrad.helpers import getenv
-
+import logging
+import os
+import torch
```
aiden-dnewman, line 6: i think torch needs to be added to the project dependencies

buildexo/build_exo.py
```diff
@@ -0,0 +1,69 @@
...
+            "--macos-app-icon=docs/exo-logo.icns",
+            "--include-module=exo.inference.mlx.models.llama",
+            "--include-module=exo.inference.mlx.models.deepseek_v2",
+            "--include-module=exo.inference.mlx.models.base",
+            "--include-module=exo.inference.mlx.models.llava",
+            "--include-module=exo.inference.mlx.models.qwen2",
+            "--include-distribution-meta=mlx",
+            "--include-module=mlx._reprlib_fix",
+            "--include-module=mlx._os_warning",
+            f"--include-data-files=./env/lib/{python_version}/site-packages/mlx/lib/mlx.metallib=mlx/lib/mlx.metallib",
```
aiden-dnewman, line 34: i think this is making assumptions about where your virtual env is located. I think maybe this should use site packages. For example:

```
site_packages = site.getsitepackages()[0]
...

and then replace `env` in the f-string with `{site_packages}`
```

(code above is paraphrased, not tested)

.circleci/config.yml
```diff
@@ -216,4 +329,6 @@ workflows:
       - discovery_integration_test
       - chatgpt_api_integration_test_mlx
       - test_macos_m1
+      - build_exo_macos
+      - build_exo_linux
       # - chatgpt_api_integration_test_tinygrad
```
dtnewman, line 334: why not `build_exo_windows`?

exo/inference/tinygrad/models/llama.py
AlexCheema: does the bf16 fix work without the torch dependency?

josh1593: ah i need to remove import torch line, i was using torch tensors to fix bf16 error but that required a lot of changes to other files

the bf16 fix does not work without the torch deps,  i believe upstream changes in tinygrad's library is needed

dtnewman: but torch _is_ used below in this module, no?

buildexo/build_exo.py
josh1593: correct. its using the site modules now. thanks for catching that!

exo/inference/tinygrad/models/llama.py
josh1593: my bad, I have now removed changes I made to the fix_bf16 function. it no longer uses torch tensors

dtnewman:
A few high-level comments on this:

1) re: windows, I agree there are issues with windows more generally, but you made some small, useful changes here in `main.py` that (at least on my windows machine) get the program to start up, even if inference doesn't work. I agree that it's out of scope to get this actually running inference on windows but i think it should still be building it regardless, so that you generate an EXE file, and exo starts running (even if the actual inference doesn't work).

2) To that end, if I run command `exo` on my windows machine against the code on your branch, it starts up. But when i click the executable that build_exo.py generates, it doesn't run. It took an unusually long time to compile on my machine, but it did eventually return an executable. But when i click to run it, it opens a terminal and then shuts down pretty much immediately. Were you able to get this working?

3) The circleci process runs `build_exo.py`, but what happens to the artifacts after that? Ideally, I think you'd want to generated artifacts generate github "releases" (perhaps zip them first though)

4) When I'm testing this on my m-chip mac, where everything "just works" for exo more generally, it creates a .app file that i can just click to run. But it runs a background process, and once a user closes the tab in their browser, the app is still running. If I then click on it a second time, it's still running. I think that _ideal_ behavior would be that it runs with a little icon showing it as a service (ideally, with a small menu when you click the service icon that allows you to quit it). If you click the .app file a second time, it would just detect that it's already running and shut down.

I'm happy to clarify on any of the above, and I also have windows + linux + mac to test things if you need (although no GPU, so the first two don't work very well without being on the [torch PR](https://github.com/exo-explore/exo/pull/139)).

.circleci/config.yml
AlexCheema: do you know which dependencies are problematic right now?
tinygrad supports windows fully now so we should be able to get exo working on windows with little effort

exo/inference/tinygrad/models/llama.py
AlexCheema: does this fix the issue though? I think what you had looked like a potential fix for the bf16 issue others have ran into on CPU.

josh1593: no it does not fix the bf16 casting error,

AlexCheema: outside the scope of this PR. lets fix in a separate PR.

buildexo/build_exo.py
AlexCheema: thanks for catching that @aiden-dnewman

AlexCheema:
> A few high-level comments on this:
> 
> 1. re: windows, I agree there are issues with windows more generally, but you made some small, useful changes here in `main.py` that (at least on my windows machine) get the program to start up, even if inference doesn't work. I agree that it's out of scope to get this actually running inference on windows but i think it should still be building it regardless, so that you generate an EXE file, and exo starts running (even if the actual inference doesn't work).
> 2. To that end, if I run command `exo` on my windows machine against the code on your branch, it starts up. But when i click the executable that build_exo.py generates, it doesn't run. It took an unusually long time to compile on my machine, but it did eventually return an executable. But when i click to run it, it opens a terminal and then shuts down pretty much immediately. Were you able to get this working?
> 3. The circleci process runs `build_exo.py`, but what happens to the artifacts after that? Ideally, I think you'd want to generated artifacts generate github "releases" (perhaps zip them first though)
> 4. When I'm testing this on my m-chip mac, where everything "just works" for exo more generally, it creates a .app file that i can just click to run. But it runs a background process, and once a user closes the tab in their browser, the app is still running. If I then click on it a second time, it's still running. I think that _ideal_ behavior would be that it runs with a little icon showing it as a service (ideally, with a small menu when you click the service icon that allows you to quit it). If you click the .app file a second time, it would just detect that it's already running and shut down.
> 
> I'm happy to clarify on any of the above, and I also have windows + linux + mac to test things if you need (although no GPU, so the first two don't work very well without being on the [torch PR](https://github.com/exo-explore/exo/pull/139)).

Agree with these points.
We can drop Windows support, so no need to fix 1 and 2.
3 and 4 are required for this to be merged. @josh1593

exo/main.py
```diff
@@ -25,6 +28,8 @@
 from exo.orchestration.node import Node
 from exo.models import model_base_shards
 from exo.viz.topology_viz import TopologyViz
+import os
```
dtnewman, line 31: nitpick, but this should go up near the other non-exo imports

exo/main.py
```diff
@@ -201,16 +206,33 @@ async def run_model_cli(node: Node, inference_engine: InferenceEngine, model_nam
   finally:
     node.on_token.deregister(callback_id)
 
+def open_web_chat():
+  if web_chat_urls:
+    import subprocess
+    import os
```
dtnewman, line 212: this is already imported at the top and i think subprocess should be too

PR #368, opened 2024-10-21
Manual networking with configuration files

ianpaul10:
Ref: https://github.com/exo-explore/exo/issues/279

This PR includes the following changes:

* adds `manual` option to `--discovery-module` command flag
* adds a new `--discovery-config-path` command flag to point to a json config file outlining the expected network config
* `NetworkTopology` python data class to parse the config file 
* `ManualDiscovery` discovery class to connect to other nodes in the network and perform health checks.
* Added tests for NetworkTopology & ManualDiscovery (tried to follow the pattern of the other Discovery classes for tests, not sure if that's sufficient or not)


exo/main.py
```diff
@@ -78,6 +80,10 @@
   discovery = UDPDiscovery(args.node_id, args.node_port, args.listen_port, args.broadcast_port, lambda peer_id, address, device_capabilities: GRPCPeerHandle(peer_id, address, device_capabilities), discovery_timeout=args.discovery_timeout)
 elif args.discovery_module == "tailscale":
   discovery = TailscaleDiscovery(args.node_id, args.node_port, lambda peer_id, address, device_capabilities: GRPCPeerHandle(peer_id, address, device_capabilities), discovery_timeout=args.discovery_timeout, tailscale_api_key=args.tailscale_api_key, tailnet=args.tailnet_name)
+elif args.discovery_module == "manual":
+  if not args.discovery_config_path:
+    raise ValueError(f"--discovery-config-path is required when using manual discovery. Please provide a path to a config json file.")
```
ianpaul10, line 85: Here, and throughout this PR I raise exceptions when something in the config isn't correct. I don't see a lot of other situations where errors are being thrown back to the user elsewhere in the codebase, but I'm not sure how else to handle this.

How do you think these sorts of errors should be handled? One alternative is to just ignore them and assume no peers.

exo/networking/manual/manual_discovery.py
```diff
@@ -0,0 +1,74 @@
...
+    if DEBUG_DISCOVERY >= 2: print("Starting task to find peers from config...")
+    while True:
+      for peer_id, peer_config in self.peers_in_network.items():
+        if DEBUG_DISCOVERY >= 2: print(f"Checking peer {peer_id=} at {peer_config.address}:{peer_config.port}")
+        peer = self.known_peers.get(peer_id)
+        if not peer:
+          if DEBUG_DISCOVERY >= 2: print(f"{peer_id=} not found in known peers. Adding.")
+          new_peer_handle = self.create_peer_handle(peer_id, f"{peer_config.address}:{peer_config.port}", peer_config.device_capabilities)
+          self.known_peers[peer_id] = new_peer_handle
+          peer = new_peer_handle
```
ianpaul10, line 64: Should this addition of peers just be done inside the `discover_peers` function?

exo/networking/manual/network_topology_config.py
```diff
@@ -0,0 +1,48 @@
...
+class NetworkTopology:
+  """Configuration of the network. A collection outlining all nodes in the network, including the node this is running from."""
+
+  peers: Dict[str, PeerConfig]
+  """
+  node_id to PeerConfig. The node_id is used to identify the peer in the discovery process. The node that this is running from should be included in this dict.
+  """
+
+  @classmethod
+  def from_path(cls, path: str) -> "NetworkTopology":
```
ianpaul10, line 26: This could be simplified drastically if you're ok with including an additional dependency on pydantic https://docs.pydantic.dev/latest/examples/files/ (and we could get json + toml + yml for free)

exo/networking/manual/test_manual_discovery.py
```diff
@@ -0,0 +1,103 @@
+import asyncio
```
ianpaul10, line 1: Not sure what sort of structure you're trying to follow for unittests, so I just tried to follow suit with UDP & tailscale's tests. LMK if there are other things you think I should test or include

exo/networking/manual/manual_discovery.py
AlexCheema: It's fine here. few changes:
- make sure we're health checking them the peers and only add healthy peers
- try catch the body of the loop, if this fails once the loop should continue
- sleep in the `finally` block of the try

exo/networking/manual/network_topology_config.py
AlexCheema: lets do this. fine with having dependency on `pydantic` for this

exo/networking/manual/test_manual_discovery.py
AlexCheema: all good.

AlexCheema (approved):
clean PR, thanks for making these comments and asking for clarification. few changes needed. lfg

AlexCheema (changes requested):
accidentally approved the last one. requesting changes.

exo/main.py
AlexCheema: This is good.

ianpaul10:
Thanks for the review! I believe I addressed all of the comments, so it should be ready for re-review whenever you get a chance

AlexCheema:
lgtm

AlexCheema:
Merged.
Please email alex@exolabs.net with your ethereum address for your $200 bounty in USDC.

PR #373, opened 2024-10-22
Getting the setup script to work on Intel based Mac machines

willschneider15:
When trying to run the setup script on an Intel-based Mac I ran into the following issue installing MLX:

Screenshot 2024-10-22 at 1 54 33 AM

To fix the issue I updated the logic to check that they are using Arm64 in addition to Darwin before installing the MLX packages.

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -2,12 +2,4 @@
 
 python3 -m venv .venv
 source .venv/bin/activate
-
-# Check if running on macOS with Apple Silicon
-if [[ "$(uname)" == "Darwin" && "$(uname -m)" == "arm64" ]]; then
-    echo "Installing with Apple Silicon support..."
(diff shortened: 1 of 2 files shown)


install.sh
AlexCheema, line 1: Is it possible to put this logic in `setup.py` since it's not required to use `install.sh`

AlexCheema (changes requested):

install.sh
willschneider15: Yep I just pushed the update

willschneider15: Tested on my Intel Mac.

If possible, it would be a good idea to test the setup on an M series (Apple Silicon) chip before merging, to ensure that MLX installs correctly.

AlexCheema:
Awesome. LGTM!

Issue #372, opened 2024-10-21
Auto selection of inference engine


AlexCheema:
Right now, inference engines are not interoperable e.g. you can't run one machine on `MLX` and another on `tinygrad` - they need to both be using the same inference engine.

We should automatically handle the selection of the inference engine. We can still provide a manual override with `--inference-engine` but the default should be automatically detecting the best inference engine that is compatible with all devices in the network.

rltakashige:
Going to close as the tracking issue will resolve this.

PR #377, opened 2024-10-23
Autoselection of inference engines 

josh1593:
Issue #372 

@AlexCheema


exo/orchestration/standard_node.py
```diff
@@ -346,6 +385,7 @@ async def periodic_topology_collection(self, interval: int):
         if DEBUG >= 2: print(f"{did_peers_change=}")
         if did_peers_change:
           await self.collect_topology()
+          await self.select_best_inference_engine()
       except Exception as e:
         print(f"Error collecting topology: {e}")
```
AlexCheema, line 390: I guess this should say "Error collecting topology or selecting best inference engine"

AlexCheema:
Just one small comment so merging anyway and we can fix that later

Issue #380, opened 2024-10-23
[BOUNTY - $100] Support changing manual configuration while running


AlexCheema:
#368 + #381 added support for manual configuration but if you update the configuration file while exo is running it won't update.
This should be a simple fix, hence $100 bounty.

josh1593:
@AlexCheema working on this one

AranavMahalpure:
Hi @AlexCheema,

I’d like to help with this issue. I have a solution to enable dynamic configuration updates while the application (exo) is running, using a hot-reload approach to detect and apply changes in the configuration file.
Could you confirm if this approach aligns with your expectations? I’d also like to ask about the bounty requirements to ensure I’m meeting them fully. Would demonstrating this solution meet the criteria for the $100 bounty?
Thanks!

AranavMahalpure:
Hi @AlexCheema,

Thank you for the clarification! A periodic check every 5 seconds sounds straightforward, and I can implement that approach. I’ll set up a watcher to check the configuration file for any changes every 5 seconds and reload it if an update is detected.

Just to confirm—would this basic 5-second polling approach fulfill the bounty requirements, or is there any additional functionality you’d like me to include?

Thanks again for the guidance, and looking forward to getting started!

AranavMahalpure:
Hi @AlexCheema,

I wanted to follow up on my last message regarding the 5-second polling approach for detecting configuration changes. I’m ready to get started on implementing this solution, but I just wanted to confirm if it fully meets the bounty requirements, or if there’s any additional functionality you’d like me to include.

Let me know if you have any further guidance or adjustments. Looking forward to moving forward with this!

Thanks again,
@AranavMahalpure

Evanev7:
We're closing bounties, thank you to all who contributed, and apologies that we left this unmanaged for so long.

PR #383, opened 2024-10-24
Support changing manual configuration while running

ianpaul10:
Ref #380 

* Load config at the start of each `find` task
* Only load config and return the new values if the file's `getmtime` has been updated, otherwise return the cached peers
* If anything fails when trying to read the config file, return the cached peers (or empty dict if it's never been loaded)
* Add new cleanup peers task that removes peers from the `known_peers` list if they no-longer exist in the dict but still exist in the `known_peers` list
* Test that starts with 2 peers in the config file, adds a 3rd to the config, checks that it gets picked up by a node, and then remove it from the config, and check that it gets forgotten by the node.


exo/networking/manual/manual_discovery.py
```diff
@@ -67,3 +61,18 @@ async def task_find_peers_from_config(self):
...
+        topology = NetworkTopology.from_path(self.network_config_path)
+
+        if self.node_id not in topology.peers:
+          raise ValueError(f"Node ID {self.node_id} not found in network config file {self.network_config_path}. Please run with `node_id` set to one of the keys in the config file: {[k for k, _ in topology.peers]}")
+
+        peers_in_network: Dict[str, PeerConfig] = topology.peers
+        peers_in_network.pop(self.node_id)
+    except Exception as e:
+        if DEBUG_DISCOVERY >= 2: print(f"Error when loading network config file from {self.network_config_path}. Please update the config file in order to successfully discover peers. Exception: {e}")
+        peers_in_network = {}
```
ianpaul10, line 75: The only gotcha I found with handling dynamic updates to the config file is if someone is manually changing the json (instead of pasting in a fully new & correct json blob) it could be loaded/read in a WIP/incorrect state.

So my solution to that is to just swallow any exception we get when reading from the config and return an empty dict (so there will be nothing for `task_find_peers_from_config` to iterate over and load until the config file is set correctly)

exo/networking/manual/manual_discovery.py
```diff
@@ -42,17 +38,30 @@ async def discover_peers(self, wait_for_peers: int = 0) -> List[PeerHandle]:
...
+      peers_from_config = self._get_peers().items()
+      if peers_from_config:
+        peers_to_remove = [peer for peer in self.known_peers.keys() if peer not in peers_from_config]
+
+        for peer in peers_to_remove:
+          if DEBUG_DISCOVERY >= 2: print(f"{peer} is no longer found in the config. Removing...")
+          try: del self.known_peers[peer]
+          except KeyError: pass
+
+      await asyncio.sleep(5.0)
```
ianpaul10, line 53: I think a cleanup of "stale" peers (one that has been added to the know_peers list but is no longer in the config file) is now necessary.

Related to my other comment around handling the WIP state of the config file, only thing I'm unsure of it how long it should sleep for. I feel like the delay on looking for stale peers should be longer than how often we try and find/add new peers, but I'm not 100% sure.

exo/networking/manual/manual_discovery.py
```diff
@@ -62,8 +75,35 @@ async def task_find_peers_from_config(self):
...
+      peers_in_network.pop(self.node_id)
+
+      self._cached_peers = peers_in_network
+      self._last_modified_time = current_mtime
+
+      return peers_in_network
+
+    except Exception as e:
+      if DEBUG_DISCOVERY >= 2: print(f"Error when loading network config file from {self.network_config_path}. Please update the config file in order to successfully discover peers. Exception: {e}")
+      return self._cached_peers
```
ianpaul10, line 109: The only gotcha I found with handling dynamic updates to the config file is if someone saves a WIP/invalid json (instead of pasting in a fully new & correct json blob).

So the current solution is to just swallow any exception we get when reading from the config and return the cached config (until the config file is set correctly)

exo/networking/manual/test_manual_discovery.py
```diff
@@ -12,17 +13,20 @@
...
   async def asyncSetUp(self):
-    self.peer1 = mock.AsyncMock()
+    self.peer1 = mock.AsyncMock(spec=Node)
     self.peer1.connect = mock.AsyncMock()
-    self.discovery1 = ManualDiscovery(root_path, "node1", create_peer_handle=lambda peer_id, address, device_capabilities: self.peer1)
-    _ = self.discovery1.start()
+    self.server1 = GRPCServer(self.peer1, "localhost", 8000)
+    await self.server1.start()
+    self.discovery1 = ManualDiscovery(root_path, "node1", create_peer_handle=lambda peer_id, address, device_capabilities: GRPCPeerHandle(peer_id, address, device_capabilities))
+    await self.discovery1.start()
```
ianpaul10, line 21: Fixes a runtime warning that showed up when I made the logs more detailed.

ianpaul10:
@AlexCheema This PR should address the follow-up to updating the manual config while exo is running.

exo/networking/manual/manual_discovery.py
```diff
@@ -69,3 +68,45 @@ async def task_find_peers_from_config(self):
...
+        for peer in peers_to_remove:
+          if DEBUG_DISCOVERY >= 2: print(f"{peer} is no longer found in the config but is currently a known peer. Removing from known peers...")
+          try: del self.known_peers[peer]
+          except KeyError: pass
+
+      await asyncio.sleep(1.0)
+
+  def _get_peers(self):
+    try:
+      current_mtime = os.path.getmtime(self.network_config_path)
```
AlexCheema, line 88: Isn't this blocking?
Blocking every 1 second is going to slow exo down significantly since it runs on one thread.

Perhaps we can move these operations to a thread pool executor?

exo/networking/manual/manual_discovery.py
```diff
@@ -69,3 +68,45 @@ async def task_find_peers_from_config(self):
...
+
+      if self._cached_peers is not None and self._last_modified_time is not None and current_mtime <= self._last_modified_time:
+        return self._cached_peers
+
+      topology = NetworkTopology.from_path(self.network_config_path)
```
AlexCheema, line 93: Is `from_path` blocking? Same comment as above. Can we make all I/O operations to `self.network_config_path` happen on a thread pool executor with one thread?

exo/networking/manual/manual_discovery.py
AlexCheema: That's a fine solution.

AlexCheema (changes requested):
Need it to be asyncified. Then it's good to go.

exo/networking/manual/manual_discovery.py
ianpaul10: Yes, good catch. I updated all calls to `self.network_config_path` to run within a single thread pool executor. I think things should now be non-blocking and thread safe

ianpaul10:
@AlexCheema Thanks for the notes, reading from the config file should now be non-blocking

exo/networking/manual/manual_discovery.py
```diff
@@ -60,12 +64,65 @@ async def task_find_peers_from_config(self):
...
+    while True:
+      peers_from_config = await self._get_peers()
+      if peers_from_config:
+        peers_to_remove = [peer for peer in self.known_peers.keys() if peer not in peers_from_config]
+
+        for peer in peers_to_remove:
+          if DEBUG_DISCOVERY >= 2: print(f"{peer} is no longer found in the config but is currently a known peer. Removing from known peers...")
+          self.known_peers.pop(peer, None)
+
+      await asyncio.sleep(1.0)
```
AlexCheema, line 85: I'm not sure why this is needed. I think this may be overcomplicating it.

The way `StandardNode` is set up, it should be possible to simply return the latest set of peers, i.e. your `_get_peers` is enough.

AlexCheema (changes requested):

exo/networking/manual/manual_discovery.py
ianpaul10: Ah I see, TIL a bit more about StandardNode, thanks!

I thought it was necessary to handle cleaning up nodes that were originally added to the config file (and thus added to `self.known_peers`) but removed afterwards.

exo/networking/manual/manual_discovery.py
```diff
@@ -57,15 +59,54 @@ async def task_find_peers_from_config(self):
...
         except Exception as e:
           if DEBUG_DISCOVERY >= 2: print(f"Exception occured when attempting to add {peer_id=}: {e}")
+      self.known_peers = new_known_peers
       await asyncio.sleep(1.0)
 
       if DEBUG_DISCOVERY >= 2: print(f"Current known peers: {[peer.id() for peer in self.known_peers.values()]}")
+
+  async def _get_peers(self):
+    try:
+      async with self._lock:
```
AlexCheema, line 73: Is the lock even needed now? I only see one place where `_get_peers` is called

AlexCheema (changes requested):

exo/networking/manual/manual_discovery.py
ianpaul10: No longer needed, you're right. Removed.

ianpaul10:
@AlexCheema Bump. I think this should be ready to merge now that the lock is removed and simplified.

AlexCheema (approved):

AlexCheema:
Merged!
Certainly worthy of a $200 retrospective bounty. Please email alex@exolabs.net to claim the bounty!

PR #396, opened 2024-10-28
Update helpers.py

FFAMax:
Tmp folder permission may mislead root cause showing issue with binding socket

```
Detected system: Linux
Inference engine name after selection: tinygrad
get_inference_engine called with: tinygrad
Using inference engine: TinygradDynamicShardInferenceEngine with shard downloader: HFShardDownloader
Trying to find available port host='0.0.0.0' port=61382
Unable write to file
  File "/home/user/exo/exo/helpers.py", line 62, in find_available_port
    write_used_port(port, used_ports)
  File "/home/user/exo/exo/helpers.py", line 47, in write_used_port
    with open(used_ports_file, "w") as f:
         ^^^^^^^^^^^^^^^^^^^^^^^^^^
PermissionError: [Errno 13] Permission denied: '/tmp/exo_used_ports'

  File "/home/user/exo/.venv/bin/exo", line 5, in <module>
    from exo.main import run
  File "/home/user/exo/exo/main.py", line 72, in <module>
    args.node_port = find_available_port(args.node_host)
                     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/user/exo/exo/helpers.py", line 66, in find_available_port
    raise RuntimeError (e)
RuntimeError: [Errno 13] Permission denied: '/tmp/exo_used_ports'
```


exo/helpers.py
```diff
@@ -54,11 +54,15 @@ def write_used_port(port, used_ports):
...
-    if DEBUG >= 2: print(f"Trying to find available port {port=}")
+    if DEBUG >= 2: print(f"Trying to bind port {port=} on address {host=}")
     try:
       with socket.socket(socket.AF_INET, socket.SOCK_STREAM) as s:
         s.bind((host, port))
-      write_used_port(port, used_ports)
+      try:
+        write_used_port(port, used_ports)
+      except Exception as e:
+        if DEBUG >= 2: print(f"Unable write to file")
```
dtnewman, line 64: I'd make this a bit more helpful of an error message, for example "Unable to write to file using the write_used_port function" to make this more debuggable.

dtnewman:
I'm curious what the impetus here is. What system are you running this on where you are having trouble writing to /tmp? And do you have any idea why?

dtnewman:
> I'm curious what the impetus here is. What system are you running this on where you are having trouble writing to /tmp? And do you have any idea why?

in any case, I have a [different PR](https://github.com/exo-explore/exo/pull/401/files#diff-b203aae941b21636a6f6f3e878e4aeb4ae92ccdf7a527cc6446565c8f5a38934) which has a bunch of changes, but one of them (which can possibly be separated out) is a "config.py" file where I setup a mechanism for persistent storage. Instead of using /tmp, I write settings to a "user data directory" (e.g. `~/.local/share/exo` on linux). Anything currently written to tmp probably belongs there instead and can easily be added. But I'm concerned that whatever issue is coming up for you here with `tmp` directory permissions might come up there as well.

FFAMax:
> I'm curious what the impetus here is. What system are you running this on where you are having trouble writing to /tmp? And do you have any idea why?

It is linux. Issue is:
User A run exo and file created.
User B run exo and cannot create the same file due already exist and no perms to use it.

(Debian)

FFAMax:
> > I'm curious what the impetus here is. What system are you running this on where you are having trouble writing to /tmp? And do you have any idea why?
> 
> in any case, I have a [different PR](https://github.com/exo-explore/exo/pull/401/files#diff-b203aae941b21636a6f6f3e878e4aeb4ae92ccdf7a527cc6446565c8f5a38934) which has a bunch of changes, but one of them (which can possibly be separated out) is a "config.py" file where I setup a mechanism for persistent storage. Instead of using /tmp, I write settings to a "user data directory" (e.g. `~/.local/share/exo` on linux). Anything currently written to tmp probably belongs there instead and can easily be added. But I'm concerned that whatever issue is coming up for you here with `tmp` directory permissions might come up there as well.

That a good idea to organize storage. That issue with /tmp is weird because I do not expect any setting stick global session (like only one process can be running on the system). For example that file with ports written into it related tp particular PID and should be good keep it in temp as it is temporary for each run, but add PID into the name and delete wile once program closed (even crashed - ideally).

AlexCheema:
Not sure how this helps.

AlexCheema:
I'm still not sure exactly how this helps. It reraises the exception so doesn't it just end up with the same behaviour with a slightly different print?

FFAMax:
@AlexCheema this to help in troubleshooting.
Instead of saying 'Unable open a socket' will point to more specific issue related to the file permissions. This will not resolve the root cause but help save time on troubleshooting: instead of checking why socket can not be opened, it will point to the permissions with a file.
But as we discussed with Daniel Newman, better refactor this portion to move away from using file /tmp/exo_used_ports. At least I didn't found why we need this file in first place.

Evanev7:

Issue #103, opened 2024-07-30
Make the default model an even smaller model


AlexCheema:
The default model should be something smaller that can be downloaded quicker.
We'd also want to make https://github.com/exo-explore/exo/issues/16 work to make the download even quicker.

rltakashige:
EXO 1.0 does not have a default model - please raise an issue if this would be a helpful feature.

PR #399, opened 2024-10-29
[tidy] smallest model first

thenatlog:
**Make the default model an even smaller model #103**  

- removes select logic which incorrectly selected the third element in the model list as default


AlexCheema (approved):
Please fix indentation to 2 space indent then I will review

AlexCheema (changes requested):
Please fix indentation to 2 space indent then I will review

exo/tinychat/index.html
```diff
@@ -29,8 +29,8 @@
     <div x-show="errorMessage" x-transition.opacity x-text="errorMessage" class="toast">
     </div>
 <div class="model-selector">
-<select @change="if (cstate) cstate.selectedModel = $event.target.value" x-model="cstate.selectedModel">
```
AlexCheema, line 32: don't we still need this to set the selectedModel

exo/tinychat/index.html
```diff
@@ -29,8 +29,8 @@
...
-<option selected="" value="llama-3.2-1b">Llama 3.2 1B</option>
+<select>
+<option value="llama-3.2-1b">Llama 3.2 1B</option>
```
AlexCheema, line 33: shouldn't this option have a `selected` property so that it's by default selected

AlexCheema (changes requested):

exo/tinychat/index.html
thenatlog: the first item in the list is selected by default in <select> tags

exo/tinychat/index.html
AlexCheema: @thenatlog

PR #400, opened 2024-10-29
Fix formatting

AlexCheema:
@thenatlog hopefully this simplifies things. now we only use yapf for formatting. no linting. instructions in README.


thenatlog (approved):
LGTM

Issue #378, opened 2024-10-23
[BOUNTY - $100] Support Llama 3.2 1B on tinygrad


AlexCheema:
- Currently we support Llama 3.2 1B on MLX but not tinygrad
- Add support for Llama 3.2 1B
- Might just work out of the box, if not I think the issue will be in the changes that were made to RoPE (Rotary Position Embedding) from llama 3.1 to 3.2
- You'll also need to add an entry to `models.py`

Sanchay-T:
Hi @AlexCheema,

I’d love to work on adding support for Llama 3.2 1B in tinygrad.  

Thanks!
Sanchay

AlexCheema:
> Hi @AlexCheema,
> 
> I’d love to work on adding support for Llama 3.2 1B in tinygrad.
> 
> Thanks! Sanchay

Go for it!

Sanchay-T:
Hey @AlexCheema ! Thanks for assigning me this issue. I'm new to this codebase and trying to understand how everything works before diving into implementing Llama 3.2 support for tinygrad.

I've spent some time reading through the code, and here's what I understand so far (please correct me if I'm wrong anywhere!):

When someone sends a message in the chat, it starts from the frontend in `index.html` where Alpine.js handles the UI:

```javascript
async processMessage(value) {
    const response = await fetch("/v1/chat/completions", {
      method: "POST",
      body: JSON.stringify({
        model: this.cstate.selectedModel,  // This is where we specify Llama 3.2
        messages: this.cstate.messages,
        stream: true,
      }),
    });
```

This message then goes through the router in the backend:

```python
@router.post("/v1/chat/completions")
async def chat_completions(request: Request):
    body = await request.json()
    model_name = body.get("model", "llama-3.1-8b")
```

From there, it gets sent to either the MLX or tinygrad implementation. I see that MLX already supports Llama 3.2 1B, but tinygrad needs to be updated. Looking at `llama.py` in tinygrad, I think the main changes needed are in the RoPE (Rotary Position Embedding) implementation:

```python
def precompute_freqs_cis(dim: int, end: int, theta: float = 10000.0, dtype=dtypes.half):
    freqs = 1.0/(theta**(Tensor.arange(0, dim, 2)[:(dim // 2)]/dim))
    # This part might need updating for 3.2
```

I'm thinking of approaching this implementation in the following way:

[...]

PR #403, opened 2024-10-31
tidy: use smallest model with ChatGPT API

maujim:
Use the smallest possible model as the fallback in chatGPT api #103 

We don't support llama-3.2-1b on tinygrad backend which could be annoying for Linux users. However, we still merged this change for the tinychat frontend (#399) so I think it's ok. 

If its available, I will work on that feature next since its a good beginner bounty (#378)

### DIFF
--- a/exo/api/chatgpt_api.py
+++ b/exo/api/chatgpt_api.py
@@ -231,7 +231,7 @@ class ChatGPTAPI:
     stream = data.get("stream", False)
     chat_request = parse_chat_request(data, self.default_model)
     if chat_request.model and chat_request.model.startswith("gpt-"):  # to be compatible with ChatGPT tools, point all gpt- model requests to llama instead
-      chat_request.model = self.default_model if self.default_model.startswith("llama") else "llama-3.1-8b"
+      chat_request.model = self.default_model if self.default_model.startswith("llama") else "llama-3.2-1b"
     if not chat_request.model or chat_request.model not in model_base_shards:
       if DEBUG >= 1: print(f"Invalid model: {chat_request.model}. Supported: {list(model_base_shards.keys())}. Defaulting to {self.default_model}")
       chat_request.model = self.default_model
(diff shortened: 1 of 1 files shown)


exo/api/chatgpt_api.py
```diff
@@ -227,12 +229,12 @@ async def handle_post_chat_completions(self, request):
     data = await request.json()
     if DEBUG >= 2: print(f"Handling chat completions request from {request.remote}: {data}")
     stream = data.get("stream", False)
-    chat_request = parse_chat_request(data)
+    chat_request = parse_chat_request(data, self.default_model)
     if chat_request.model and chat_request.model.startswith("gpt-"):  # to be compatible with ChatGPT tools, point all gpt- model requests to llama instead
-      chat_request.model = "llama-3.1-8b"
+      chat_request.model = self.default_model if self.default_model.startswith("llama") else "llama-3.2-1b"
```
maujim, line 234: Not sure if we should have a `self.default_llama_model` as well or just hardcoded for this fallback is ok.

exo/api/chatgpt_api.py
```diff
@@ -163,6 +163,8 @@ def __init__(self, node: Node, inference_engine_classname: str, response_timeout
     self.prompts: PrefixDict[str, PromptSession] = PrefixDict()
     self.prev_token_lens: Dict[str, int] = {}
     self.stream_tasks: Dict[str, asyncio.Task] = {}
+    self.default_model = "llama-3.2-1b"
```
AlexCheema, line 166: no need to add this. just change the string `llama-3.1-8b` to `llama-3.2-1b`. Is there any reason we'd want this as a class variable

AlexCheema (changes requested):

exo/api/chatgpt_api.py
maujim: 1. One place to change the default_model. Notably, we do the "default model check" twice, once in the function that handles the endpoint, and again in the `parse_chat_request`
2. Also, I thought it could be easier for testing with different default models.

Issue #305, opened 2024-10-07
exo stops downloading when a chatgpt api request times out


AlexCheema:
- This was introduced recently and is very annoying as now you need to keep requesting to keep it downloading
- The best fix here would be to make it so during downloading the `process_prompt` task is shielded but after download is complete, it can be cancelled
- For now it's fine to shield the entire `process_prompt`

AlexCheema:
follow up: #307

PR #407, opened 2024-11-03
Make process_prompt Cancellable Outside Downloads

### LINKED ISSUE #307: [BOUNTY - $200] make it so process_prompt is cancellable outside of a download (follow up to #305)
See #305 

Currently the fix is to shield `process_prompt` however this also makes it so that if a chatgpt api request times out, it never gets cancelled even when a download is not in progress.

This can lead to behaviour where exo keeps generating even after a chatgpt api request has timed out.

**Gotchas:** make sure however exo works right is actually cancellable end-to-end. If an inference is in progress, that needs to be cancellable - InferenceEngines need to support this behaviour.

rahat2134:
This PR implements protection for downloads at their source in HFShardDownloader, ensuring downloads complete regardless of request cancellation. This is a follow-up fix to #306, moving from blanket request shielding to targeted download protection.

### Changes Made
1. Modified `HFShardDownloader.ensure_shard` to:
   - Protect downloads at their source
   - Handle cancellations gracefully
   - Properly manage concurrent downloads
2. Added comprehensive tests in `test_hf_shard_download.py`
3. Removed request-level protection from ChatGPTAPI

### Implementation Details
```python
async def ensure_shard(self, shard: Shard) -> Path:
    # ... existing download handling ...
    try:
        return await download_task
    except asyncio.CancelledError:
        # Continue download despite cancellation
        await event.wait()
        return result
```
- Downloads protected where they actually happen
- Clean handling of concurrent requests
- Proper resource cleanup
- Robust error handling

### Test Coverage
Added comprehensive tests in `test_hf_shard_download.py`:
1. `test_download_protection`: Verifies downloads complete despite cancellation
2. `test_multiple_downloads`: Ensures concurrent downloads work correctly
3. `test_download_error_handling`: Validates proper error handling

### Testing
```bash
test/test_hf_shard_download.py::test_download_protection PASSED
test/test_hf_shard_download.py::test_multiple_downloads PASSED
test/test_hf_shard_download.py::test_download_error_handling PASSED
```

### Breaking Changes
None. This improves download protection while maintaining backward compatibility.

### Checklist
- [x] Tests added for all scenarios
- [x] Download protection moved to source
- [x] Concurrent downloads handled properly
- [x] Resource cleanup implemented
- [x] Previous approach removed cleanly

### Related Issues
- Fixes #307
- Follow-up to #306
- Original issue #305

**Note to Reviewer:** Per your feedback, I've moved the protection to where downloads actually happen in HFShardDownloader, rather than trying to protect at the request level. This ensures downloads complete regardless of what triggers them (process_prompt or any other request). The solution is focused and thoroughly tested.


AlexCheema:
I like the idea, the code is well written and the tests are extensive.

However, this doesn't quite solve the problem of preventing downloads from being cancelled -- a download can be started in `process_prompt` or could be started by another request. I think the idea is correct -- we should prevent downloads from being cancelled but this isn't quite how to do it. Let me know what you think.

AlexCheema:
> > I like the idea, the code is well written and the tests are extensive.
> > However, this doesn't quite solve the problem of preventing downloads from being cancelled -- a download can be started in `process_prompt` or could be started by another request. I think the idea is correct -- we should prevent downloads from being cancelled but this isn't quite how to do it. Let me know what you think.
> 
> @AlexCheema Thank you for the feedback! You're absolutely right about the limitation. I see now that checking downloads only at the start of process_prompt doesn't fully protect downloads, since:
> 
> 1. Downloads could start after the check
> 2. Downloads could be triggered by other requests
> 3. We need protection at the download level, not the request level.
> 
> Would it be better to:
> 
> 1. Move the protection closer to the download operation itself in HFShardDownloader?
> 2. Track active downloads more granularly?
> 
> Could you guide me on which approach would be most appropriate? I can then update the PR accordingly.

Yes I think your suggestions are good. I'm not sure I have a full solution in my head right now I'd need to look more deeply. If you want to take a look and suggest something or make more changes in this PR go ahead

rahat2134:
@AlexCheema PTAL.

rahat2134:
@AlexCheema ptal!

Evanev7:

PR #413, opened 2024-11-04
Quantized models support tinygrad

RashikShahjahan:


AlexCheema (changes requested):
I don't think this does quite what it should do.
It looks like this takes an existing model and quantizes it.
The bounty is for supporting already quantized models.

PR #418, opened 2024-11-08
Llama-3.2 tinygrad Support (1B & 3B models)

BatSmacker84:
This PR adds support for the `llama-3.2-1b` and `llama-3.2-3b` models when using the tinygrad inference engine.
This PR should also close issue #378.

# Changes/Additions

The models used for tinygrad inference are not from the official Meta repo due to it requiring authentication. Instead, I have chosen models from "unsloth", which they claim are the exact same as the original Meta models, but without needing authentication to download. I also specifically went for the "Instruct" versions of the models, as that is what seems to be preferred based on the other models.

The architecture for Llama-3.2 is similar to other Llama3 versions, so the model parameters for the 1B and 3B model sizes have been added alongside the 8B and 70B parameters. The model sizes are then parsed outside of the function call before being passed in for better readability.

The only major change compared to previous models is the rope scaling factor and tie word embeddings. Two new parameters have been added to accommodate this: `rope_scaling` and `tie_word_embeddings`. Both are ignored by default, so behavior of other models should in theory be unchanged. I am not entirely sure what specific effects these two parameters have on the model, but without them and the respective code all outputs from the models are complete nonsense tokens.

# Testing

I have verified myself that both the 1B and 3B versions of Llama-3.2 function properly under tinygrad with the new changes. However, I have not been able to verify if prior versions of Llama3 still function as expected under tinygrad. This is due to any models above 3B parameters (ie. `llama3.1-8b`) not fitting on my GPU. My RX 7800XT has 16GB of VRAM, so in theory it should be able to handle 8B models no problem. This behavior was observed before my changes, so it is not being caused by my changes. I am not sure what is causing this, but it unfortunately is preventing me from testing other models to verify that they still work. Like I said before, in theory my changes should not impact other models because they are ignored by default.

# About the Bounty...

I am not certain on how bounties are treated. If this is considered "sniping" the bounty or something similar, then I apologize. I was trying out the project and encountered the issue I described above, so I needed to be able to run smaller models with tinygrad. There hasn't been any noticeable activity on the issue over the past few days from what I can tell, but I could be missing something. In any case, I made these changes because I wanted to improve the project to better suit my needs, not because I wanted to quickly grab a bounty before someone else did.

### DIFF
--- a/exo/inference/tinygrad/inference.py
+++ b/exo/inference/tinygrad/inference.py
@@ -22,6 +22,8 @@ TOP_P = 0.9
 ALPHA_F = 0.1
 ALPHA_P = 0.0
 MODEL_PARAMS = {
+  "1B": {"args": {"dim": 2048, "n_heads": 32, "n_kv_heads": 8, "n_layers": 16, "norm_eps": 1e-5, "rope_theta": 500000, "vocab_size": 128256, "hidden_dim": 8192, "rope_scaling": {"factor": 32.0, "high_freq_factor": 4.0, "low_freq_factor": 1.0, "original_max_position_embeddings": 8192, "rope_type": "llama3"}, "tie_word_embeddings": True}, "files": 1},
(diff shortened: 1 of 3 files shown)


AlexCheema:
Great work.
Send me your ethereum address to alex@exolabs.net to claim your bounty.

PR #422, opened 2024-11-10
Added a clear all history button

samiamjidkhan:
Why:
- Users have to delete chat histories one by one
- Time-consuming and tedious for users with many chat sessions
- No bulk delete option exists

What:
- Added a centrally positioned button below 'tinychat'
- Button shows only when histories exist
- Added a confirmation dialog to prevent accidental deletions
- Style consistent with existing UI

### DIFF
--- a/exo/inference/tinygrad/inference.py
+++ b/exo/inference/tinygrad/inference.py
@@ -55,7 +55,7 @@ class TinygradDynamicShardInferenceEngine(InferenceEngine):
     self.shard_downloader = shard_downloader
     self.executor = ThreadPoolExecutor(max_workers=1)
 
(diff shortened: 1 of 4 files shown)


samiamjidkhan:
ezgif com-crop


> Why:
> 
> * Users have to delete chat histories one by one
> * Time-consuming and tedious for users with many chat sessions
> * No bulk delete option exists
> 
> What:
> 
> * Added a centrally positioned button below 'tinychat'
> * Button shows only when histories exist
> * Added a confirmation dialog to prevent accidental deletions
> * Style consistent with existing UI

samiamjidkhan:
There was a little bug in `inference.py` which is also fixed.

AlexCheema:
Really great work!
Insta merge.

PR #433, opened 2024-11-12
Enabled inference engine intercompatibility

blindcrone:
If inference engines can run the same model, nodes on those inference engines can now interoperate

### DIFF
--- a/exo/download/shard_download.py
+++ b/exo/download/shard_download.py
@@ -8,7 +8,7 @@ from exo.helpers import AsyncCallbackSystem
 
 class ShardDownloader(ABC):
   @abstractmethod
-  async def ensure_shard(self, shard: Shard) -> Path:
+  async def ensure_shard(self, shard: Shard, inference_engine_name: str) -> Path:
     """
         Ensures that the shard is downloaded.
         Does not allow multiple overlapping downloads at once.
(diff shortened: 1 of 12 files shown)


exo/download/shard_download.py
```diff
@@ -27,7 +27,7 @@ def on_progress(self) -> AsyncCallbackSystem[str, Tuple[Shard, RepoProgressEvent
 
 
 class NoopShardDownloader(ShardDownloader):
-  async def ensure_shard(self, shard: Shard) -> Path:
+  async def ensure_shard(self, shard: Shard, inference_engine_name: str) -> Path:
```
dtnewman, line 30: what's the intention of adding `inference_engine_name` here?

blindcrone: Necessary because of the change to the abstractmethod signature

dtnewman: Im assuming this is for some future where different inference engines save the model in different ways. Whatever the reason, maybe add it to the docstring for the abstract method `ensure_shard` above?

dtnewman: OK... sorry, I see what it's doing now. On first pass I didn't notice that the class in hf_shard_download.py inherits from the parent in this file :)

AlexCheema:
This is fire

PR #447, opened 2024-11-12
Error message timeout removed, added close button

cadenmackenzie:
I was having difficulty reading the error messages when the timeout was set to 5 seconds so I removed the timeout and added a close button instead.

I also added the error name and stack trace to be shown in the error box so that the user has more information when trying to debug.

### DIFF
--- a/exo/tinychat/index.css
+++ b/exo/tinychat/index.css
@@ -202,6 +202,22 @@ main {
     top: 0; /* Position at the top of the page */
     left: 0; /* Extend from the left edge */
     right: 0; /* Extend to the right edge */
+    display: flex;
+    justify-content: space-between;
+    align-items: center;
+}
+
+.toast-close-button {
+    background: none;
+    border: none;
+    color: white;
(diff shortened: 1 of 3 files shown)


AlexCheema (changes requested):
Can we do both? 30 sec timeout and also you can close it.

cadenmackenzie:
Yes, I like that. I added the timeout back with 30 seconds. It could even be 20 seconds. Do you like having the stack trace included or does that make it too cluttered?

AlexCheema:
> Yes, I like that. I added the timeout back with 30 seconds. It could even be 20 seconds. Do you like having the stack trace included or does that make it too cluttered?

Maybe a "Show Details" button that expands the stack trace?

cadenmackenzie:
Good idea. 

Made the requested changes. See loom for quick [demo](https://www.loom.com/share/6772c7fb38c34334be788d1503a9b0a2?sid=dddb5960-32b3-4492-97d6-01b2a2ffba0a).

PR #448, opened 2024-11-13
Added a small script to compile grpc

blindcrone:


AlexCheema:
Makes a lot of sense. No more DM's about how to gen grpc :)

dtnewman:
> Makes a lot of sense. No more DM's about how to gen grpc :)

Do not worry. You will most definitely still get DMs about how to gen grpc 😁

PR #449, opened 2024-11-13
Remove inference state

blindcrone:
This requires caching start_pos in StatefulModel for Tinygrad... for now


AlexCheema:
Woo! That `inference_state` stuff always bugged me.

PR #462, opened 2024-11-15
Refactors to simplify messaging and properly batch inputs

blindcrone:
This simplifies and refactors some of the orchestration messaging, done in service of enabling evaluation and training

Currently this isn't quite ready for primetime as it seems to break exo run (but not exo)

### DIFF
--- a/exo/orchestration/standard_node.py
+++ b/exo/orchestration/standard_node.py
@@ -102,11 +102,7 @@ class StandardNode(Node):
   def get_topology_inference_engines(self) -> List[List[str]]:
     return self.topology_inference_engines_pool
   
-  async def encode_prompt(self, shard: Shard, prompt):
-    toks = await self.inference_engine.encode(shard, prompt)
-    return toks
-  
-  async def process_result(
+  async def process_inference_result(
     self,
     shard,
     result: np.ndarray,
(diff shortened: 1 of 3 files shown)


exo/orchestration/standard_node.py
```diff
@@ -255,46 +243,57 @@ async def _process_tensor(
...
-      else:
-        target_peer = next((p for p in self.peers if p.id() == target_id), None)
-        if not target_peer:
-          raise ValueError(f"Peer for {next_partition_index} not found")
-        if is_tensor:
-          if DEBUG >= 1: print(f"Sending tensor to {target_peer.id()}: {tensor_or_prompt}")
-          await target_peer.send_tensor(next_shard, tensor_or_prompt, request_id=request_id)
-        else:
-          await target_peer.send_prompt(next_shard, tensor_or_prompt, request_id=request_id)
+    if DEBUG >= 1: print(f"target partition index: {target_index}")
```
dtnewman, line 260: I do see some clairity in having `forward_prompt` and `forward_tensor` broken out and more explicitly named, but what about:

```
  async def forward_prompt(self, base_shard: Shard, prompt: str, request_id: str) -> None:
    await self._forward_prompt_or_tensor_helper(base_shard, prompt, request_id)

  async def forward_tensor(self, base_shard: Shard, tensor: np.ndarray, request_id: str) -> None:
    await self._forward_prompt_or_tensor_helper(base_shard, tensor, request_id)
```

and then renaming the current function `_forward_prompt_or_tensor_helper`?

blindcrone: Most of the intention in this refactor in general is clarifying the code paths so it's easier to extend code that passes through it, which is damn near everything since this is the glue between network, inference, and application code. Removing the ambiguous dispatch situations is the whole reason I did it

AlexCheema: Seems good

AlexCheema (approved):
great work

PR #463, opened 2024-11-15
Eval working

blindcrone:
Added facilities for processing examples (which currently consist of an input, a target, and a length) and a means of evaluating them against a (currently hard-defaulted) loss function in a distributed fashion across the shards of the exo network

A lot of the piping here will make it easier to do distributed training.


exo/main.py
```diff
@@ -221,16 +267,36 @@ def handle_exit():
       print("Error: Model name is required when using 'run' command or --run-model")
       return
     await run_model_cli(node, inference_engine, model_name, args.prompt)
+  elif args.command == "eval" or args.command == 'train':
+    data_path = args.data
+    train, val, test = load_dataset(data_path)
+    model_name = args.model_name
+    if args.command == 'eval':
+      if not model_name:
+        print("Error: Much like a human, I can't evaluate anything without a model")
```
dtnewman, line 276: I do love this line, but maybe add in something like `Please pass in a parameter for model_name (e.g. "exo eval llama-3.2-3b")`  so that users don't have to parse through the code in main.py to get this working (replace with a working example command if I didn't get that right)

exo/main.py
```diff
@@ -221,16 +267,36 @@ def handle_exit():
...
+        return
+      await eval_model_cli(node, inference_engine, model_name, test, args.batch_size)
+    else:
+      if not model_name:
+        print("Error: This train ain't leaving the station without a model")
```
dtnewman, line 281: again, an actionable command would be helpful here (at the end! I'm not saying don't keep this part of the message 😄 )

exo/networking/grpc/grpc_server.py
```diff
@@ -69,19 +69,40 @@ async def SendTensor(self, request, context):
...
+      n_layers=request.shard.n_layers,
     )
+    loss = np.frombuffer(request.loss.tensor_data, dtype=np.dtype(request.loss.dtype)).reshape(request.loss.shape)
+    request_id = request.request_id
 
+    if shard.is_first_layer():
+      asyncself.node.backward_loss(shard, loss, request_id)
+    if DEBUG >= 5: print(f"SendTensor tensor {shard=} {example=} {target=} {length=} {request_id=} result: {result}")
+    return loss
+    return node_service_pb2.Tensor(tensor_data=tensor_data, shape=result.shape, dtype=str(result.dtype)) if result is not None else node_service_pb2.Tensor()
```
dtnewman, line 104: two returns here ⚠️

blindcrone: O, that's probably an error rebasing introduced, lol

dtnewman:
@blindcrone I put in some comments that I hope are useful. Note that these are all somewhat stylistic/superficial since I havent actually fetched or tested the branch on my machine.

exo/orchestration/standard_node.py
```diff
@@ -190,15 +180,108 @@ async def _process_prompt(self, base_shard: Shard, prompt: str, request_id: Opti
...
+    target: np.ndarray, 
+    length: np.ndarray,
+    request_id: Optional[str] = None,
+    train: bool = False,
+  ):
+    shard = self.get_current_shard(base_shard)
+    if shard.is_first_layer():
+      resp = await self.process_example(shard, example, target, length, request_id)
+    else:
+      if request_id is None:
```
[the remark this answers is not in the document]
blindcrone: That happens further down in the various process_ functions and so is unnecessary here. Basically this whole enqueue_ thing exists because unlike prompts, working with tensors doesn't have an implicit way of forwarding to the first layer, but we still want the originator of a request to be able to track its id

exo/main.py
blindcrone: I appreciate the sentiment of writing good error messages but this is still very much a work in progress

blindcrone:
Okay, this now in theory trains across nodes on MLX

PR #468, opened 2024-11-18
Download progress box height increase

cadenmackenzie:
Each time something was downloading I would go to scroll so I could see the full progress. I added a minimum height so the user can see the full download info. Also added some padding so it had some breathing room. See before and after images.

Before:
Screenshot 2024-11-17 at 5 51 27 PM

After:
Screenshot 2024-11-17 at 5 56 44 PM

### DIFF
--- a/exo/tinychat/index.css
+++ b/exo/tinychat/index.css
@@ -167,6 +167,8 @@ main {
 .download-progress {
   margin-bottom: 12em;
   overflow-y: auto;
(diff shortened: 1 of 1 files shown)


AlexCheema:
Thanks

KayleyK1820 (commented):
be522b98b149f54c8d3be78fdbd98f185a3ac062

PR #470, opened 2024-11-18
package exo as installable

josh1593:
#302


exo/main.py
```diff
@@ -3,7 +3,11 @@
 import signal
 import json
 import logging
+import platform
+import os
+import sys
 import time
+import subprocess
```
dtnewman, line 10: I don't think `subprocess` gets used

exo/api/chatgpt_api.py
```diff
@@ -174,13 +181,25 @@ def __init__(self, node: Node, inference_engine_classname: str, response_timeout
...
+      self.static_dir = Path(__file__).parent.parent/"tinychat"
+      self.app.router.add_get("/", self.handle_root)
+      self.app.router.add_static("/", self.static_dir, name="static")
 
     self.app.middlewares.append(self.timeout_middleware)
     self.app.middlewares.append(self.log_request)
+  
+  async def handle_quit(self, request):
+    print("Received quit signal")
+    from exo.main import shutdown
```
dtnewman, line 196: now that `shutdown` is used in multiple places, I (personally) think it'd be more intuitive for it to live in helpers.py

exo/main.py
```diff
@@ -127,6 +131,11 @@
   lambda req_id, tokens, __: topology_viz.update_prompt_output(req_id, inference_engine.tokenizer.decode(tokens)) if topology_viz and hasattr(inference_engine, "tokenizer") else None
 )
 
+if not args.model_seed_dir is None:
+  try:
+    await move_models_to_hf()
```
AlexCheema, line 136: Pass in `args.model_seed_dir`

exo/download/hf/hf_helpers.py
```diff
@@ -99,9 +100,18 @@ async def get_auth_headers():
 
 def get_repo_root(repo_id: str) -> Path:
   """Get the root directory for a given repo ID in the Hugging Face cache."""
-  sanitized_repo_id = repo_id.replace("/", "--")
+  sanitized_repo_id = str(repo_id).replace("/", "--")
   return get_hf_home()/"hub"/f"models--{sanitized_repo_id}"
 
+async def move_models_to_hf():
```
AlexCheema, line 106: Needs to take an argument

AlexCheema (changes requested):

exo/download/hf/hf_helpers.py
```diff
@@ -99,9 +100,18 @@ async def get_auth_headers():
...
 
+async def move_models_to_hf(seed_dir: Union[str, Path]):
+  """Move model in resources folder of app to .cache/huggingface/hub"""
+  source_dir = Path(seed_dir)
+  dest_dir = get_hf_home()/"hub"
+  await aios.makedirs(dest_dir, exist_ok=True)
+  for path in source_dir.iterdir():
+    if path.is_dir() and path.startswith("models--"):
+      dest_path = dest_dir / path.name
+      shutil.move(str(path), str(dest_path))
```
AlexCheema, line 114: any way to make this async?

exo/main.py
```diff
@@ -127,6 +131,15 @@
   lambda req_id, tokens, __: topology_viz.update_prompt_output(req_id, inference_engine.tokenizer.decode(tokens)) if topology_viz and hasattr(inference_engine, "tokenizer") else None
 )
 
+if not args.models_seed_dir is None:
+  try:
+    if is_frozen():
+      seed_dir = Path(sys.argv[0]).parent
+      await move_models_to_hf(seed_dir)
```
AlexCheema, line 138: no need for special case.

AlexCheema (changes requested):

AlexCheema:
LGTM

OKHand-Zy:
> this is addressed on pr #473

Hi, I'd like to know if the change made on line 186 of exo/inference/mlx/sharded_utils.py is crucial.
```Python
tokenizer = load_tokenizer(model_path, tokenizer_config) => tokenizer = await resolve_tokenizer(model_path)
```
When I try to support local models, I encounter an issue, but it's resolved after reverting the change. So I'm wondering if this change is really necessary. If it's not that important, I'd prefer to use the old method.

OKHand-Zy:
> This change is correct. We should keep using `resolve_tokenizer` as it's async. We should not have sync blocking I/O code.
> 
> @OKHand-Zy you will need to fix your code to work with `resolve_tokenizer`

Okay, I'll try to modify my code to make `resolve_tokenizer` work properly.

PR #472, opened 2024-11-19
Update some versions to support Python >= 3.9 and fix tinygrad thread issues

AlexCheema:
- Updated some version to work with Python >= 3.9. Tested with Python 3.9,3.10,3.11,3.12,3.13 and works
- tinygrad was occasionally giving errors related to sqlite code running on different threads. I made it now so everything runs on the same thread. It's tricky because even a line of code like `tensor.numpy()` could run tinygrad code because it's all lazy evaluated


AlexCheema:
LGTM 👍

Issue #165, opened 2024-08-22
Enhancement: support local and custom models


vlbosch:
At the moment, exo only supports loading models that are in the huggingface cache-folder. In case you have your own quants of a given model, you'd need to create a folder under the HF-cache that follows the conventions of HF-downloader (like for example: "~/.cache/huggingface/hub/models--mlx-community--Meta-Llama-3.1-8B-Instruct-4bit/refs/main") with a dummy commit sha hash. 

Would be great if exo could support loading models from any given folder, to support local and custom models. Thanks!

Sanchay-T:
Hi @vlbosch, @leo4life2,

I would love to take on this issue and work on supporting local and custom models in Exo! Please assign it to me.

Also, feel free to check out my portfolio at [sanchay.me](https://sanchay.me/) to learn more about my background and expertise.

I've also commented on two other issues that I’m interested in. If those get assigned elsewhere, I'd still love to work on this issue. Looking forward to contributing!

Best regards,
Sanchay-T

AlexCheema:
Assigned @Sanchay-T

PR #475, opened 2024-11-20
Support local model with inference-engine mlx

OKHand-Zy:
[Enhancement: support local and custom models #165](https://github.com/exo-explore/exo/issues/165)

This is a modified version of the code that I've made functional, although the code quality may not be ideal. It supports running local path models using mlx for both CLI and ChatAPI. However, I've encountered an issue where the response quality from the CLI is subpar. I would appreciate any suggestions on how to improve or optimize the code for better results.

## Changes:
- Added a "How to use local models" section to the README.
- Added the --add-local-model args.
- Implemented init_exo_env to configure local model cards and the local model store.
- Added bypass logic for local models using if...else statements.


OKHand-Zy:
I understand that the automatic instantiation of a default directory structure could potentially lead to inconsistencies across various systems and environments. However, isn't this approach similar to how HF downloads function?

OKHand-Zy:
New Updates:

*   Automatic broadcasting of local model downloads is now supported.
*   Added `--stored-model-ip` and `--stored-model-port` arguments.
*   CLI and ChatAPI now support running local models.
*   Added a 'storedhost' directory within the download directory for local model code.

Note: This feature only supports the `--inference-engine mlx` option.

exo/helpers.py
DreamsOfControl, line 1: I think at Line 34 get_system_info should be get_mac_system_info as that is the name of the variable in the main.py file.

OKHand-Zy: I agree.

Evanev7:

PR #480, opened 2024-11-21
Train working

blindcrone:
Currently this trains on MLX on llama-3.2-3B, but I had to pull a different version of it because I guess the MLX ops needed to train on quantized models are not yet implemented

Definitely still a toy, but it does as advertised: This is distributed training across an exo network cluster


exo/inference/mlx/stateful_model.py
```diff
@@ -1,12 +1,12 @@
-from typing import Dict, Tuple
+from typing import Dict, Tuple, Optional
 from collections import OrderedDict
 
 import mlx.core as mx
 import mlx.nn as nn
 from mlx_lm.models.cache import make_prompt_cache
+import numpy as np
```
dtnewman, line 7: I don't see lumpy used in the new code below

blindcrone: TBH this was an intermediate change that I think got rolled into a cleanup commit. This whole file might be unnecessary as I hoisted the whole caching mechanism to the inference engine instead of using a subclass

exo/train/dataset.py
```diff
@@ -0,0 +1,80 @@
...
+
+    # Collect batches from dataset
+    for i in range(0, len(indices) - batch_size + 1, batch_size):
+      yield batch(i)
+
+    if not train:
+      break
+
+class Dataset:
+  def __init__(self, path: Path, preprocess=lambda item: item, loadline=json.loads, metrics={}):
```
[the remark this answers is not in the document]
blindcrone: Yup. I was initially going to extend this to be more generally applicable to other datasets and make the length masks a parameter and the dataset loader work for "any" dataset, but it became clear that this would take more design to matter and wasn't a priority at this moment

eugenehp:
Very interesting. @dtnewman what's your take on the training?

dtnewman:
> Very interesting. @dtnewman what's your take on the training?

My overall take is that @blindcrone is a brilliant engineer, doing amazing work here. I think it will be _awesome_ when we can all train our own models, locally. I can't wait to see this built out further.

Some high level notes though:

1) Running command `exo train` doesn't work. It throws `Error: This train ain't leaving the station without a model`. I think this would be more helpful if it were `Error: This train ain't leaving the station without a model (e.g. command 'exo train llama-3.2-1b').

2) The readme should probably be updated to explain how this works (although I do understand this is a work in progress, so maybe that's for later)

3) Could use some better error handling... `_process_example` in standard_node.py currently returns `None` when an exception is called. So when I run this with 'exo train llama-3.2-1b', it fails, but the error gets somewhat obscured since it returns anything at all. I'm not sure that it's worth return anything in the error case there, rather than just raising an exception.

4) @blindcrone I'm assuming you built this against llama-3.2-3b since that's what's modified in the `model_cards` in models.py. I'm assuming that if you flipped `mlx-community/Llama-3.2-1B-Instruct-4bit` to `mlx-community/Llama-3.2-1B-Instruct` it would probably work for the smaller model too?

5) I am running it now and it seems to be working so far, although I'm only 12/1000 steps in. I might need to run this overnight instead :)

eugenehp:
Awesome feedback @dtnewman, indeed great work from @blindcrone, thank you so much for the effort!

I'm probably around 2/1000 steps in reading this PR, so trying to see what would be the architecture choice if training worked as good as inference. (what kind of bottlenecks to expect).

blindcrone:
Since training on the Exo network is a whole new use case, I think the features of it will need to be built out over time, and in conversation with the community as they play with this capability on their clusters. However, some deeper architectural decisions need to be made in order to support this functionality, so I'm requesting this merge specifically to avoid massive conflicts further down the line

exo/inference/inference_engine.py
```diff
@@ -24,6 +24,22 @@ async def decode(self, shard: Shard, tokens: np.ndarray) -> str:
   async def infer_tensor(self, request_id: str, shard: Shard, input_data: np.ndarray) -> np.ndarray:
     pass
   
+  async def save_session(self, key, value):
+    self.session[key] = value
```
AlexCheema, line 28: `self.session` should be initialised in `InferenceEngine`

exo/inference/inference_engine.py
```diff
@@ -24,6 +24,22 @@ async def decode(self, shard: Shard, tokens: np.ndarray) -> str:
...
+  
+  async def ensure_session(self, key, check, value_gen, hook=None):
```
AlexCheema, line 30: I don't see any usages of `ensure_session` that use `value_gen` and `hook`

exo/inference/inference_engine.py
```diff
@@ -24,6 +24,22 @@ async def decode(self, shard: Shard, tokens: np.ndarray) -> str:
...
+      await self.save_session(key, value_gen())
+      if hook is not None:
+        hook()
+  
+  async def ensure_session_match(self, key, check, value_gen):
+    if key not in self.session or not check(self.session[key]):
+      await self.save_session(key, value_gen())
+  
+  async def clear_session(self):
+    session.empty()
```
AlexCheema, line 41: should be `self.session`?

exo/inference/inference_engine.py
AlexCheema, line 1: Generally I think it might be a good idea to move the session things into a separate class.

exo/inference/mlx/sharded_inference_engine.py
```diff
@@ -37,6 +41,18 @@ def __init__(self, shard_downloader: ShardDownloader):
     self.shard = None
     self.shard_downloader = shard_downloader
     self.executor = ThreadPoolExecutor(max_workers=1)
+    self.caches = OrderedDict()
+    self.session = {}
```
AlexCheema, line 45: Feels off to have this initialised here when it's used in the base class

exo/inference/mlx/sharded_inference_engine.py
```diff
@@ -53,24 +69,84 @@ async def decode(self, shard: Shard, tokens) -> str:
     await self.ensure_shard(shard)
     tokens = await asyncio.get_running_loop().run_in_executor(self.executor, self.tokenizer.decode, tokens)
     return tokens
+
+  async def save_checkpoint(self, shard: Shard, path: str):
```
AlexCheema, line 73: Should also be an abstract method in `InferenceEngine`

exo/inference/mlx/sharded_inference_engine.py
```diff
@@ -53,24 +69,84 @@ async def decode(self, shard: Shard, tokens) -> str:
...
+    await self.ensure_shard(shard)
+    await asyncio.get_running_loop().run_in_executor(self.executor, self.model.save_weights, path)
+
+  async def load_checkpoint(self, shard: Shard, path: str):
```
AlexCheema, line 77: Should also be an abstract method in `InferenceEngine`

exo/inference/mlx/sharded_inference_engine.py
```diff
@@ -53,24 +69,84 @@ async def decode(self, shard: Shard, tokens) -> str:
...
+    await self.ensure_shard(shard)
+    await asyncio.get_running_loop().run_in_executor(self.executor, self.model.load_weights, path)
     
   async def infer_tensor(self, request_id: str, shard: Shard, input_data: np.ndarray) -> np.ndarray:
     await self.ensure_shard(shard)
-    output_data: np.ndarray = np.array(await asyncio.get_running_loop().run_in_executor(self.executor, self.model, mx.array(input_data), request_id))
+    loop = asyncio.get_running_loop()
+    state = await self.poll_state(request_id)
+    x = mx.array(input_data).astype(mx.int64) if self.shard.is_first_layer() else mx.array(input_data)
```
AlexCheema, line 85: what does this line do? what's the cast to int64 for?

exo/inference/mlx/sharded_inference_engine.py
```diff
@@ -53,24 +69,84 @@ async def decode(self, shard: Shard, tokens) -> str:
...
+    output_data: np.ndarray = np.array(await loop.run_in_executor(self.executor, lambda: self.model(x, **state)))
```
AlexCheema, line 86: Interesting, so now we have an input (ephemeral) + state (persistent) that we give to each call to the model. This makes sense, cool.

exo/inference/mlx/sharded_inference_engine.py
```diff
@@ -53,24 +69,84 @@ async def decode(self, shard: Shard, tokens) -> str:
...
     return output_data
 
+  async def evaluate(self, request_id: str, shard: Shard, inputs, targets, lengths, loss: str = "length_masked_ce"):
+    await self.ensure_shard(shard)
+    await self.save_session('loss', loss_fns[loss])
+    loop = asyncio.get_running_loop()
+    #print(f"evaluate in <- {inputs}")
+    x = mx.array(inputs).astype(mx.int64) if self.shard.is_first_layer() else mx.array(inputs)
```
AlexCheema, line 94: what's the int64 for?

exo/inference/mlx/sharded_inference_engine.py
```diff
@@ -53,24 +69,84 @@ async def decode(self, shard: Shard, tokens) -> str:
...
+    if 'lossname' not in self.session or 'LVaG' not in self.session or self.session['lossname'] != loss:
+      await self.save_session('lossname', loss)
+      await self.save_session('LVaG', nn.value_and_grad(self.model, loss_fns[loss]))
+    if 'opt' not in self.session:
+      await self.save_session('opt', opt(lr))
+    return True
+
+  async def train(self, request_id: str, shard: Shard, inputs, targets, lengths, loss: str = "length_masked_ce", opt=optim.SGD, lr=1e-5):
+    loop = asyncio.get_running_loop()
+    nothin = await self.ensure_train(shard, loss, opt, lr)
```
AlexCheema, line 116: why is `nothin` here?

exo/inference/mlx/sharded_inference_engine.py
```diff
@@ -53,24 +69,84 @@ async def decode(self, shard: Shard, tokens) -> str:
...
 
       def load_shard_wrapper():
         return asyncio.run(load_shard(model_path, shard))
 
-      model_shard, self.tokenizer = await loop.run_in_executor(self.executor, load_shard_wrapper)
+      model_shard, self.tokenizer = await asyncio.get_running_loop().run_in_executor(self.executor, load_shard_wrapper)
       self.shard = shard
-      self.model = await loop.run_in_executor(self.executor, StatefulModel, model_shard) 
+      self.model = model_shard 
+      self.caches = OrderedDict()
```
AlexCheema, line 150: I guess the caches could be used by different shards simultaneously - is there a reason to reset this here?

exo/inference/mlx/stateful_model.py
AlexCheema, line 1: Looks like `StatefulModel` is obsolete now. Should we remove it?

exo/inference/mlx/sharded_inference_engine.py
```diff
@@ -53,24 +69,84 @@ async def decode(self, shard: Shard, tokens) -> str:
...
+    y = mx.array(targets)
+    l = mx.array(lengths)
+
+    score, gradients = await loop.run_in_executor(self.executor, train_step, x, y, l)
+    #print(f"{score=}")
+      
+    layers = [{k: v["weight"] for k,v in l.items() if 'weight' in v} for l in gradients if l]
+    #print(layers[0])
+
+    return score, np.array(layers[0]['input_layernorm'])
```
AlexCheema, line 134: if we just return `input_layernorm` wouldn't it be better to just pick that one out?

exo/inference/tinygrad/inference.py
AlexCheema, line 1: Not going to look at the tinygrad stuff since that's still WIP right?

exo/main.py
```diff
@@ -208,6 +219,59 @@ def clean_path(path):
         path = path.strip('Optional("').rstrip('")')
     return os.path.expanduser(path)
 
+async def hold_outstanding(node: Node):
```
AlexCheema, line 222: If we're going to use the `Node` type here then it should be updated to include `outstanding_requests`.
Alternatively, the `Node` abstract class seems a literal unnecessary and we can probably just get rid of it and rename `StandardNode` to `Node` since that's the only implementation and it's not really a clean abstraction (how the heck would someone write another `Node` implementation)

exo/main.py
```diff
@@ -208,6 +219,59 @@ def clean_path(path):
...
+  while True:
```
AlexCheema, line 223: Why not simplify the condition with `while node.outstanding_request:`

AlexCheema (changes requested):
Looks good as a prototype for training.
No big structural changes needed, just some small things you can check out.

exo/inference/inference_engine.py
blindcrone: Ah yup, that's an artifact of a prior implementation. I thought we could set it up to use the same kind of check every time and eliminate some redundant code, but no dice

exo/inference/inference_engine.py
blindcrone: Yea, though this actually didn't end up being used so I'm gonna get rid of it for simplification's sake

exo/inference/mlx/sharded_inference_engine.py
blindcrone: I pulled that from an MLX example because at some point the inference engine didn't like the way the tokens were encoded, but I think it's unnecessary now as they're properly typed elsewhere in the code. Will test

exo/inference/mlx/sharded_inference_engine.py
blindcrone: Yea, the specific reason for this change is that stateful models really mess with training, so we want to persist state in a way such that models can be used in both "stateful" and "unstateful" modes

exo/inference/mlx/sharded_inference_engine.py
blindcrone: I was trying to get around a weird quirk where it didn't seem to run the async routing properly if I didn't return something from it. Honestly it was completely baffling why this worked, and something I've been meaning to investigate

exo/inference/mlx/sharded_inference_engine.py
blindcrone: Yea, the initial comprehension was giving me some useful debug information while making this work, but what we actually need is a more general way of detecting the first layer that returns a gradient

exo/inference/mlx/stateful_model.py
blindcrone: Probably, though it's used in a test case that doesn't have the whole inference engine pulled in, so I'm not sure how crucial that test is but it needs to be instrumented differently to run

exo/inference/tinygrad/inference.py
blindcrone: Yea, tinygrad training is not in a working state. I did have to update the inference engine to work in the same way the MLX one has been modified to, however

exo/main.py
blindcrone: Yea, I think there's a little bit of "premature abstract base class" syndrome here. It's a common illness of OO code. Happy to refactor that

exo/main.py
blindcrone: I vaguely remember trying that and being confused that it didn't work, but I can double-check. Python async has some quirks I'm not completely clear on at times

exo/inference/inference_engine.py
blindcrone: Probably, yea. Also state information going there might be sane

exo/inference/mlx/sharded_inference_engine.py
blindcrone: Agreed

exo/inference/mlx/sharded_inference_engine.py
blindcrone: Mmm, this is where I'm not sure. Some inference engines aren't going to support training just by nature. Maybe eventually we'll need a "TrainEngine" that works separately or something, but adding training bits to the base class means it'll make implementing new inference engines that can't support training obnoxious

exo/inference/mlx/sharded_inference_engine.py
blindcrone: Since the caches are per-layer, any time the shard needs to change the caches are almost certainly invalid for the new shard

exo/inference/inference_engine.py
AlexCheema: Looks good now.

exo/inference/inference_engine.py
AlexCheema: For a future PR.

exo/inference/mlx/sharded_inference_engine.py
AlexCheema: Did you add this?

exo/inference/mlx/sharded_inference_engine.py
AlexCheema: True. we can also just throw a `NotImplementedError` by default and then if you try to train on an `InferenceEngine` that doesn't support it, you see a big fat error that tells you it's not supported

exo/inference/mlx/sharded_inference_engine.py
AlexCheema: I guess it worked since you removed it!

exo/inference/mlx/sharded_inference_engine.py
AlexCheema: I think that makes a lot of sense -- being more explicit about side effects (i.e. explicit state handling, more functional style) will help a lot I think

exo/inference/mlx/sharded_inference_engine.py
AlexCheema: That's fine, I guess.

exo/inference/mlx/stateful_model.py
AlexCheema: That test can be removed.

exo/inference/mlx/sharded_inference_engine.py
AlexCheema: Sounds good. Just comment again when this is implemented.

exo/main.py
AlexCheema: Makes sense.
Python asyncio definitely has its quirks. I found this resource helpful: https://realpython.com/async-io-python/

exo/inference/mlx/stateful_model.py
blindcrone: K done

exo/inference/mlx/sharded_inference_engine.py
blindcrone: Mm, I suppose I could try but I think doing it in a model-agnostic way or adding a customization point might be smart so this may should be future work for when multiple engines support training

exo/inference/mlx/sharded_inference_engine.py
AlexCheema: Bump.

exo/inference/mlx/sharded_inference_engine.py
```diff
@@ -53,24 +68,84 @@ async def decode(self, shard: Shard, tokens) -> str:
...
+    loop = asyncio.get_running_loop()
+    nothin = await self.ensure_train(shard, loss, opt, lr)
+    def train_step(inp, tar, lng):
+      lval, grad = self.session['LVaG'](self.model, inp, tar, lng)
+      gradlayers = grad['model']['layers']
+      self.session['opt'].update(self.model, grad)
+      mx.eval(self.model.parameters(), self.session['opt'].state, lval)
+      return lval, gradlayers
+
+    x = mx.array(inputs).astype(mx.int64) if self.shard.is_first_layer() else mx.array(inputs)
```
AlexCheema, line 123: int64

exo/inference/mlx/sharded_inference_engine.py
AlexCheema: Makes sense.

exo/inference/mlx/sharded_inference_engine.py
AlexCheema: Okay, marking as future work.

AlexCheema:
Please see unresolved threads @blindcrone

exo/main.py
blindcrone: Okay, pushed a fix for this

blindcrone:
Okay so at this point it seems like removing the abstract base class as part of the requirements for merging this is starting to produce conflicts. I think I've resolved every outstanding issue and will try to integrate the changes from main into the reconciled node class, but maintaining this as a separate branch will continue to be less feasible the more other changes happen elsewhere now that this refactor is part of it

blindcrone:
Fixed the conflicts AFAICT, hopefully this can be merged now

exo/inference/mlx/sharded_inference_engine.py
blindcrone: Ah, I responded to the wrong conversation. I think checkpoint loading makes sense for pure inference engine use cases but saving only makes sense if we expect the weights to change, and so could only work if training is a capability of the inference engine in question. I can add the abstractmethod but it's a design decision that will have to be worked around later if using inference engines that don't support training

blindcrone: I'll put it in for now, and we can cross that bridge as we come to it I suppose

AlexCheema (approved):

PR #481, opened 2024-11-21
use new tinychat UI

roryclear:
css changes copied over from tinygrad, left and right, instead of red and green.

don't think tinygrad like the red and green
https://github.com/geohotstan/tinycorp_meetings/blob/master/2024-11-18/README.md


AlexCheema:
thanks

PR #486, opened 2024-11-22
restore the cursor to the terminal on exit from CLI

dtnewman:
Running on a MacBook, the cursor disappears if I exit the CLI using CTRL + C. This is a minor change that restores the cursor on exit.


AlexCheema:
nice!! many people complained about this one in passing haha

PR #487, opened 2024-11-22
update tinygrad version

roryclear:
- threefry is gone from tinygrad along with np random
- added change from [f4ec39](https://github.com/tinygrad/tinygrad/commit/f4ec39fe58259216697304c453d519bc26a43e5a) which was breaking llama inference for me

I've only tested on one mac (so not certain it will still work on a cluster) by just hardcoding the inference engine name to tinygrad, getting the same outputs as main


AlexCheema:
tested on a cluster and works.

AlexCheema:
please email alex@exolabs.net to claim $200 bounty @roryclear

Issue #461, opened 2024-11-15
FLUX Support


AlexCheema:
Add support for this model with the MLX backend: https://huggingface.co/black-forest-labs/FLUX.1-dev

There's already an example of FLUX using MLX here: https://github.com/ml-explore/mlx-examples/blob/main/flux/flux/model.py

varshith15:
it is able to run on my 32gb mac :)
can i take this up @AlexCheema ?

AlexCheema:
> it is able to run on my 32gb mac :) can i take this up @AlexCheema ? wanted to explore sd, flux for the longest, this will be a really good chance

It should be, definitely at Q4. Assigned - good luck!

ZhihaoAIRobotic:
Want to participate too!

AlexCheema:
> Want to participate too!

Sure, if you and @varshith15 agree to work together I will award the full $500 to both of you.

varshith15:
@ZhihaoAIRobotic im down, lets talk on discord :)

ZhihaoAIRobotic:
No problem  :)

Incharajayaram:
Hi @AlexCheema, if the issue hasn't been solved yet, I would like to take this up. Please assign me. If you want, I could provide you with a comprehensive plan of what I am going to implement.

Evanev7:
We're closing bounties, thank you to all who contributed, and apologies that we left this unmanaged for so long.

PR #498, opened 2024-11-25
Add FLUX support with MLX backend

vishwamartur:
Related to #461

Add support for the FLUX model with the MLX backend.

* **New Model Class**: Add `exo/inference/mlx/models/flux.py` to define the `FluxModel` and `Model` classes for the FLUX model.
* **Inference Engine Update**: Modify `exo/inference/mlx/sharded_inference_engine.py` to include necessary imports for the FLUX model.
* **Testing**: Add `exo/inference/mlx/test_sharded_flux.py` to include tests for the FLUX model, ensuring proper integration and functionality.
* **Documentation**: Update `README.md` to include instructions for using the FLUX model with the MLX backend, along with usage examples.

### DIFF
--- a/README.md
+++ b/README.md
@@ -266,3 +266,38 @@ exo supports the following inference engines:
 - ✅ [GRPC](exo/networking/grpc)
 - 🚧 [Radio](TODO)
 - 🚧 [Bluetooth](TODO)
+
+## FLUX Model Support
+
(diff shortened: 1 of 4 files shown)


AlexCheema:
Did you test this?

AlexCheema:
I don't think this works. Please re-open if that's not that case.

PR #542, opened 2024-12-07
Support CPU-Only CLANG

wbic16:
In addition to applying this patch to tinygrad, I had to also install a number of Python packages. I've listed them below. I can look at integrating them in the install process if you want.

```
apt install build-essential -y
apt install clang -y
apt install python3.12-venv -y
apt install linux-libc-dev -y
apt install python3-dev -y
apt install python3-pip -y
pip3 install llvmlite
pip3 install numba


```

### DIFF
--- a/exo/inference/tinygrad/models/llama.py
+++ b/exo/inference/tinygrad/models/llama.py
@@ -275,6 +275,12 @@ def convert_from_huggingface(weights: Dict[str, Tensor], model: Transformer, n_h
 
 
 def fix_bf16(weights: Dict[Any, Tensor]):
+  if Device.DEFAULT == "CLANG":
+    # TODO: without casting to float16, 70B llama OOM on tinybox.
+    return {
+      k: (v.llvm_bf16_cast(dtypes.float32).to(v.device) if v.dtype == dtypes.bfloat16 else v) 
+      for k, v in weights.items()
+    }
   if getenv("SUPPORT_BF16", 1):
     # TODO: without casting to float16, 70B llama OOM on tinybox.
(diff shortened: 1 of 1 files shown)


qwertystars:
I could not run it. I got a error "No such file or directory: 'clang'"
[log.txt](https://github.com/user-attachments/files/18099604/log.txt)
log files for command "DEBUG=9 CLANG=1 exo" why is this happening? On my other linux which has no gpu it doesn't show problem.

Nurb4000:
> In addition to applying this patch to tinygrad, I had to also install a number of Python packages. I've listed them below. I can look at integrating them in the install process if you want.
> 
Since many of us will want to use spare CPU only nodes to supplement a 'master' node with GPU i think making it easy to install in both cases would be a good thing.

wbic16:
> I could not run it. I got a error "No such file or directory: 'clang'" [log.txt](https://github.com/user-attachments/files/18099604/log.txt) log files for command "DEBUG=9 CLANG=1 exo" why is this happening? On my other linux which has no gpu it doesn't show problem.

Have you installed clang on your system?

qwertystars:
So sorry, i forgot to install clang. But the issues still exists like clang is ok but the command is not running properly.

qwertystars:
I just found out that nvidia doesn't like clang lol. The funny thing is when I ran this code on a container that didn't have nvidia it worked. Any other way to not use clang?

AlexCheema:
This isn't an ideal fix but since CLANG isn't working on most devices, I'll merge this for now.

PR #569, opened 2024-12-22
Use `#!/usr/bin/env bash` for better portability

deftdawg:
This change allows the install script and others to operate on systems where bash is not installed in `/bin` (i.e. nixOS)


AlexCheema:
Max portability. Thank you!

PR #590, opened 2025-01-06
Fix the /v1/models API to output proper OpenAI compatible endpoint

metaspartan:
Modify the `/v1/models` API to output a proper OpenAI compatible endpoint with an object and a `data` object containing the models list.

* Change the `handle_get_models` method in `exo/api/chatgpt_api.py` to wrap the models list in an object with a `data` field.
* Add an `object` field with the value "list" to the response format.


AlexCheema:
Awesome - thank you

PR #647, opened 2025-01-28
Add 4-bit to the end of DeepSeek V3/R1 model descriptions

deftdawg:
I was surprised that the DeepSeek R1 model was only ~390GB...  Digging into it a bit, I noticed because this because it's actually the 4-bit version of the model.   This PR adds that to the model's key/description. 

There may have been a reason why this was omitted, like to have it preferentially sort to the top or match loading from the CLI.   If that's the case, feel free to close this PR.


AlexCheema:
All the default models in exo are 4-bit. Happy to add it to the description but lets keep the `deepseek-r1` short name.

Are you running DeepSeek R1? What hardware are you running on?

deftdawg:
> Are you running DeepSeek R1? What hardware are you running on?

Was tinkering with some of the small models/quants on my M3 laptop, seeing if I could get it to talk to my AMD RX6900XT while I wait for access to company's Studio M2 Ultra/192...

Expecting to run a mix R1 and Qwen-2.5-coder depending on work type and max context window...  Have some small projects I think either will do well on and 1 project I really need help with that I think both will struggle with (150k+ loc 😭 )

AlexCheema (approved):

PR #607, opened 2025-01-17
Fixes for cross-platform operability

tensorsofthewall:
Hello all,
I've updated the exo library as part of my attempt to bring [Windows Native support](https://github.com/exo-explore/exo/issues/186).  This is NOT the actual update for Windows Native support, but these changes are important as they can help fix issues related to other PRs (#139).

### Fixes
1, Updated the **mlx.core** import that prevents exo from running on Linux/Windows nodes, since mlx isn't available at all for these systems. Linux/Windows nodes will use **numpy** instead. **_[Tested]_**
2. Added AMD GPU querying through [**pyrsmi**](https://github.com/ROCm/pyrsmi) (official Python bindings for [**_rocm-smi-lib_**](https://github.com/ROCm/rocm_smi_lib) [AMD supported]). **_[Untested]_**
3. Added Windows Device Capabilities, along with the aforementioned AMD GPU querying. exo on Windows now reports the available GPU and VRAM/TFLops correctly. **_[Tested]_**
4. Minor fix for type hint in inference-engine.

Please review and suggest changes. Feedback is appreciated.

Thanks!

PR #649, opened 2025-01-29
Fix AMD device capabilities fields

eclecticc:
Resolves an issue introduced in PR #607


CarlosR759:
Hi this fix will be in the next release of exo ? If I'm not wrong I'm having this issue with running AMD=1 exo in the env . If that's the case when will be the new release available ? I installled Exo and it works with cpu using the clang flag, but I really want to test it with the amd gpu.  Also many thanks for the fix.

Issue #488, opened 2024-11-23
tinygrad threading issue


AlexCheema:
How to reproduce:

- Run a cluster on 2 nodes
- Run a request
- Restart one node
- Run another request

```
Traceback (most recent call last):
  File "/Users/alex/exo/exo/api/chatgpt_api.py", line 288, in handle_post_chat_completions
    await asyncio.wait_for(asyncio.shield(asyncio.create_task(self.node.process_prompt(shard, prompt, request_id=request_id))), timeout=self.response_timeout)
  File "/usr/local/python-X.Y.Z-static/lib/python3.12/asyncio/tasks.py", line 520, in wait_for
    return await fut
           ^^^^^^^^^
  File "/Users/alex/exo/exo/orchestration/standard_node.py", line 155, in process_prompt
    resp = await self._process_prompt(base_shard, prompt, request_id)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/alex/exo/exo/orchestration/standard_node.py", line 184, in _process_prompt
    resp = await self.forward_prompt(shard, prompt, request_id, 0)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/alex/exo/exo/orchestration/standard_node.py", line 272, in forward_prompt
    await target_peer.send_prompt(next_shard, prompt, request_id=request_id)
  File "/Users/alex/exo/exo/networking/grpc/grpc_peer_handle.py", line 81, in send_prompt
    response = await self.stub.SendPrompt(request)
               ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/alex/exo/.venv/lib/python3.12/site-packages/grpc/aio/_call.py", line 327, in __await__
    raise _create_rpc_error(
grpc.aio._call.AioRpcError: <AioRpcError of RPC that terminated with:
        status = StatusCode.UNKNOWN
        details = "Unexpected <class 'sqlite3.ProgrammingError'>: SQLite objects created in a thread can only be used in that same thread. The object was created in thread id 6375518208 and this is thread id 18458554368."
        debug_error_string = "UNKNOWN:Error received from peer  {grpc_message:"Unexpected <class \'sqlite3.ProgrammingError\'>: SQLite objects created in a thread can only be used in that same thread. The object was created in thread id 
[...]

Issue #389, opened 2024-10-28
Crash due: SQLite objects created in a thread can only be used in that same thread


FFAMax:
Found instability:

```
Error processing tensor for shard Shard(model_id='mlabonne/Meta-Llama-3.1-8B-Instruct-abliterated', start_layer=0, end_layer=15,
n_layers=32): SQLite objects created in a thread can only be used in that same thread. The object was created in thread id
139810278716992 and this is thread id 139809641199168.
Traceback (most recent call last):
  File "/home/ffamax/exo/exo/orchestration/standard_node.py", line 239, in _process_tensor
    result, inference_state, is_finished = await self.inference_engine.infer_tensor(request_id, shard, tensor,
inference_state=inference_state)
  File "/home/ffamax/exo/exo/inference/tinygrad/inference.py", line 76, in infer_tensor
    await self.ensure_shard(shard)
  File "/home/ffamax/exo/exo/inference/tinygrad/inference.py", line 97, in ensure_shard
    self.model = await asyncio.get_event_loop().run_in_executor(self.executor, build_transformer, model_path, shard, "8B" if "8b"
in shard.model_id.lower() else "70B")
  File "/usr/lib/python3.10/concurrent/futures/thread.py", line 58, in run
    result = self.fn(*self.args, **self.kwargs)
  File "/home/ffamax/exo/exo/inference/tinygrad/inference.py", line 48, in build_transformer
    load_state_dict(model, weights, strict=False, consume=False)  # consume=True
  File "/home/ffamax/exo/.venv/lib/python3.10/site-packages/tinygrad/nn/state.py", line 129, in load_state_dict
    else: v.replace(state_dict[k].to(v.device)).realize()
  File "/home/ffamax/exo/.venv/lib/python3.10/site-packages/tinygrad/tensor.py", line 3475, in _wrapper
    if _METADATA.get() is not None: return fn(*args, **kwargs)
  File "/home/ffamax/exo/.venv/lib/python3.10/site-packages/tinygrad/tensor.py", line 213, in realize
    run_schedule(*self.schedule_with_vars(*lst), do_update_stats=do_update_stats)
  File "/home/ffamax/exo/.venv/lib/python3.10/site-packages/tinygrad/engine/realize.py", line 222, in run_schedule
    for ei in lower_schedule(schedule):
  File "/home/ffamax/exo/.venv/lib/python3.10/site-packages/tinygrad/engine/realize.py", line 215, in lower_schedule
    raise e
[...]

PR #653, opened 2025-01-29
Tiny fixes: SQLite Thread issue

AlexCheema:
- Fixes tinygrad issue where some code runs on different threads leading to errors like this: `SQLite objects created in a thread can only be used in that same thread. The object was created in thread id 6375518208 and this is thread id 18458554368.`
- Fixes visual bug where tinychat would be open and cause other nodes to see download progress even though not in progress
- Fix visual bug where prompt outputs would not show in the TUI after 3 requests


AlexCheema:
Fixes #488

AlexCheema:
Fixes #389

PR #656, opened 2025-01-30
Add `--gpu` CLI option so Linux users can choose a single GPU ID for tinygrad

freerainboxbox:
### Original Behavior

1. User has to set `VISIBLE_DEVICES` in shell for tinygrad to accept choice of GPUs.
2. Even if set to something other than `0`, `linux_device_capabilities()` will fetch only index 0 (this was hardcoded in).

### New Behavior

1. The user may specify `--gpu` followed by an integer. If parsed incorrectly or omitted, it will default to `0`. `VISIBLE_DEVICES` in the parent environment is effectively abstracted away from the user, so if they want to pick a device, they can pick only one for the instance.*
2. Device capabilities will fetched according to this index, and the user will see the correct GPU as confirmation in the TUI (see behavior of `topology.py`).

Users can run multiple GPUs at once by running multiple instances of exo. This workaround I wrote may be obsolete if we implement multi-GPU in a single instance of exo, but for now, this will allow users to take advantage of multiple GPUs in one device.

### DIFF
--- a/exo/main.py
+++ b/exo/main.py
@@ -69,6 +69,7 @@ parser.add_argument("--save-checkpoint-dir", type=str, default="checkpoints", he
 parser.add_argument("--node-id", type=str, default=None, help="Node ID")
 parser.add_argument("--node-host", type=str, default="0.0.0.0", help="Node host")
 parser.add_argument("--node-port", type=int, default=None, help="Node port")
+parser.add_argument("--gpu", type=int, default=0, help="GPU ID to use")
 parser.add_argument("--models-seed-dir", type=str, default=None, help="Model seed directory")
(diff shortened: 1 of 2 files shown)


pharazon:
This would be useful. In the meanwhile to get exo to support multiple GPUs, you could run multiple exo instances on a single machine, each for different GPUs.

pharazon:
Also I can't get this to work. I have 2xGTX3060 running on screen. The first exo -instance connects to the cluster ok.

The second exo-instance that has been started by exo --gpu 1 creates it's own cluster with 1 node, but does not connect to the first instance or the cluster.

Evanev7:

PR #669, opened 2025-02-02
Add toggle to show only models downloaded locally

pavel-rodionov:
With a growing list of supported models it might be tricky to find particular one. 

* Add a toggle to show only models downloaded locally 

Screenshot 2025-02-01 at 23 47 55


AlexCheema:
Helpful - thank you!

PR #700, opened 2025-02-14
feat: optimize download progress polling logic

li-linfeng:
- Add conditional polling based on download status
- Add debug logs for tracking polling lifecycle
- Fix continuous polling issue on page load
- Implement proper cleanup of polling intervals
- Add checkOngoingDownloads to handle download state persistence


AlexCheema:
I see the issue - we're polling every second so if `fetchDownloadProgress` takes >1s we'll overwhelm the server.
Can we just change `startDownloadProgressPolling` to be the same as `startModelPolling`, using async?

PR #705, opened 2025-02-17
adding current model name to input container information

cadenmackenzie:
With the new collapsable model navigation in the sidebar, it is not immediately clear when the page loads what model is being used for the chat. 

My solution: I added the model name to the input container information but some other options would be:

1.  To put it in the header
2. To put a "Current Model" in the side bar under Network Topology that is always visible.
3. Expand the model selection to the model that is currently select.

Let me know if you would prefer an alternate solution.


AlexCheema:
Perfect, yes this is a big issue right now.

PR #713, opened 2025-02-19
refactor(tinychat): improve download progress polling mechanism

li-linfeng:
- Add exponential backoff retry strategy for download progress polling
- Replace interval-based polling with while loop and dynamic intervals
- Add isPolling flag to prevent concurrent polling
- Add allProgressComplete flag for better completion tracking
- Remove downloadProgressInterval as it's no longer needed
- Optimize polling termination conditions

The changes make the download progress tracking more efficient and reliable by:
1. Using exponential backoff (1s to 32s) when no progress is detected
2. Automatically stopping polling after max retries or completion
3. Preventing multiple concurrent polling operations

### DIFF
--- a/exo/tinychat/index.js
+++ b/exo/tinychat/index.js
@@ -28,8 +28,9 @@ document.addEventListener("alpine:init", () => {
     imagePreview: null,
 
     // download progress
+    isPolling: false,  // To check if the download progress is polling
+    allProgressComplete: false, // To check if the download progress is all complete
     downloadProgress: null,
-    downloadProgressInterval: null, // To keep track of the polling interval
 
     // Pending message storage
     pendingMessage: null,
@@ -436,7 +437,6 @@ document.addEventListener("alpine:init", () => {
         }
       }
     },
-
     async fetchDownloadProgress() {
       try {
         const response = await fetch(`${this.endpoint}/download/progress`);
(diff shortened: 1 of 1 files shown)


exo/tinychat/index.js
```diff
@@ -593,6 +616,8 @@ document.addEventListener("alpine:init", () => {
           throw new Error(data.error || 'Failed to start download');
         }
 
+        await this.startDownloadProgressPolling();
```
AlexCheema, line 619: should this be awaited? isn't this going to loop forever? (while true in `this.startDownloadProgressPolling`)

AlexCheema (changes requested):

exo/tinychat/index.js
li-linfeng: 1. After clicking to start download, the download progress cannot be obtained in real-time successfully
2. In this.startDownloadProgressPolling(), there are several termination conditions set:
 - All downloads are in complete status
 - If the current download progress is not obtained, the timer interval will increase exponentially, and the loop will stop after a maximum of 4 iterations

Therefore, considering all these conditions, it will not remain in a loop state indefinitely

li-linfeng: ```await``` here need to be removed

JakeHillion:

Issue #701, opened 2025-02-14
Load distribution between nodes.


morlz:
I think it is worth sorting nodes by flops and executing more layers on those nodes where there is more performance. This will speed up execution in cases where the model size is smaller than the cluster memory. It is necessary to pass the size of the current model to PartitioningStrategy for this to work.

AlexCheema:
This will be the default in exo v2!

rltakashige:
A version of this is in EXO 1.0 - plan to improve this though.

rltakashige:
Please open again if this is still an issue!

PR #719, opened 2025-02-20
Optimize load distribution between nodes

vishwamartur:
Related to #701

Implement load distribution between nodes based on flops and memory.

* **Partitioning Strategy**:
  - Add `flops` attribute to `Partition` class in `exo/topology/partitioning_strategy.py`.
  - Update `map_partitions_to_shards` function to consider `flops` when mapping partitions to shards.
  - Add `get_flops` method to `PartitioningStrategy` class to calculate the flops of each partition.

* **Ring Memory Weighted Partitioning Strategy**:
  - Update `partition` method in `exo/topology/ring_memory_weighted_partitioning_strategy.py` to consider both memory and flops for partitioning.
  - Add `calculate_flops_weight` helper function to calculate the weight of each node based on its flops.

* **Node Class**:
  - Update `Node` class in `exo/orchestration/node.py` to implement logic to sort nodes by flops for load distribution.
  - Add `sort_nodes_by_flops` method to sort nodes by their flops.

* **Inference Engine**:
  - Add `get_flops` method to `InferenceEngine` class in `exo/inference/inference_engine.py` to get the flops of the current node.

* **Sharded Inference Engine**:
  - Add `get_flops` method to `MLXDynamicShardInferenceEngine` class in `exo/inference/mlx/sharded_inference_engine.py` to get the flops of the current node.

### DIFF
--- a/exo/inference/inference_engine.py
+++ b/exo/inference/inference_engine.py
@@ -50,6 +50,9 @@ class InferenceEngine(ABC):
 
     return output_data, inference_state
 
+  async def get_flops(self) -> float:
+    return self.device_capabilities.flops.fp32
+
 
 inference_engine_classes = {
   "mlx": "MLXDynamicShardInferenceEngine",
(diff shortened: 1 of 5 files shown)


JoaGamo:
Problem I'm getting with this change (that was implemented in https://github.com/water-vapor/exo fork) is that when the Flops can't be obtained, for example, when the GPU isn't in the hardcoded list, this will make a division by zero error. I would add a fallback to prevent this exception.

JakeHillion:

PR #716, opened 2025-02-19
Improvements to the ChatGPT API to more closely match the official OpenAI endpoint and improve compatibility

joshuacoles:
Hey, as I mentioned on Discord, we are experimenting with a Mac mini Exo deploy at work and noticed that the ChatGPT API offered by Exo behaved differently than the official OpenAI endpoint in a way which made working with 3rd party chat clients difficult.

To make the Exo API behave more similarly to OpenAI's, we have done the following:

1. Remove the EOS token from the output.
    - This was the primary issue as 3rd party chat clients did not expect the EOS token (`<|eot_id|>` for LLAMA 3.2) to be included in the response, so they were not filtering them.
    - It did not manifest in the built-in Tinychat UI as the EOS token was only emitted with `finish_reason="stop"`, in which case Tinychat ignored the content of the delta.
2. When streaming, the API now emits a `data: [DONE]` event to indicate completion before terminating the stream.
3. When streaming, only include the message under the `delta` key rather than both that and the `message`, mirroring the OpenAI type.
4. If the streamed content is empty (as will occur for the finishing chunk now that the EOS token is stripped), set `delta: {}`.

## Testing

To test this you can use the [llm command line tool](https://github.com/simonw/llm) configured to talk to exo by editing the following file "~/Library/Application\ Support/io.datasette.llm/extra-openai-models.yaml" on macOS to contain the following,

```yaml
- model_id: llama-3.2-1b
  model_name: llama-3.2-1b
  api_base: "http://localhost:52415/v1"
```

### Old

```
❯ llm chat -m llama-3.2-1b
Chatting with llama-3.2-1b
Type 'exit' or 'quit' to exit
Type '!multi' to enter multiple lines, then '!end' to finish
> Hello
Hello! How can I assist you today?<|eot_id|>
```

### New

```
❯ llm chat -m llama-3.2-1b
Chatting with llama-3.2-1b
Type 'exit' or 'quit' to exit
Type '!multi' to enter multiple lines, then '!end' to finish
> Hello
Hello! How can I assist you today?
```

Note the lack of "<|eot_id|>" in the output.

PR #720, opened 2025-02-20
Max completion tokens

joshuacoles:
This PR builds on the changes from #716 adding the `max_completion_tokens` property to the ChatGPT API's `ChatCompletionRequest` in alignment with the [OpenAI API spec](https://platform.openai.com/docs/api-reference/chat/create#chat-create-max_completion_tokens). This is used to provide a client defined limit on the maximum number of tokens returned from inference, as long as this is less than the `node.max_generate_tokens` specified on the command line. This supports both the new `max_completion_tokens` and the older `max_tokens` option if the former is not present.

The main structural change in this PR is the addition of the `GenerationOptions` parameter to the `SendPrompt` and `SendTensor` grpc methods and the methods which they call. This allows for request specific options during inference. I kept this separate from the existing `InferenceState` as this seemed focused on stable diffusion specific data which could change over the course of inference.

If any more changes are made to the parent PR I will rebase this PR to match.


joshuacoles:
Also @AlexCheema whilst I was adding this I noticed that the `SendPrompt` and `SendTensor` both return a Tensor as per the grpc definitions but the corresponding methods on Node (`Node#process_prompt` and `Node#process_tensor`) do not return anything, so this tensor is always empty.

I see this changed from `Tensor` -> `Empty` in c9ded9ba967335f414321e3ac51100333b1d14fb then back in 9954ce8e4ddc573de48a56503d87287c7b2b067d. I have left this signature unchanged but wanted to check which is the expected behaviour?

joshuacoles:
Closed in favour of the combined PR #734

PR #735, opened 2025-02-25
Support structured responses from the ChatGPT API

joshuacoles:
This PR implements the `response_format` property from the OpenAI API Spec ([link](https://platform.openai.com/docs/api-reference/chat/create#chat-create-response_format)) to allow specifying. It is implemented using the [Low Level Guidance](https://github.com/guidance-ai/llguidance) project which allows token masking based on JSON schemas and a variant of BNF grammars.

There is support for JSON decoding, both with and without a supplied schema (although only supports a [subset](https://github.com/guidance-ai/llguidance/blob/main/docs/json_schema.md) of JSON schemas). This is done by incorporating the masking process into the BufferedOutput object introduced in #734.

As with my other PRs this builds on them incrementally the others (#734, #720, #716) should be reviewed prior to this, I will keep this branch up to date with any changes from the prior PRs.

## Still todo

- Testing
- Performance testing to minimise regressions in inference, especially when doing so without a grammar.


joshuacoles:
Closed in favour of consolidated PR #771

PR #770, opened 2025-03-08
Add Dockerfiles, Helm chart and related GHA workflow

s3rj1k:


MattJCroteau:
Personally, I'm hoping this merges (and I'll probably snag it before then).  I'm thinking this might be a good way to solve the 'exo only supports 1 GPU per host" problem (That has a 300 dollar bounty on btw).  I plan to allocate an external IP per container and assign one GPU to each. A little out of the box but this would make that possible. So Thank You from me!

s3rj1k:
[6a7f22df9897f04f75e5b2f1e7a537ba13be58ad.patch](https://github.com/user-attachments/files/21823449/6a7f22df9897f04f75e5b2f1e7a537ba13be58ad.patch)

Change is here if somebody needs it, seems like project has run out of steam

PR #771, opened 2025-03-08
Structural generation and function calling

joshuacoles:
This is a combined PR for structural generation and OpenAI spec compatible function calling. This is still WIP but builds on my previous work with tests and a cleaner git history and implementation. I will fill this description in as I go


joshuacoles:
This is now ready for review! @AlexCheema This adds the infrastructure and function calling to Exo using [LLGuidance](https://github.com/guidance-ai/llguidance) for guided generation, as well as greater support for custom models and chat templates which I needed to implement during testing.

There are a number of edges left to smooth out but these can be done in parallel with the review process. Possible improvements,

- [ ] Re-enable message mapping for image messages
- [ ] Add an example that provides a better chat experience. Currently models I have been using are very insistent on calling functions at the detriment of conversation.
- [ ] Add MCP support to Tinychat
- [ ] Enable parallel calling
- [ ] Investigate allowing tool calling to be interspersed with standard generation (I am uncertain if this is spec compliant but could be cool).
- [ ] See if we can interface better with LLGuidance, both in as it is currently implemented, but also to avoid needing to manually write parsers for the grammar definitions we already have in lark format.
- [ ] Tests, there are some but more, specifically unit tests, would be good

joshuacoles:
As a review of what is included in this PR,

- Support for locally defined models read from a JSON file specified on the command line.
    - This allows a user to use `mlx_lm.convert` to convert an existing model to MLX and use it without code changes. This may be made defunct by support for hugging face models but for now is useful in testing.
- Support for specifying a custom chat template per model.
- Support for guided generation of:
    1. A JSON object (optionally conforming to a JSON schema)
    2. Output matching a regex (in rust syntax I believe)
    3. Output matching a lark grammar
- Support for tool calling, using guided generation, returning the tool call in a structured form in line with the OpenAI spec.

Note this does not currently implement incremental tool calling due to limitations in the parsing infrastructure and a desire to reduce complexity.

joshuacoles:
I have now added parallel tool calling for the watt format and explicitly disabled it for the llama python tag format as it seems to be disabled in other places (see <https://github.com/ollama/ollama/blob/4aeb67ef4c0c734e0077af39e0b8be1252662cc0/template/index.json#L95>), if we want to explore adding this in the future we will need to make sure that the chat template we use supports it.

In addition I have added some basic tests to validate tool parsing for both formats.

joshuacoles:
I am going to continue to cleanup the git history to make it easier to review commit wise but I would say this is now feature complete. There are more things to add but you've got to stop somewhere and this is enough for our immediate use case.

JakeHillion:

PR #797, opened 2025-03-20
CPU Mode Timeout Problem

steve15963:
Low spec computers sometimes have timeouts.
Because all cores are executing matrix operations.
So I increased the timeout time because the gRPC server could not respond.

### DIFF
--- a/exo/orchestration/node.py
+++ b/exo/orchestration/node.py
@@ -560,7 +560,10 @@ class Node:
         continue
 
       try:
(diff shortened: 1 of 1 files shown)


steve15963:
and CPU Mode need 4 Dependency 
then we can install 2 line command
sudo apt-get install libgl1-mesa-glx libgl1-mesa-dev clang
pip install numba llvmlite

PR #810, opened 2025-04-03
Don't assume all network interfaces are /24 in size

kdkd:
Rather than replacing the final octet of each local address with .255, look in "ifconfig" (mac), "ip addr" (linux), or "ipconfig /all" (windows) for the correct netmask and create the actual broadcast address.

This fixes issues where a device has multiple nearby /30 interfaces where the first three octets were identical, causing the wrong broadcast address to get used.

e.g. if the host has 1.0.0.1/30 and 1.0.0.5/30, both interfaces were attempting to broadcast to 1.0.0.255 which is incorrect. This patch makes it correctly send broadcasts to 1.0.0.4 and 1.0.0.8.

This is tested and in use in Mac and Linux, but only a simple test on Windows was done since we aren't using that here.

### DIFF
--- a/exo/helpers.py
+++ b/exo/helpers.py
@@ -15,6 +15,7 @@ import tempfile
 import json
 from concurrent.futures import ThreadPoolExecutor
 import traceback
+import struct
 
 DEBUG = int(os.getenv("DEBUG", default="0"))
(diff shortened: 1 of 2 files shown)


Evanev7:

PR #824, opened 2025-04-26
feat(project): add pyproject.toml for PEP 621 metadata

hotwa:
- define project metadata (name, version, requires-python) and dependencies in pyproject.toml
- configure setuptools.build_meta as build-backend for PEP 660 editable installs
- replace setup.py install_requires & entry_points with PEP 621 fields
- include package-data and optional-dependencies for apple_silicon, nvidia-gpu, etc.

Background
The project currently relies solely on setup.py to declare its dependencies and entry points. This prevents full compatibility with PEP 660 editable installs and modern tooling like uv sync and pip install -e ., causing some dependencies to be skipped. We need a pyproject.toml so that build back-ends and dependency managers can understand our metadata natively.

Changes
Add pyproject.toml using PEP 621 to declare:

Project metadata: name, version, description, requires-python

dependencies: migrated from install_requires

optional-dependencies: migrated from extras_require

scripts: replaces entry_points for the exo console script

package-data: includes tinychat assets under exo

Retain the existing setup.py for backward compatibility; it can be removed once downstream consumers have migrated.

Update CI workflows (if any) to run pip install -e . and/or uv sync against both setup.py and pyproject.toml flows.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -1,0 +1,68 @@
+[build-system]
+requires = [
+  "setuptools>=66.0", 
+  "wheel"
+]
+build-backend = "setuptools.build_meta"
+
+[project]
+name = "exo"
+version = "0.0.1"
+description = "Distributed LLM inference framework"
+readme = "README.md"            # adjust if you use another file
+requires-python = ">=3.12.0"
+dependencies = [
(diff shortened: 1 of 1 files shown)


JakeHillion:

PR #851, opened 2025-06-23
fix: Update mlx dependency to >=0.22.1 for Apple Silicon compatibility

dharun-web:
## Changes Made
- Updated `mlx` dependency in `setup.py` from `==0.22.0` to `>=0.22.1`.  
- Adjusted `mlx-lm` to `>=0.21.1` for consistency.

## Why This Is Needed
- `mlx==0.22.0` does not exist in PyPI, causing installation failures on Apple Silicon (macOS).  
- This change ensures compatibility with available versions while maintaining functionality.

## Testing
1. **Local Verification**:
   ```bash
   pip install -e .  # Successfully installed mlx>=0.22.1

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -35,8 +35,8 @@ install_requires = [
 extras_require = {
   "formatting": ["yapf==0.40.2",],
   "apple_silicon": [
-    "mlx==0.22.0",
-    "mlx-lm==0.21.1",
+    "mlx>=0.22.0",
+    "mlx-lm>=0.21.1",
   ],
   "windows": ["pywin32==308",],
   "nvidia-gpu": ["nvidia-ml-py==12.560.30",],


AbhiByte:
lgtm!

AbhiByte (approved):

manueltarouca:
Nice

pierow2k:
Wouldn't it be safer to pin to `==0.22.1` (which **is** available on PyPI) rather than opening it up with `>=0.22.1`? A looser constraint could let through future, untested releases that might break the build. 

Have we confirmed that *all* versions newer than 0.22.1 are compatible? (Up through 0.26.1 in the case of `mlx` and through 0.25.2 in the case of `mlx-lm`)

anirudhp06 (commented):
When u mention 
> Updated mlx dependency in setup.py from `==0.22.0` to `>=0.22.1`

But when checked in files of your commit, you have only changed `==0.22.0` to `>=0.22.0` which is completely different from what the commit message.

Is this something you missed while checking or is it intentional?

mymstudios:
image

JakeHillion:

PR #915, opened 2025-12-18
8000 -> 52415

Evanev7:
8000 is a common dev port - we should use the EXO port 5 (E) 24 (X) 15 (O)!


mbaprogress008-stack:
Wonderful

PR #922, opened 2025-12-19
Add GitHub Issue templates

JustSamuel:
# Description
Looking through the issues to find something to pick up is quite difficult as it is largely unstructured and uncategorized (bug reports, support requests and features). I looked over them to find the most commonly requested information and tried to wrap those up in issue templates.

It could also be considered to _not_ have support questions in GitHub issues (as is common practice) but have a separate support channel. Given that there are already quite a lot of support questions it might be nicer to just embrace them and label them as such.

## Types of changes
- Documentation improvement

### DIFF
--- a/.github/ISSUE_TEMPLATE/1-bug-report.yml
+++ b/.github/ISSUE_TEMPLATE/1-bug-report.yml
@@ -1,0 +1,70 @@
+name: "🐛 Bug Report"
+description: Report a reproducible bug or regression.
+title: "[Bug]: "
+labels: ["Type: Bug", "Status: Triage"]
+body:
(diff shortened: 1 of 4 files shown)


JakeHillion:
Hi, I think this should be covered. We have a pull request template with different casing and issue templates for bugs and features. If you'd like to make improvements to them, please submit a separate PR/re-open this one updating the existing files rather than adding new ones. Thanks!

JustSamuel:
> We have a pull request template with different casing and issue templates for bugs and features.

My bad, I should have checked better :sweat_smile: I re-opend an old PR since the other was automatically closed but check well enough

PR #928, opened 2025-12-19
refactor all information sources (including ipless rdma discovery)

Evanev7:
## Motivation

Information gathering is tightly coupled to MacMon - we should start generalizing our information sources so we can add more in future.

## Changes

Added a new system to gather any information. Currently, it is attached to the Worker - though this is mostly to keep the data processing logic simple. It could be made independent quite easily.

I also refactored topology to include different kinds of connections as we can gather RDMA connections without having a pre-existing socket connection, and made the relevant placement updates. We should no longer need the network locations script in the app.

Other sources of information now include:
- static node information like "model" and "chip" (macos, "Unknown" fallback)
- device friendly name (macos, falls back to device hostname)
- network interfaces + ips (cross platform)
- thunderbolt interfaces (macos)
- thunderbolt connections (macos)
- RAM usage (cross platform)
- per-device configuration written to EXO_HOME/config.toml

## Limitations

The current events added by the InfoGatherer are much too broad and don't follow proper Pydantic validation.

Model and Chip are not cross platform concepts.

We do not differentiate between unified and non-unified memory systems. this should be added to static information ASAP.

A lot of this data collection is based on simple timers. Watching the SC store on macos is the correct way to gather some of this information, but requires a detour into rust for macos.

## Why It Works

The InfoGatherer is a generic subsystem which returns a union of metric datatypes. It writes them to an event, which is applied to state. It is currently re-spawned with the worker so each cluster receives the correct information.

As for topology, macOS identifies TB ports with a uuid in SPThunderboltDataType, and also stores remote uuids if it can find them. These changes read that data with the system_profiler, hopefully not so often as to cause notable performance impacts (though this should be tuned) but frequently enough for moderate responsiveness. 
As we can identify TB connections between devices without needing ips attached to each interface, we can remove the network setup script (almost) completely.

### Manual Testing
Spawn RDMA instances without enabling DHCP on the RDMA interfaces.

### Automated Testing
Updated the current master and shared tests to cover the topology refactor and new events.

### DIFF
--- a/README.md
+++ b/README.md
@@ -42,7 +42,7 @@ exo includes a built-in dashboard for managing your cluster and chatting with mo
   <summary>Qwen3-235B (8-bit) on 4 × M3 Ultra Mac Studio with Tensor Parallel RDMA</summary>
   <img src="docs/benchmarks/jeffgeerling/mac-studio-cluster-ai-full-1-qwen3-235b.jpeg" alt="Benchmark - Qwen3-235B (8-bit) on 4 × M3 Ultra Mac Studio with Tensor Parallel RDMA" width="80%" />
   <p>
-    <strong>Source:</strong> <a href="https://www.jeffgeerling.com/blog/2025/15-tb-vram-on-mac-studio-rdma-over-thunderbolt-5">Jeff Geerling: 15 TB VRAM on Mac Studio – RDMA over Thunderbolt 5</a>
+    <strong>Source:</strong> <a href="https://www.jeffgeerling.com/blog/2025/15-tb-vram-on-mac-studio-rdma-over-thunderbolt-5">Jeff Geerling: 15 TB VRAM on Mac Studio – RDMA over Thunderbolt 5</a>
   </p>
 </details>
 
@@ -50,7 +50,7 @@ exo includes a built-in dashboard for managing your cluster and chatting with mo
   <summary>DeepSeek v3.1 671B (8-bit) on 4 × M3 Ultra Mac Studio with Tensor Parallel RDMA</summary>
   <img src="docs/benchmarks/jeffgeerling/mac-studio-cluster-ai-full-2-deepseek-3.1-671b.jpeg" alt="Benchmark - DeepSeek v3.1 671B (8-bit) on 4 × M3 Ultra Mac Studio with Tensor Parallel RDMA" width="80%" />
   <p>
-    <strong>Source:</strong> <a href="https://www.jeffgeerling.com/blog/2025/15-tb-vram-on-mac-studio-rdma-over-thunderbolt-5">Jeff Geerling: 15 TB VRAM on Mac Studio – RDMA over Thunderbolt 5</a>
+    <strong>Source:</strong> <a href="https://www.jeffgeerling.com/blog/2025/15-tb-vram-on-mac-studio-rdma-over-thunderbolt-5">Jeff Geerling: 15 TB VRAM on Mac Studio – RDMA over Thunderbolt 5</a>
   </p>
 </details>
 
@@ -58,7 +58,7 @@ exo includes a built-in dashboard for managing your cluster and chatting with mo
   <summary>Kimi K2 Thinking (native 4-bit) on 4 × M3 Ultra Mac Studio with Tensor Parallel RDMA</summary>
   <img src="docs/benchmarks/jeffgeerling/mac-studio-cluster-ai-full-3-kimi-k2-thinking.jpeg" alt="Benchmark - Kimi K2 Thinking (native 4-bit) on 4 × M3 Ultra Mac Studio with Tensor Parallel RDMA" width="80%" />
   <p>
-    <strong>Source:</strong> <a href="https://www.jeffgeerling.com/blog/2025/15-tb-vram-on-mac-studio-rdma-over-thunderbolt-5">Jeff Geerling: 15 TB VRAM on Mac Studio – RDMA over Thunderbolt 5</a>
+    <strong>Source:</strong> <a href="https://www.jeffgeerling.com/blog/2025/15-tb-vram-on-mac-studio-rdma-over-thunderbolt-5">Jeff Geerling: 15 TB VRAM on Mac Studio – RDMA over Thunderbolt 5</a>
   </p>
 </details>
 
@@ -163,6 +163,24 @@ This starts the exo dashboard and API at http://localhost:52415/
 
 **Important note for Linux users:** Currently, exo runs on CPU on Linux. GPU support for Linux platforms is under development. If you'd like to see support for your specific Linux hardware, please [search for existing feature requests](https://github.com/exo-explore/exo/issues) or create a new one.
 
+**Configuration Options:**
+
+- `--no-worker`: Run exo without the worker component. Useful for coordinator-only nodes that handle networking and orchestration but don't execute inference tasks. This is helpful for machines without sufficient GPU resources but with good network connectivity.
+
+  ```bash
+  uv run exo --no-worker
+  ```
+
+**File Locations (Linux):**
+
+exo follows the [XDG Base Directory Specification](https://specifications.freedesktop.org/basedir-spec/basedir-spec-latest.html) on Linux:
+
+- **Configuration files**: `~/.config/exo/` (or `$XDG_CONFIG_HOME/exo/`)
+- **Data files**: `~/.local/share/exo/` (or `$XDG_DATA_HOME/exo/`)
+- **Cache files**: `~/.cache/exo/` (or `$XDG_CACHE_HOME/exo/`)
+
+You can override these locations by setting the corresponding XDG environment variables.
+
 ### macOS App
 
 exo ships a macOS app that runs in the background on your Mac.
@@ -174,6 +192,19 @@ The macOS app requires macOS Tahoe 26.2 or later.
 Download the latest build here: [EXO-latest.dmg](https://assets.exolabs.net/EXO-latest.dmg).
 
 The app will ask for permission to modify system settings and install a new Network profile. Improvements to this are being worked on.
+
+**Custom Namespace for Cluster Isolation:**
+
+The macOS app includes a custom namespace feature that allows you to isolate your exo cluster from others on the same network. This is configured through the `EXO_LIBP2P_NAMESPACE` setting:
+
+- **Use cases**:
+  - Running multiple separate exo clusters on the same network
+  - Isolating development/testing clusters from production clusters
+  - Preventing accidental cluster joining
+
+- **Configuration**: Access this setting in the app's Advanced settings (or set the `EXO_LIBP2P_NAMESPACE` environment variable when running from source)
+
+The namespace is logged on startup for debugging purposes.
 
 #### Uninstalling the macOS App
 
@@ -318,6 +349,52 @@ For further details, see:
 
 - API basic documentation in [docs/api.md](docs/api.md).
 - API types and endpoints in [src/exo/master/api.py](src/exo/master/api.py).
+
+---
+
+## Benchmarking
+
+The `exo-bench` tool measures model prefill and token generation speed across different placement configurations. This helps you optimize model performance and validate improvements.
+
+**Prerequisites:**
+- Nodes should be running with `uv run exo` before benchmarking
+- The tool uses the `/bench/chat/completions` endpoint
+
+**Basic usage:**
+
+```bash
+uv run bench/exo_bench.py \
+  --model llama-3.2-1b \
+  --pp 128,256,512 \
+  --tg 128,256
+```
+
+**Key parameters:**
+
+- `--model`: Model to benchmark (short ID or HuggingFace ID)
+- `--pp`: Prompt size hints (comma-separated integers)
+- `--tg`: Generation lengths (comma-separated integers)
+- `--max-nodes`: Limit placements to N nodes (default: 4)
+- `--instance-meta`: Filter by `ring`, `jaccl`, or `both` (default: both)
+- `--sharding`: Filter by `pipeline`, `tensor`, or `both` (default: both)
+- `--repeat`: Number of repetitions per configuration (default: 1)
+- `--warmup`: Warmup runs per placement (default: 0)
+- `--json-out`: Output file for results (default: bench/results.json)
+
+**Example with filters:**
+
+```bash
+uv run bench/exo_bench.py \
+  --model llama-3.2-1b \
+  --pp 128,512 \
+  --tg 128 \
+  --max-nodes 2 \
+  --sharding tensor \
+  --repeat 3 \
+  --json-out my-results.json
+```
+
+The tool outputs performance metrics including prompt tokens per second (prompt_tps), generation tokens per second (generation_tps), and peak memory usage for each configuration.
 
 ---
 


src/exo/shared/topology.py
```diff
@@ -37,167 +37,166 @@ def from_snapshot(cls, snapshot: TopologySnapshot) -> "Topology":
...
-        self._node_id_to_rx_id_map[node.node_id] = rx_id
-        self._rx_id_to_node_id_map[rx_id] = node.node_id
+        self._vertex_indices[node] = rx_id
+        self._graph[rx_id] = node
 
     def node_is_leaf(self, node_id: NodeId) -> bool:
         return (
-            node_id in self._node_id_to_rx_id_map
-            and len(self._graph.neighbors(self._node_id_to_rx_id_map[node_id])) == 1
+            node_id in self._vertex_indices
```
rltakashige, line 64: We have inconsistent naming of nodes. In some places, we call them `node_id` and others `node`. 

Could we choose one? I think I prefer `node_id` personally.

src/exo/shared/topology.py
```diff
@@ -37,167 +37,166 @@ def from_snapshot(cls, snapshot: TopologySnapshot) -> "Topology":
...
             cycle = [tb_graph[idx] for idx in cycle_idx]
             cycles.append(cycle)
 
         return cycles
 
-    def get_subgraph_from_nodes(self, nodes: list[NodeInfo]) -> "Topology":
-        node_idxs = [node.node_id for node in nodes]
-        rx_idxs = [self._node_id_to_rx_id_map[idx] for idx in node_idxs]
+    def get_subgraph_from_nodes(self, nodes: list[NodeId]) -> "Topology":
+        node_idxs = [node for node in nodes]
```
rltakashige: Why node_idxs?

src/exo/shared/types/topology.py
```diff
@@ -1,37 +1,32 @@
...
+class LinkType(str, Enum):
+    Thunderbolt = "Thunderbolt"
+    Ethernet = "Ethernet"
+    WiFi = "WiFi"
+
+
+class SocketConnection(FrozenModel):
+    sink_multiaddr: Multiaddr
+
+    def __hash__(self):
```
rltakashige, line 31: Why does SocketConnection need a hash where RDMAConnection does not?

src/exo/master/placement_utils.py
```diff
@@ -214,86 +215,55 @@ def get_mlx_ibv_devices_matrix(
...
-    for interface in node_info.node_profile.network_interfaces:
-        if interface.name not in ["en2", "en3", "en4", "en5", "en6", "en7"]:
-            continue
-        logger.info(f" | {interface.name}: {interface.ip_address}")
-        if interface.ip_address != ip_address:
-            continue
+    # TODO: Prioritise ETHERNET > ??WIFI > TB for coordinator
+    for connection in cycle_digraph.get_all_connections_between(node_i, node_j):
+        if isinstance(connection, SocketConnection):
+            yield connection.sink_multiaddr.ip_address
```
rltakashige: Is it a problem that TB is no longer an option? Probably this is behaviour we want -> just change the comment to reflect.

src/exo/shared/apply.py
```diff
@@ -188,120 +193,99 @@ def apply_runner_deleted(event: RunnerDeleted, state: State) -> State:
...
+        case NodeConfig():
+            pass
+        case MiscData():
+            profile.friendly_name = info.friendly_name
+        case StaticNodeInformation():
+            profile.model_id = info.model
+            profile.chip_id = info.chip
+        # TODO: makes me slightly sad
+        case Sequence():
+            if info != []:
```
rltakashige: if `event == []: break`, just to reduce even one indent here.

src/exo/shared/apply.py
```diff
@@ -188,120 +193,99 @@ def apply_runner_deleted(event: RunnerDeleted, state: State) -> State:
...
+                                ),
+                            )
+                            for tb_conn in info
+                            if tb_conn.source_uuid in conn_map
+                            if tb_conn.sink_uuid in conn_map
+                        ]
+                        topology.replace_all_out_tb_connections(
+                            event.node_id, as_rdma_conns
+                        )
+                        # TODO:
```
rltakashige: TODO?

src/exo/shared/topology.py
Evanev7: good catch, previously these were node_id + profile combos so the naming is all vestigial. node_id is better.

src/exo/shared/types/topology.py
Evanev7: The Multiaddr type is unhashable as-is, where the pair of strings is hashable. I could fix this instead by making Multiaddr inherit from FrozenModel.

src/exo/master/placement_utils.py
Evanev7: did we completely remove TB as an optional coordinator?

src/exo/master/placement.py
```diff
@@ -51,33 +52,34 @@ def place_instance(
...
     all_nodes = list(topology.list_nodes())
 
     logger.info("finding cycles:")
-    cycles = topology.get_cycles()
-    singleton_cycles = [[node] for node in all_nodes]
-    candidate_cycles = list(
-        filter(lambda it: len(it) >= command.min_nodes, cycles + singleton_cycles)
-    )
+    cycles = topology.get_cycles() + [[node] for node in all_nodes]
+    logger.info(cycles)
```
rltakashige: Let's not log too much during placement, given how many times this will be called.

src/exo/shared/topology.py
Evanev7: any opinion on the function names? list_node_ids is clunky, and the underlying graph function is just _graph.nodes()

src/exo/shared/apply.py
Evanev7: oops

src/exo/shared/apply.py
Evanev7: can't break out of a match. I just realised that there are some legit meanings of [] though, so I can't merge this until I've split up these events properly. no network interfaces is possible and is distinct from no TBConnections, and both should be applied to state.

src/exo/shared/topology.py
Evanev7: oops  :)

src/exo/master/placement_utils.py
rltakashige: What confused me was that the change exists but isn't backported yet. This maintains the existing functionality here, so you can mark this comment as resolved.

src/exo/shared/apply.py
rltakashige: Nice catch.

src/exo/shared/topology.py
rltakashige: Was also thinking on this - I think `list_nodes` is fine, given it's typed, as long as we're consistent in writing `for node_id in topology.list_nodes()`. 

(The fact that `list_nodes` returns an iterator is a little surprising to me, and I'd rather something like `iter_nodes`, but we do the same for connections so it's fine)

src/exo/shared/types/topology.py
rltakashige: Would rather that.

Evanev7: oh! the port should not be part of the hash - two socketconnections with the same port are equivalent data. That does mean I should just store the ipaddress here, but we don't have a dedicated ipaddress type on this branch and we just mock multiaddrs everywhere instead.

Evanev7: gm btw haha

src/exo/master/main.py
```diff
@@ -200,9 +201,7 @@ async def _command_processor(self) -> None:
     async def _plan(self) -> None:
         while True:
             # kill broken instances
-            connected_node_ids = set(
-                [x.node_id for x in self.state.topology.list_nodes()]
-            )
+            connected_node_ids = set([x for x in self.state.topology.list_nodes()])
```
JakeHillion:
```suggestion
            connected_node_ids = set(self.state.topology.list_nodes())
```

src/exo/master/placement.py
```diff
@@ -52,19 +54,16 @@ def place_instance(
...
-        filter(lambda it: len(it) >= command.min_nodes, cycles + singleton_cycles)
-    )
+    cycles = topology.get_cycles() + [[node] for node in all_nodes]
+    candidate_cycles = list(filter(lambda it: len(it) >= command.min_nodes, cycles))
     cycles_with_sufficient_memory = filter_cycles_by_memory(
-        candidate_cycles, command.model_meta.storage_size
+        candidate_cycles, node_profiles, command.model_meta.storage_size
     )
-    if not cycles_with_sufficient_memory:
+    if len(cycles_with_sufficient_memory) == 0:
```
JakeHillion, line 64: this change is a no-op as I understand it

src/exo/master/placement.py
```diff
@@ -94,13 +93,15 @@ def place_instance(
     smallest_tb_cycles = [
         cycle
         for cycle in smallest_cycles
-        if topology.get_subgraph_from_nodes(cycle).is_thunderbolt_cycle(cycle)
+        if topology.get_subgraph_from_nodes(
+            [node.node_id for node in cycle]
```
JakeHillion: `node` is confusing here as a name

src/exo/master/placement.py
```diff
@@ -94,13 +93,15 @@ def place_instance(
...
+        ).is_thunderbolt_cycle([node.node_id for node in cycle])
```
JakeHillion: same here about node, and this repeated comprehension is quite ugly

src/exo/master/placement_utils.py
```diff
@@ -214,71 +215,38 @@ def get_mlx_ibv_devices_matrix(
...
-        logger.info(f" | {interface.name}: {interface.ip_address}")
-        if interface.ip_address != ip_address:
-            continue
-
-        logger.info("Found")
-        return f"rdma_{interface.name}"
-
-    return None
+    """Find all IP addresses that connect node i to node j."""
+    # TODO: Prioritise ETHERNET > ??WIFI > TB for coordinator
```
JakeHillion: seems like an incorrect todo

src/exo/master/placement_utils.py
```diff
@@ -214,71 +215,38 @@ def get_mlx_ibv_devices_matrix(
...
                 )
                 raise ValueError(
-                    "Current ibv backend requires all-to-all rdma connections"
+                    "Current jaccl backend requires all-to-all RDMA connections"
                 )
 
     return matrix
 
 
 def _find_connection_ip(
```
JakeHillion, line 279: Existing, but this function is badly named and has a poorly typed return value

src/exo/master/placement_utils.py
```diff
@@ -375,31 +343,32 @@ def get_mlx_ring_hosts_by_node(
...
 
-    def get_ip_for_node(n: NodeInfo) -> str:
-        if n.node_id == rank_0_node.node_id:
+    def get_ip_for_node(n: NodeId) -> str:
+        if n == coordinator:
             return "0.0.0.0"
 
-        ip = _find_ip_prioritised(n, rank_0_node, cycle_digraph)
-        if ip:
+        for ip, _ in _find_connection_ip(n, coordinator, cycle_digraph):
```
JakeHillion: This seems like a regression. Bad rebase?

src/exo/shared/apply.py
```diff
@@ -188,120 +192,90 @@ def apply_runner_deleted(event: RunnerDeleted, state: State) -> State:
...
-    # TODO: NodeCreated
-    if not topology.contains_node(event.node_id):
-        topology.add_node(NodeInfo(node_id=event.node_id))
-    topology.update_node_profile(event.node_id, event.node_profile)
+def apply_node_gathered_info(event: NodeGatheredInfo, state: State) -> State:
+    topology = copy.deepcopy(state.topology)
+    topology.add_node(event.node_id)
+    info = event.info
+    profile = state.node_profiles.get(event.node_id, NodePerformanceProfile())
+    # TODO: should be broken up into individual events instead of this monster
```
JakeHillion: ???

I'm not massively against enums of enums like this if it makes the code clearer, but this is a TODO.

src/exo/shared/constants.py
```diff
@@ -38,6 +38,7 @@ def _get_xdg_dir(env_var: str, fallback: str) -> Path:
 
 # Identity (config)
 EXO_NODE_ID_KEYPAIR = EXO_CONFIG_HOME / "node_id.keypair"
+EXO_CONFIG_FILE = EXO_CONFIG_HOME / "config.toml"
```
JakeHillion, line 41: Generally: the config doesn't appear to be used in this PR at all, other than scaffolding. I'm generally quite against landing dead code like this, especially in a PR that's already very big. Your call to remove it.

src/exo/shared/logging.py
```diff
@@ -24,6 +24,8 @@ def emit(self, record: logging.LogRecord):
         except ValueError:
             level = record.levelno
 
+        return
```
JakeHillion: This is creating dead code. Please remove it.

src/exo/shared/tests/test_apply/test_apply_node_download.py
```diff
@@ -43,7 +43,4 @@ def test_apply_two_node_download_progress():
         NodeDownloadProgress(download_progress=event2), state
     )
 
-    # TODO: This test is failing. We should support the following:
-    # 1. Downloading multiple models concurrently on the same node (one per runner is fine).
-    # 2. Downloading a model, it completes, then downloading a different model on the same node.
```
JakeHillion, line 48: Why was this removed?

src/exo/shared/topology.py
```diff
@@ -1,203 +1,215 @@
...
-from exo.shared.types.profiling import ConnectionProfile, NodePerformanceProfile
-from exo.shared.types.topology import Connection, NodeInfo
+from exo.shared.types.topology import RDMAConnection, SocketConnection
 
 
 class TopologySnapshot(BaseModel):
-    nodes: list[NodeInfo]
-    connections: list[Connection]
+    nodes: Sequence[NodeId]
+    connections: Iterable[tuple[NodeId, NodeId, SocketConnection | RDMAConnection]]
```
JakeHillion: How come we dropped from the strong type `Connection` here?

src/exo/shared/topology.py
```diff
@@ -1,203 +1,215 @@
...
-            for rx_id in self._graph.neighbors(self._node_id_to_rx_id_map[node_id])
+            self._graph[rx_id]
+            for rx_id in self._graph.neighbors(self._vertex_indices[node_id])
         ]
 
-    def out_edges(self, node_id: NodeId) -> list[tuple[NodeId, Connection]]:
-        if node_id not in self._node_id_to_rx_id_map:
+    def out_edges(
+        self, node_id: NodeId
+    ) -> Iterable[tuple[NodeId, SocketConnection | RDMAConnection]]:
```
JakeHillion: Weakened types here as well.

src/exo/shared/types/events.py
```diff
@@ -76,25 +76,15 @@ class RunnerDeleted(BaseEvent):
...
-    node_profile: NodePerformanceProfile
-
-
-class NodeMemoryMeasured(BaseEvent):
+# TODO: bikeshed this naem
+class NodeGatheredInfo(BaseEvent):
     node_id: NodeId
     when: str  # this is a manually cast datetime overrode by the master when the event is indexed, rather than the local time on the device
-    memory: MemoryPerformanceProfile
+    info: GatheredInfo  # NB: this model is UNTAGGED!!! be warned for ser/de errors.
```
JakeHillion: why is this untagged?

src/exo/shared/types/profiling.py
```diff
@@ -53,15 +54,16 @@ class NetworkInterfaceInfo(CamelCaseModel):
...
+    network_interfaces: Sequence[NetworkInterfaceInfo] = []
+    tb_interfaces: Sequence[TBIdentifier] = []
+    system: SystemPerformanceProfile = SystemPerformanceProfile()
 
 
 class ConnectionProfile(CamelCaseModel):
-    throughput: float
-    latency: float
-    jitter: float
+    pass
```
JakeHillion: This is dead now? Delete it

src/exo/shared/types/thunderbolt.py
```diff
@@ -0,0 +1,75 @@
...
+
+class TBIdentifier(CamelCaseModel):
+    rdma_interface: str
+    domain_uuid: str
+
+
+## Intentionally minimal, only collecting data we care about - there's a lot more
+
+
+class TBReceptacleTag(BaseModel, extra="ignore"):
```
JakeHillion: nit: I belive Pascal case dictates that acronyms should be modified, such that this would be `TbReceptacleTag`. I think it's more readable that way too, but your call.

src/exo/shared/types/topology.py
```diff
@@ -1,37 +1,32 @@
...
+class RDMAConnection(FrozenModel):
+    source_rdma_iface: str
+    sink_rdma_iface: str
+
+    def is_thunderbolt(self) -> bool:
+        logger.warning("duh")
+        return True
+
+
+# TODO
```
JakeHillion: why?

src/exo/utils/info_gatherer/info_gatherer.py
```diff
@@ -0,0 +1,232 @@
...
+
+
+class MacTBConnections(TaggedModel):
+    conns: Sequence[TBConnection]
+
+
+class NodeConfig(TaggedModel):
+    """Node configuration from EXO_CONFIG_FILE, reloaded from the file only at startup. Other changes should come in through the API and propagate from there"""
+
+    # TODO
```
JakeHillion: ?

src/exo/utils/info_gatherer/info_gatherer.py
```diff
@@ -0,0 +1,232 @@
...
+            try:
+                contents = (await f.read()).decode("utf-8")
+                data = tomllib.loads(contents)
+                return cls.model_validate(data)
+            except (tomllib.TOMLDecodeError, UnicodeDecodeError):
+                logger.warning("Invalid config file, skipping...")
+                return None
+
+
+class MiscData(TaggedModel):
```
JakeHillion, line 78: This is a concerning name, could we be more specific? Things named misc like this tend to grow uncontrollably.

src/exo/utils/info_gatherer/info_gatherer.py
```diff
@@ -0,0 +1,232 @@
...
+    misc_poll_interval: float | None = 60
+    system_profiler_interval: float | None = 5 if IS_DARWIN else None
+    memory_poll_rate: float | None = None if IS_DARWIN else 1
+    macmon_interval: float | None = 1 if IS_DARWIN else None
+    _tg: TaskGroup = field(init=False, default_factory=create_task_group)
+
+    async def run(self):
+        async with self._tg as tg:
+            if (macmon_path := shutil.which("macmon")) is not None:
+                tg.start_soon(self._monitor_macmon, macmon_path)
```
JakeHillion: Should this be under IS_DARWIN too?

src/exo/utils/info_gatherer/info_gatherer.py
```diff
@@ -0,0 +1,232 @@
...
+    async def _monitor_system_profiler(self):
+        if self.system_profiler_interval is None:
+            return
+        iface_map = await _gather_iface_map()
+        if iface_map is None:
+            return
+
+        old_idents = []
+        while True:
+            data = await TBConnectivity.gather()
```
JakeHillion: Shouldn't this be called SystemProfiler.gather?

src/exo/utils/info_gatherer/info_gatherer.py
```diff
@@ -0,0 +1,232 @@
...
+            nc = await NodeConfig.gather()
+            if nc is not None:
+                await self.info_sender.send(nc)
+            sni = await StaticNodeInformation.gather()
+            await self.info_sender.send(sni)
+
+    def shutdown(self):
+        self._tg.cancel_scope.cancel()
+
+    async def _monitor_misc(self):
```
JakeHillion, line 149: Generally: it would be nice if all these looped polls could share some code instead of being duplicated.

src/exo/utils/info_gatherer/info_gatherer.py
```diff
@@ -0,0 +1,232 @@
...
+            if override_memory_env
+            else None
+        )
+        if self.memory_poll_rate is None:
+            return
+        while True:
+            await self.info_sender.send(
+                MemoryUsage.from_psutil(override_memory=override_memory)
+            )
+            await anyio.sleep(self.memory_poll_rate)
```
JakeHillion, line 196: This seems like a very inefficent way to report memory overrides. If memory is overridden, we could send it once and never enter this loop - unless I'm misunderstanding the code. It's not _wrong_ though.

JakeHillion (changes requested):
The core direction here is solid - the info_gatherer abstraction is a good foundation, and I appreciate that this has been well tested in practice.

Summarising the main concerns:
1. Topology types are weaker - The unified Connection type is gone, replaced by (source, sink, connection) tuples throughout. This loses type safety and makes the API much harder to follow. I'd like to understand the motivation here.
2. Dead code / incomplete features - There's code that's unreachable (e.g. early return cutting off logic), empty classes (ConnectionProfile, NodeConfig), and scaffolding for config that isn't used yet. Also a bunch of TODOs throughout the new code, most of which aren't actionable. This should all really be cleaned up before merging.

Happy to discuss anything in detail that's unclear to get the design right, or we can iterate on the PR. I think this would benefit from a cleanup pass before it's ready for detailed review again - things like the no-op changes, unreachable code, and config addition make it hard to focus on the actual design of the core feature here - the drastically improved Thunderbolt detection. If there are self-contained pieces (like the info_gatherer refactor for existing data or the Ibv -> Jaccl rename) that could be split into separate PRs, that would make each one easier to review and merge independently, helping us all move fast.

src/exo/master/placement.py
Evanev7: it's clearer, and less likely to break if we introduce a dedicated Cycle class.

src/exo/master/placement_utils.py
Evanev7: Will follow up later

src/exo/shared/constants.py
Evanev7: I'm leaving it in, I want to build on top of it.

src/exo/shared/tests/test_apply/test_apply_node_download.py
Evanev7: The test isn't failing, and we support both of these AFAIK.

src/exo/shared/topology.py
Evanev7: fixed

src/exo/shared/types/events.py
Evanev7: bug in comment

src/exo/shared/types/thunderbolt.py
Evanev7: This is consistent with macOS's handling of PascalCase, though I'm just going to unshorten TB to Thunderbolt

src/exo/utils/info_gatherer/info_gatherer.py
Evanev7: Is there a problem with this growing? If we want metadata for the dashboard we need to collect it somewhere and I'd rather here than some random function elsewhere in the codebase

src/exo/utils/info_gatherer/info_gatherer.py
Evanev7: I don't see why it should

PR #936, opened 2025-12-20
Add Prerequisites to Readme

nightguarder:
## Motivation
Users need to know what **prerequisites** they need in order to run exo. Simple addition to docs prevents future raised issues.

## Changes

Updated ``README.md``:
 - to include installation instructions for **[uv](https://github.com/astral-sh/uv)** and **[macmon](https://github.com/vladkens/macmon)**.

Updated ``CONTRIBUTING.md``:
-  to verify these prerequisites are met before starting development.

- Standardized on brew installation instructions for macOS users to keep the guide simple.

## Why It Works

By listing these prerequisites upfront, users will set up their environment correctly before attempting to run exo.

### Manual Testing
MacBook Pro M4
- Verified that ``uv`` and ``macmon`` were missing initially, causing failures
- after installing them via brew (as documented), uv run exo starts successfully.

### DIFF
--- a/src/exo/worker/utils/macmon.py
+++ b/src/exo/worker/utils/macmon.py
@@ -1,6 +1,7 @@
 import platform
 import shutil
 from subprocess import CalledProcessError
+from typing import cast
 
 from anyio import run_process
 from pydantic import BaseModel, ConfigDict, ValidationError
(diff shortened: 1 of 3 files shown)


Evanev7:
looks good, but I'd actually prefer people use the nix flake for development. I'm on the fence as it's a bit of a turn off for newcomers, but the readme changes definitely seem good.

nightguarder:
Sorry Not familiar with ``nix``. Will look into it. I just wanted to quickly get working on the project.

Evanev7:
no worries! this is the exact problem I'm describing, it's not simple like `brew` so I think we probably should continue with brew in the CONTRIBUTING - and I think we should aim to support contributions from both dev environments as well

PR #938, opened 2025-12-20
Extend eos_token_id fix for other models

rltakashige:
## Motivation

We currently use mlx_lm's load_tokenizer instead of load. This means that some models are missing some configurations, such as eos_token_id. This is clear for a model like GLM, which does not finish token generation.

## Changes

A small stopgap, to allow eos_token_ids to be added, and a TODO for us to migrate to load. The reason we don't want to do this now is that a solid testing framework is not configured in this repo yet.

## Why It Works

It just uses the eos_token_ids I obtained from loading a tokenizer in mlx_lm and calling `tokenizer.eos_token_ids` .

### Manual Testing
Tested on several Macs.

### Automated Testing
None yet, as described.


Evanev7:
I ran the formatter so ci should pass now.

PR #959, opened 2025-12-22
migrate from libp2p to iroh

Evanev7:
## Motivation

libp2p is fairly far abstracted from networking hardware. Iroh provides an alternative, and this branch has been my work on a potential future migration. I may revive the attempt at some point, but libp2p is working well at the moment. I think the code quality here is better, as is the lack of the nightly rust toolchain, but it doesn't solve any problems we are facing right now.

## Changes

Removed our libp2p dependency, removed our rust-nightly dependency, added an iroh-gossip based exo net implementation.

## Test Plan

Minimal - basic functionality is there, but little testing.

### DIFF
--- a/src/exo/master/tests/conftest.py
+++ b/src/exo/master/tests/conftest.py
@@ -1,7 +1,6 @@
 from ipaddress import ip_address
 from itertools import count
 
-from exo.routing.connection_message import SocketAddress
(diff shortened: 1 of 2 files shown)


AndreiCravtov:
i loooove iroh

image

PR #961, opened 2025-12-22
fix: clamp download progress percentage to 0-100 range

### LINKED ISSUE #933: [BUG] Download Progress exceeding 100% through downloading a model for multiple devices.
## To Reproduce

1. Run Exo on multiple devices
2. Start download a model with my options: Tensor, MLX Ring and 3 Devices. Download the model Llama 3.2 3B (8b) 3GB

## Expected behavior

Expected value for download progress should be the mean value of every progress.

## Actual behavior

Through observation I'm thinking that it is the sum and not the mean value of progress.

## Environment

- macOS Version: macOS Tahoe 26.1
- EXO Version: 1.0.59
- Hardware:
  - 3 MacBook Pro M4 16GB
- Interconnection:
  - WiFi 6 with every device.

Image

majiayu000:
## Summary
- Clamp download progress percentage to 0-100 range to prevent display issues when downloading across multiple devices
- Applied fix to both Dashboard (TypeScript/Svelte) and macOS App (Swift)

Fixes #933

## Test plan
- [x] Verified existing Python tests pass
- [x] Verified TypeScript type checking passes (existing errors unrelated to this change)

### DIFF
--- a/app/EXO/EXO/ViewModels/InstanceViewModel.swift
+++ b/app/EXO/EXO/ViewModels/InstanceViewModel.swift
@@ -10,11 +10,11 @@ struct DownloadProgressViewModel: Equatable {
 
     var fractionCompleted: Double {
         guard totalBytes > 0 else { return 0 }
-        return Double(downloadedBytes) / Double(totalBytes)
+        return min(1.0, max(0.0, Double(downloadedBytes) / Double(totalBytes)))
     }
 
     var percentCompleted: Double {
-        fractionCompleted * 100
+        min(100.0, fractionCompleted * 100)
     }
 
     var formattedProgress: String {
--- a/dashboard/src/routes/+page.svelte
+++ b/dashboard/src/routes/+page.svelte
@@ -382,6 +382,11 @@ function toggleInstanceDownloadDetails(nodeId: string): void {
 		return (meta.model_id as string) ?? (meta.modelId as string) ?? null;
 	}
 
+	// Helper to clamp percentage between 0 and 100
+	function clampPercent(value: number): number {
+		return Math.min(100, Math.max(0, value));
+	}
+
 	// Helper to parse download progress from payload
 	function parseDownloadProgress(payload: Record<string, unknown>): DownloadProgress | null {
 		const progress = payload.download_progress ?? payload.downloadProgress;
@@ -408,16 +413,16 @@ function toggleInstanceDownloadDetails(nodeId: string): void {
 				downloadedBytes: fDownloaded,
 				speed: (fd.speed as number) ?? 0,
 				etaMs: (fd.eta_ms as number) ?? (fd.etaMs as number) ?? 0,
-				percentage: fTotal > 0 ? (fDownloaded / fTotal) * 100 : 0
+				percentage: clampPercent(fTotal > 0 ? (fDownloaded / fTotal) * 100 : 0)
 			});
 		}
-		
+
 		return {
 			totalBytes,
 			downloadedBytes,
 			speed,
 			etaMs: etaMs || (speed > 0 ? ((totalBytes - downloadedBytes) / speed) * 1000 : 0),
-			percentage: totalBytes > 0 ? (downloadedBytes / totalBytes) * 100 : 0,
+			percentage: clampPercent(totalBytes > 0 ? (downloadedBytes / totalBytes) * 100 : 0),
 			completedFiles,
 			totalFiles,
 			files
@@ -496,7 +501,7 @@ function toggleInstanceDownloadDetails(nodeId: string): void {
 				downloadedBytes,
 				speed: totalSpeed,
 				etaMs: totalSpeed > 0 ? ((totalBytes - downloadedBytes) / totalSpeed) * 1000 : 0,
-				percentage: totalBytes > 0 ? (downloadedBytes / totalBytes) * 100 : 0,
+				percentage: clampPercent(totalBytes > 0 ? (downloadedBytes / totalBytes) * 100 : 0),
 				completedFiles,
 				totalFiles,
 				files: allFiles
@@ -603,7 +608,7 @@ function toggleInstanceDownloadDetails(nodeId: string): void {
 				downloadedBytes,
 				speed: totalSpeed,
 				etaMs: totalSpeed > 0 ? ((totalBytes - downloadedBytes) / totalSpeed) * 1000 : 0,
-				percentage: totalBytes > 0 ? (downloadedBytes / totalBytes) * 100 : 0,
+				percentage: clampPercent(totalBytes > 0 ? (downloadedBytes / totalBytes) * 100 : 0),
 				completedFiles,
 				totalFiles,
 				files: allFiles
--- a/src/exo/worker/download/download_utils.py
+++ b/src/exo/worker/download/download_utils.py
@@ -418,6 +418,19 @@ def calculate_repo_progress(
     all_downloaded_bytes = sum(
         (p.downloaded.in_bytes for p in file_progress.values()), 0
     )
+    # Defensive check: ensure downloaded never exceeds total to prevent >100% progress
+    if all_total_bytes > 0 and all_downloaded_bytes > all_total_bytes:
+        overflow_files = [
+            (p.file_path, p.downloaded.in_bytes, p.total.in_bytes)
+            for p in file_progress.values()
+            if p.downloaded.in_bytes > p.total.in_bytes
+        ]
+        if overflow_files:
+            logger.warning(
+                f"Download progress overflow detected: {all_downloaded_bytes}/{all_total_bytes} bytes. "
+                f"Files with overflow: {overflow_files}"
+            )
+        all_downloaded_bytes = all_total_bytes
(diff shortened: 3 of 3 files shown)


app/EXO/EXO/ViewModels/InstanceViewModel.swift
```diff
@@ -10,11 +10,11 @@ struct DownloadProgressViewModel: Equatable {
 
     var fractionCompleted: Double {
         guard totalBytes > 0 else { return 0 }
-        return Double(downloadedBytes) / Double(totalBytes)
+        return min(1.0, max(0.0, Double(downloadedBytes) / Double(totalBytes)))
```
DevKiDCosmo, line 13: This prevents an overflow but doesn't calculate the correct percentage if the previous couldn't.

app/EXO/EXO/ViewModels/InstanceViewModel.swift
```diff
@@ -10,11 +10,11 @@ struct DownloadProgressViewModel: Equatable {
...
     }
 
     var percentCompleted: Double {
-        fractionCompleted * 100
+        min(100.0, fractionCompleted * 100)
```
DevKiDCosmo, line 17: This also won't help if the `fractionCompleted` will exceed 1. `fractionCompleted` needs to be averaged. The min will fix the overflow but it is not mathematically correct.

dashboard/src/routes/+page.svelte
```diff
@@ -382,6 +382,11 @@ function toggleInstanceDownloadDetails(nodeId: string): void {
 		return (meta.model_id as string) ?? (meta.modelId as string) ?? null;
 	}
 
+	// Helper to clamp percentage between 0 and 100
+	function clampPercent(value: number): number {
+		return Math.min(100, Math.max(0, value));
+	}
+
```
DevKiDCosmo, line 389: Same here. But I think the percentage is calculated somewhere else. The percentage is calculated in `ModelCard.svelte`

DevKiDCosmo: The use of the `clampPercent` function won't fix the problem.

majiayu000:
Thanks for the feedback @DevKiDCosmo! You're absolutely right - I've updated the PR to fix the root cause instead of just clamping.

## Root Cause Analysis

After investigating, I found the real issue in `download_utils.py`:

```python
# Line 637 (before fix)
total=Memory.from_bytes(file.size or 0),  # BUG: file.size=None → total=0
```

HuggingFace API sometimes returns `file.size = None`. The code used `file.size or 0`, setting total to 0. But `downloaded_bytes` came from actual disk file size (e.g., 512MB). This caused:

```
total = 0, downloaded = 512MB → progress = ∞%
```

## The Fix

1. **Handle `file.size = None` properly** - Use `downloaded_bytes` as fallback for total when size info is unavailable:
   ```python
   effective_total = file.size if file.size is not None else downloaded_bytes
   ```

2. **Fix status comparison** - The old code `downloaded_bytes == file.size` fails when `file.size` is `None`:
   ```python
   if file.size is not None:
       status = "complete" if downloaded_bytes == file.size else "not_started"
   else:
       status = "complete" if downloaded_bytes > 0 else "not_started"
   ```

3. **Add defensive logging** - If overflow still occurs (edge cases), log a warning with details for debugging.

The frontend `clampPercent` is kept as a safety net, but the backend now calculates mathematically correct values.

src/exo/worker/download/download_utils.py
```diff
@@ -628,16 +641,24 @@ def on_progress_wrapper(
 
     for file in filtered_file_list:
         downloaded_bytes = await get_downloaded_size(target_dir / file.path)
+        # Fix: When file.size is None, use downloaded_bytes as total to avoid progress overflow
+        # This ensures downloaded <= total when size info is unavailable from HuggingFace API
+        effective_total = file.size if file.size is not None else downloaded_bytes
```
DevKiDCosmo, line 646: It would be better if a retry is implemented to get the real size instead of the download size. If the download size is used then the progress would always be 100% because it will check with the downloaded size. Instead of using 100%, we could use UNDEFINED or UNKNOWN.

src/exo/worker/download/download_utils.py
```diff
@@ -418,6 +418,19 @@ def calculate_repo_progress(
...
+            (p.file_path, p.downloaded.in_bytes, p.total.in_bytes)
+            for p in file_progress.values()
+            if p.downloaded.in_bytes > p.total.in_bytes
+        ]
+        if overflow_files:
+            logger.warning(
+                f"Download progress overflow detected: {all_downloaded_bytes}/{all_total_bytes} bytes. "
+                f"Files with overflow: {overflow_files}"
+            )
+        all_downloaded_bytes = all_total_bytes
```
DevKiDCosmo, line 433: When an overflow happens only a warning is logged. But it will happen if the file size is None. So getting the right filesize is crucial if we want to implement such. Also exceeding `all_total_bytes` should not happening if the filesize is sent by Huggingface. It is just a though.

src/exo/worker/download/download_utils.py
DevKiDCosmo: if !file.size -> status = "UNKOWN" || "UNDEFINED"; effective_total = -1.

DevKiDCosmo:
How can I be a contributor? I want to contribute more.

Evanev7:
> How can I be a contributor? I want to contribute more.

You're doing great! The repo is essentially 3 days old so I doubt we're going to push anything yet, but it's nice to see a familiar name in the issues.

Evanev7:
This review is perfect though, no comments from me.

src/exo/worker/download/download_utils.py
Evanev7: I'd rather not use -1 as a sentinel value here - probably update the filedownloadprogress to make the total optional.

majiayu000:
This fixes the >100%/∞ symptom, but it still treats "unknown size" as "complete."\nWhen  is , using  as total and marking status  will show 100% even if the file is only partially downloaded. That’s still mathematically/semantically incorrect.\n\nSuggestion:\n- Only include files with known __TEXT	__DATA	__OBJC	others	dec	hex in  / percentage.\n- Track  separately (count + downloaded bytes), and show "X files size unknown" or an indeterminate progress state in UI when needed.\n- Optional: retry/HEAD to fetch size for , otherwise mark status as  instead of .\n\nThis keeps progress accurate without hiding edge cases.

DevKiDCosmo:
That would work. I would implement a separate UI Block that keeps track of unknown size of downloads. Also I'd implement a bg daemon that checks if unknowns exists and retries. That should handle every case

src/exo/worker/download/download_utils.py
DevKiDCosmo: To what would you update the filedownloadprogress var? -1 would is a universal code for proceeding with failure.

DevKiDCosmo:
Everybody ok if I add to `RepoDownloadProgress` new properties like `downloadUnkownSize` or `downloadUnkownFiles` etc? I would then add to `+page.svelte` or update it with a new additional box for unknowns. Then we have not to fight for statuses like unknown or undefined.

src/exo/worker/download/download_utils.py
Evanev7: I can see your C++ experience! I'd propagate the None file size to the RepoDownloadProgress and fix the downstream code that expects it to always be present. The type checker will catch an int | None, it won't catch a check for -1.

Evanev7:
this is really growing to be two separate issues - the max size is sometimes None and the dashboard is 0 - N * 100% rather than 0 - 100%. I'd rather we fixed the latter now, and moved the former into a new PR at this point.

DevKiDCosmo:
A third bug was mentioned in the original issue. The display graph error with the launch button. So we will fix now the download worker but the dashboard and so will get their own issue, right?

Summaries: This will only contain the download worker fix (resize to 0-100 and separate download history and optional with the bg daemon for retrying retrieving inf.) and another issue is open for the displaying.

Any misunderstanding?

Evanev7:
> A third bug was mentioned in the original issue. The display graph error with the launch button. So we will fix now the download worker but the dashboard and so will get their own issue, right?
> 
> Summaries: This will only contain the download worker fix (resize to 0-100 and separate download history and optional with the bg daemon for retrying retrieving inf.) and another issue is open for the displaying.
> 
> Any misunderstanding?

Yes, I think there has been some misunderstanding.

I suggest we keep this PR to just the original issue - #933, which you suggested fixing in #965  though closed that PR. That is specifically the downloads exceeding 100%, and potentially the placement previews being buggy. This should be as straightforward as dividing download progress by the number of nodes participating in the download.

The fact that the huggingface api returns no file.size is the issue I am suggesting we separate out into a separate PR, as we clearly have some disagreements about the proper way to handle this is.

DevKiDCosmo:
If so, I would create a new issue regarding Hugginface and the bg daemon. So let's close this and merge.

Evanev7:
Closing as stale for now, things are moving quickly and I want to let other team members move forward elsewhere.

PR #962, opened 2025-12-22
feat: respect http_proxy and https_proxy env variables

### LINKED ISSUE #518: Request:  respect `https_proxy` and `http_proxy` env variables for network requests
### Description
I encountered an issue where `exo` does not respect the `https_proxy` and `http_proxy` environment variables for network requests, even though the system environment is properly configured.

The root cause appears to be the way `aiohttp` is used in the `exo` codebase. `aiohttp` supports proxy usage via environment variables through setting `trust_env=True` for aiohttp.ClientSession . However, it seems the current implementation in `exo` does not leverage this feature.

This limitation prevents users from routing traffic through a proxy, which is critical in certain network setups such as corporate environments or restricted networks.

### Steps to Reproduce
1. Set up proxy environment variables:
   ```bash
   export http_proxy=http://proxy-server:port
   export https_proxy=http://proxy-server:port
```
2.Run exo with a command that makes network requests (e.g., model downloading).
3.Observe that the requests do not route through the specified proxy, and connections fail in environments where direct access is blocked.

### Expected Behavior
`exo` should honor the `http_proxy`, `https_proxy`, and no_proxy environment variables automatically when making network requests.

### Actual Behavior
`exo` fails to use the proxy settings, resulting in connection errors in proxied environments.

### Proposed Solution
Update the `aiohttp` usage in `exo` to automatically respect environment proxy settings. This can be achieved by ensuring that  `trust_env=True` is  explicitly passed to `aiohttp.ClientSession()` like this:

```python
async with aiohttp.ClientSession(trust_env=True) as session:
        async with session.get(url) as response:
            print(await response.text())
```

### Environment
`exo` version:  latest main branch
aiohttp version:  aiohttp-3.10.11
Operating System:  windows11+wsl(ubuntu 20.04)
Python Version:  3.12

majiayu000:
## Summary
- Add `trust_env=True` to aiohttp.ClientSession for proxy environment variable support
- Added tests to verify the configuration

Fixes #518

## Test plan
- [x] Added test for trust_env configuration
- [x] All existing tests pass

### DIFF
--- a/src/exo/worker/download/download_utils.py
+++ b/src/exo/worker/download/download_utils.py
@@ -276,6 +276,7 @@ def create_http_session(
             sock_read=sock_read_timeout,
             sock_connect=sock_connect_timeout,
         ),
+        trust_env=True,  # Respect http_proxy, https_proxy, no_proxy env variables
     )
 
 
--- a/src/exo/worker/download/tests/test_http_session.py
+++ b/src/exo/worker/download/tests/test_http_session.py
@@ -1,0 +1,31 @@
+"""Tests for HTTP session configuration."""
+
+import pytest
+
+from exo.worker.download.download_utils import create_http_session
+
+
+@pytest.mark.asyncio
+async def test_create_http_session_trusts_env():
+    """Verify that create_http_session creates a session with trust_env=True.
+
+    This ensures the session respects http_proxy, https_proxy, and no_proxy
+    environment variables for network requests.
+    """
+    session = create_http_session()
+    try:
+        assert session.trust_env is True, (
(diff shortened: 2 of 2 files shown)


Evanev7:
If you need this, can you make it an opt-in command line argument? I see no reason for this to be the default as people using homelabs will presumably not be using the app.

majiayu000:
Thanks for the feedback. I’m going to close this PR for now. If this is still desired later, I can reopen or submit a new PR with an opt‑in flag (e.g., --trust-env) so it’s not default for homelab users.

PR #963, opened 2025-12-22
fix: update macOS app to use correct API port (52415)

### LINKED ISSUE #960: macOS: EXO.app doesn’t start API server + README uses outdated API port
This issue is the result of installing Exo on a Mac M4 Pro desktop and an M1 MacBook Pro laptop, and asking ChatGPT to solve issues. Most of this was dictated by ChatGPT:
Description:
While setting up Exo on macOS, I ran into a couple of small but confusing issues that seem worth documenting or fixing.
1. EXO.app does not expose the API server
On macOS, double-clicking EXO.app launches the node/agent, but does not start the HTTP API server. As a result, endpoints like /instance/previews are unavailable unless Exo is started manually via the CLI.
Running:
/Applications/EXO.app/Contents/Resources/exo/exo --verbose
does start the API correctly.
This behaviour isn’t currently documented and can make it look like Exo isn’t running.
2. README references outdated API port
The README suggests querying:
http://localhost:52415/instance/previews
However, the current macOS build reports:
Dashboard & API Ready
http://localhost:8000
So 52415 appears to be outdated (or platform/version-specific).
3. (Minor) MacMon warning on macOS
On startup, Exo logs:
MacMon not found in PATH
This appears non-fatal (API and worker continue running), but it may be confusing to users.
Environment
macOS (Apple Silicon)
Installed via EXO.app
Exo version: 1.0.59.

Exo is running fine on http://localhost:8000 after following a series of instructions from ChatGPT.

majiayu000:
## Summary
- Update macOS app to use the correct default API port (52415) instead of 8000
- Fixed in ClusterStateService, ContentView, and BugReportService

Fixes #960

## Test plan
- [x] Verified the default port in main.py is 52415
- [x] Updated all Swift files that reference the API port

### DIFF
--- a/app/EXO/EXO/ContentView.swift
+++ b/app/EXO/EXO/ContentView.swift
@@ -212,7 +212,7 @@ struct ContentView: View {
 
     private var dashboardButton: some View {
         Button {
-            guard let url = URL(string: "http://localhost:8000/") else { return }
+            guard let url = URL(string: "http://localhost:52415/") else { return }
             NSWorkspace.shared.open(url)
         } label: {
             HStack {
--- a/app/EXO/EXO/Services/BugReportService.swift
+++ b/app/EXO/EXO/Services/BugReportService.swift
@@ -35,7 +35,7 @@ struct BugReportService {
     }
 
     func sendReport(
-        baseURL: URL = URL(string: "http://127.0.0.1:8000")!,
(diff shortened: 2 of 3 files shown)


AlexCheema:
lgtm. thanks for the contribution!

majiayu000:
Happy to help.

PR #965, opened 2025-12-22
fix: Download Progress exceeding 100% through downloading a model for multiple devices.

DevKiDCosmo:
## Motivation

Fixing bug https://github.com/exo-explore/exo/issues/933. For further information, go to the issue.

## Changes

I add a helper function that takes the value of every node and calculates the corresponding mean value. (Only for speed and for downloaded bytes)

## Why It Works

It creates a map all progress values and calculates the mean for speed and downloaded bytes. So the progress will stay in the range of 0 and 100 and the eta etc. are now correctly calculated.

## Test Plan

I can't test somehow.

### Manual Testing

Run the code on every of the device with

- macOS Version: macOS Tahoe 26.1
- EXO Version: 1.0.59
- Hardware:
  - 3 MacBook Pro M4 16GB
- Interconnection:
  - WiFi 6 with every device.

### Automated Testing

CI/CD

### DIFF
--- a/app/EXO/EXO/ViewModels/InstanceViewModel.swift
+++ b/app/EXO/EXO/ViewModels/InstanceViewModel.swift
@@ -135,6 +135,7 @@ extension ClusterState {
         var totalCompletedFiles = 0
         var totalFileCount = 0
         var hasActiveDownload = false
+        var activeNodeCount = 0
 
         for nodeId in nodeIds {
             guard let nodeDownloads = downloads[nodeId] else { continue }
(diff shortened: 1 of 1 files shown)


app/EXO/EXO/ViewModels/InstanceViewModel.swift
```diff
@@ -135,6 +135,7 @@ extension ClusterState {
         var totalCompletedFiles = 0
         var totalFileCount = 0
         var hasActiveDownload = false
+        var activeNodeCount = 0
```
DevKiDCosmo, line 138: This keeps track of node for averaging.

app/EXO/EXO/ViewModels/InstanceViewModel.swift
```diff
@@ -149,19 +150,26 @@ extension ClusterState {
                 }
                 totalCompletedFiles += progress.completedFiles ?? 0
                 totalFileCount += progress.totalFiles ?? 0
+                activeNodeCount += 1
             }
         }
 
         guard hasActiveDownload else { return nil }
 
+        // Mean across active nodes to avoid inflated aggregates
```
DevKiDCosmo, line 159: Looks more or less fine.

Issue #974, opened 2025-12-22
Still unexpected dependencies when building from source


davmonk:
when attempting to build from source on macOS, there are unstated dependencies including: 
uv
npm
there may be more
could you please approach the documentation on a clean system?

Evanev7:
Closed by #975

PR #975, opened 2025-12-22
Add all prerequisites to README

rltakashige:
## Motivation

Addresses #974 
```
INFO: pip is looking at multiple versions of exo to determine which version is compatible with other requirements. This could take a while.
ERROR: Could not find a version that satisfies the requirement exo-pyo3-bindings (from exo) (from versions: none)
ERROR: No matching distribution found for exo-pyo3-bindings
```

## Changes

Describes Rust dependency for building from source

### Manual Testing

Tested locally and runs after this setup without exo-pyo3-bindings error

### DIFF
--- a/README.md
+++ b/README.md
@@ -64,6 +64,10 @@ There are two ways to run exo:
 ### Run from Source (Mac & Linux)
 
 **Prerequisites:**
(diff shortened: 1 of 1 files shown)


rltakashige:
Not sure whether I should add the npm dependency to the README here too.

PR #981, opened 2025-12-22
ci: migrate build-app to github hosted runners

JakeHillion:
## Motivation

`build-app.yml` was running on self-hosted runners to get the Xcode262_beta toolchain. This is out of beta now and [supported by the GitHub hosted Runners](https://github.com/actions/runner-images/blob/main/images/macos/macos-26-arm64-Readme.md#xcode) so let's switch to those for ease of management.

## Changes

- Switched to GitHub hosted runner.
- Selected the right xcode toolchain & moved the metal test.
- Re-ordered the workflow to install all dependencies at the beginning, making it fail fast if any are missing.
- Added previously missing dashboard build with npm.
- Added previously missing `packaging/pyinstaller/exo.spec` file.
- Removed some stale/cleanup steps as these runners are ephemeral.

### Manual Testing

Modified my branch to build the DMG without a tag (or publishing it). The result is: https://github.com/exo-explore/exo/actions/runs/20439809785/job/58729834295

On 3x Mac Studio M3 Ultra 96GB connected with Thunderbolt:
- Installed the 0.0.0-alpha.0 app. It immediately requested I update to the latest from Sparkle, which is good.
- Clicked the dashboard link from the tray. Correctly opened "localhost:52415" and rendered the dashboard.
- Downloaded "Llama 3.3 70B (8-bit)". Downloaded successfully.

Things that aren't working:
- Running any model. They get stuck in UNKNOWN and we get no logs. If I `uv run exo` from the same commit this is not an issue.
- Discovery seems to be broken. I can't get these machines to pick each other up when connected all-to-all with Thunderbolt 5 & RDMA should be working.

I think it's a better idea to land this as is then debug - it's already quite a messy change and does get the DMG building at least. Note that landing this change _does not_ release this app, we have to push a tag for that.

### Automated Testing

### DIFF
--- a/.github/workflows/build-app.yml
+++ b/.github/workflows/build-app.yml
@@ -7,7 +7,7 @@ on:
 
 jobs:
   build-macos-app:
-    runs-on: [self-hosted, XCode262_Beta]
+    runs-on: "macos-26"
     env:
       SPARKLE_VERSION: 2.8.1
       SPARKLE_DOWNLOAD_PREFIX: ${{ secrets.SPARKLE_DOWNLOAD_PREFIX }}
@@ -21,6 +21,10 @@ jobs:
       EXO_LIBP2P_NAMESPACE: ${{ github.ref_name }}
 
     steps:
+      # ============================================================
+      # Checkout and tag validation
+      # ============================================================
+
       - name: Checkout
         uses: actions/checkout@v4
         with:
@@ -29,7 +33,6 @@ jobs:
       - name: Derive release version from tag
         run: |
           VERSION="${GITHUB_REF_NAME#v}"
-          # Detect alpha tags
           if [[ "$VERSION" == *-alpha* ]]; then
             echo "IS_ALPHA=true" >> $GITHUB_ENV
           else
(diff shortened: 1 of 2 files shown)


.github/workflows/build-app.yml
```diff
@@ -40,53 +43,54 @@ jobs:
...
           fi
-          echo "Metal toolchain is installed."
-
-      - name: Install Just
-        run: brew install just
 
-      - name: Install AWS CLI
-        run: brew install awscli
+      - name: Install Homebrew packages
+        run: brew install just awscli macmon node
```
Evanev7: npm is included in the macos runner by default - do we want to use the brew one just for consistency?

JakeHillion: Nice! I can clean this up. Hopefully in a future change we'll Nix build the dashboard anyway.

JakeHillion:
> have some thoughts on the pyinstaller but I presume you just checked it out from beta_1. lgtm otherwise.

Yep, that's cleanly transplanted, missed it in the first PR. Probably best to land it as is then iterate.

PR #964, opened 2025-12-22
feat: conform to XDG Base Directory Specification on Linux

majiayu000:
## Summary
Implements XDG Base Directory Specification for Linux systems:
- Config files (node_id, keyring) in `$XDG_CONFIG_HOME/exo` (default: `~/.config/exo`)
- Data files (models, state, logs) in `$XDG_DATA_HOME/exo` (default: `~/.local/share/exo`)
- Cache files (ipc) in `$XDG_CACHE_HOME/exo` (default: `~/.cache/exo`)

Backward compatibility maintained:
- `EXO_HOME` env var still works for legacy setups
- macOS/Windows continue using `~/.exo` directory

Fixes #905

## Test plan
- [x] Added comprehensive tests for XDG path resolution
- [x] Tested Linux XDG paths with custom env vars
- [x] Tested Linux default XDG paths
- [x] Tested legacy EXO_HOME compatibility
- [x] Tested macOS traditional paths
- [x] All existing tests pass

PR #988, opened 2025-12-22
feat: conform to XDG Base Directory Specification on Linux

Evanev7:
This is an extension of #964 with some cleanup.

### DIFF
--- a/src/exo/shared/constants.py
+++ b/src/exo/shared/constants.py
@@ -2,15 +2,14 @@ import os
 import sys
 from pathlib import Path
 
-EXO_HOME_ENV = os.environ.get("EXO_HOME", None)
+_EXO_HOME_ENV = os.environ.get("EXO_HOME", None)
 
 
 def _get_xdg_dir(env_var: str, fallback: str) -> Path:
-    """Get XDG directory with fallback. On non-Linux platforms, use ~/.exo."""
+    """Get XDG directory, prioritising EXO_HOME environment variable if its set. On non-Linux platforms, default to ~/.exo."""
 
-    # Setting EXO_HOME takes precedence over setting XDG_HOME
-    if EXO_HOME_ENV is not None:
-        return Path.home() / EXO_HOME_ENV
+    if _EXO_HOME_ENV is not None:
+        return Path.home() / _EXO_HOME_ENV
 
     if sys.platform != "linux":
         return Path.home() / ".exo"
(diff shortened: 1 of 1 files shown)


src/exo/shared/constants.py
```diff
@@ -1,35 +1,45 @@
...
-EXO_MASTER_STATE = EXO_HOME / "master_state.json"
-EXO_WORKER_STATE = EXO_HOME / "worker_state.json"
-EXO_MASTER_LOG = EXO_HOME / "master.log"
-EXO_WORKER_LOG = EXO_HOME / "worker.log"
-EXO_LOG = EXO_HOME / "exo.log"
-EXO_TEST_LOG = EXO_HOME / "exo_test.log"
+def _get_xdg_dir(env_var: str, fallback: str) -> Path:
+    """Get XDG directory with fallback. On non-Linux platforms, use ~/.exo."""
+
+    # Setting EXO_HOME takes precedence over setting XDG_HOME
```
JakeHillion: nit: the first comment is a bit wrong, and the second should be collapsed into it

rltakashige (approved):
Looks well tested and correct on my machine. LGTM

src/exo/shared/constants.py
```diff
@@ -1,35 +1,45 @@
...
-EXO_NODE_ID_KEYPAIR = EXO_HOME / "node_id.keypair"
 
-EXO_WORKER_KEYRING_FILE = EXO_HOME / "worker_keyring"
-EXO_MASTER_KEYRING_FILE = EXO_HOME / "master_keyring"
+EXO_CONFIG_HOME = _get_xdg_dir("XDG_CONFIG_HOME", ".config")
+EXO_DATA_HOME = _get_xdg_dir("XDG_DATA_HOME", ".local/share")
+EXO_CACHE_HOME = _get_xdg_dir("XDG_CACHE_HOME", ".cache")
 
-EXO_IPC_DIR = EXO_HOME / "ipc"
+# Models directory (data)
```
JakeHillion, line 27: Does this accurately represent our model data at the minute? I'm quite torn on whether it's cache or data. I think more of cache as things people might remove harmlessly assuming they'll be refilled, and data as things that would be bad to remove. Perhaps I'm wrong though, and removing model data would be super painful when you then have to redownload it.

src/exo/shared/constants.py
```diff
@@ -1,35 +1,45 @@
...
+EXO_MODELS_DIR_ENV = os.environ.get("EXO_MODELS_DIR")
+EXO_MODELS_DIR = (
+    Path(EXO_MODELS_DIR_ENV) if EXO_MODELS_DIR_ENV else EXO_DATA_HOME / "models"
+)
+
+# Log files (data/logs or cache)
+EXO_LOG = EXO_CACHE_HOME / "exo.log"
+EXO_TEST_LOG = EXO_CACHE_HOME / "exo_test.log"
```
JakeHillion, line 37: This seems odd to me, I wouldn't expect to find logs for an application in ~/.cache/exo/exo.log. Though I'd expect them to be on stderr really and captured with journald.

src/exo/shared/constants.py
Evanev7: The classic XDG spec concern. iirc chromium puts all it's cache in DATA just for simplicity, and the line is blurry to me anyway. I think DATA makes sense aside from the fact that I never think to look in .local/share for my app data.

src/exo/shared/constants.py
Evanev7: agreed, now that you point it out. .config/exo/exo.log is not better though.

PR #989, opened 2025-12-22
exo: enable multiprocessing support in PyInstaller bundles

JakeHillion:
Model loading fails silently when running from the DMG-packaged app,
despite working correctly with `uv run exo`. The bundled app spawns
child processes for model inference via multiprocessing, but these
processes fail to start in a frozen (PyInstaller) environment.

Add `freeze_support()` which is required for multiprocessing to work
in frozen applications.

Test plan:

Hardware setup: 3x Mac Studio M3 Ultra connected all-to-all with TB5

- Built a DMG using a modified .github/workflows/build-app.yml[0] to avoid
  publishing it.
- Installed on all 3 Macs, replacing the existing Exo.
- Downloaded Llama 3.3 70B (FP16).
- Downloaded Qwen3 Coder 235B A22B (8-bit).

Things that work now but didn't on the previous app:
- Topology looks good, previously there was no discovery.

What didn't work:
- Started an instance with Pipeline + MLX Ring + 3 Nodes. Failed.
- Started an instance with Tensor + MLX RDMA + 2 Nodes. Failed.

Will continue debugging the instance starting issues separately.

[0] https://github.com/exo-explore/exo/actions/runs/20461320368

### DIFF
--- a/src/exo/__main__.py
+++ b/src/exo/__main__.py
@@ -1,4 +1,7 @@
+from multiprocessing import freeze_support
+
 from exo.main import main
 
 if __name__ == "__main__":
+    freeze_support()
     main()


Evanev7:
Can you test without the -c flag? I do not believe it's necessary - freeze_suppport() should be sufficient.

JakeHillion:
Same tests pass with the small version (https://github.com/exo-explore/exo/actions/runs/20461320368)! Will squash and merge.

Evanev7 (approved):

Evanev7:
We used to explicitly call python -c to start up the runner, but we switched to normal multiprocessing a while back, so I think it was just vestigal.

PR #990, opened 2025-12-23
build-app: add branch trigger from named branch

JakeHillion:
As I've been working on the .dmg, it's become clear we need a way to test changes to the app. It's too hard to reproduce the full DMG locally to be reasonable and much more convenient to test if it's signed.

Add a feature to the build-app workflow where if you push specifically to the `test-app` branch it'll perform a build. The version is stubbed to `0.0.0-alpha.0`, which is about as low as it gets in semver so you'll always update away from it automatically with Sparkle. The resulting DMG won't be pushed to S3 but will be uploaded as a GitHub Actions artifact.

I've been using similar commits to this for a while for testing. It's worked well and not interfered with auto updating at all.

Test plan:
- Pushed this change to `test-app`.
- Generated action at https://github.com/exo-explore/exo/actions/runs/20447213358/job/58752909332
- Installed the DMG on a Mac. It worked as intended.

### DIFF
--- a/.github/workflows/build-app.yml
+++ b/.github/workflows/build-app.yml
@@ -4,6 +4,8 @@ on:
   push:
     tags:
       - "v*"
+    branches:
+      - "test-app"
 
 jobs:
   build-macos-app:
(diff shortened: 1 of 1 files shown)


Evanev7:
Just confirming, is this for producing the DMG so it can be manually tested or just testing that the app still builds in the first place?

JakeHillion:
> Just confirming, is this for producing the DMG so it can be manually tested or just testing that the app still builds in the first place?

Both. Creating the tag is way too noisy and annoying, this lets us test things in a slightly more subtle way. Specifically I intend to use this for making the changes to #989 and testing it.

Note the red signal is flakiness, this did work once on an identical commit (in the message).

Evanev7:
timestamp not available. wild

PR #991, opened 2025-12-23
Fix KV prefix cache for prompt reuse

Drifter4242:
## Motivation

KVCache wasn't working on MLX when using Kimi K2 causing slowness.
This fixes the KVCache and uses it to store the last prompt generation (only).
There are various bug fixes and some logging.

## Changes

- Wire up KVPrefixCache to runner and generate
- Fix exact match to return deepcopy (was returning reference)
- Fix trim_prompt_cache argument (was using wrong calculation)
- Fix token slicing to use best_snapshot_length (not index)
- Add _cache_length() using .offset for compatibility with older mlx_lm
- Fix prefill() to use max_tokens=1 with trim (workaround for mlx_lm bug when max_tokens is 0)
- Add clear() method for single-cache behavior
- Remove KEEP_KV_SIZE limit from prefix matching
- Add minimal logging for cache hits/misses

## Why It Works

KVCache stores the last prompt generation so it doesn't need to be recalculated.

### Manual Testing
I have two Macbook Studio M3 Ultra, each with 512Gb ram, connected with Thunderbolt 5. I ran Kimi K2 Thinking with MLX Ring and Tensor Split.
The TTFT is much faster. About 5-10 seconds rather than minutes. The logging shows it working.

### Automated Testing
No changes to automated testing
This is my first contribution, so I'm unfamiliar with how to proceed.

### DIFF
--- a/dashboard/src/lib/components/ChatForm.svelte
+++ b/dashboard/src/lib/components/ChatForm.svelte
@@ -89,7 +89,10 @@
 
   const isImageModel = $derived(() => {
     if (!currentModel) return false;
-    return modelSupportsTextToImage(currentModel);
+    return (
+      modelSupportsTextToImage(currentModel) ||
+      modelSupportsImageEditing(currentModel)
+    );
   });
 
   const isEditOnlyWithoutImage = $derived(
@@ -630,6 +633,23 @@
             <span class="sm:hidden">...</span>
           </span>
         {:else if shouldShowEditMode}
+          <span class="inline-flex items-center gap-1.5">
+            <svg
+              class="w-3.5 h-3.5"
+              fill="none"
+              viewBox="0 0 24 24"
+              stroke="currentColor"
+              stroke-width="2"
+            >
+              <path
+                stroke-linecap="round"
+                stroke-linejoin="round"
+                d="M11 5H6a2 2 0 00-2 2v11a2 2 0 002 2h11a2 2 0 002-2v-5m-1.414-9.414a2 2 0 112.828 2.828L11.828 15H9v-2.828l8.586-8.586z"
+              />
+            </svg>
+            <span>EDIT</span>
+          </span>
+        {:else if isEditOnlyWithoutImage}
           <span class="inline-flex items-center gap-1.5">
             <svg
               class="w-3.5 h-3.5"
--- a/dashboard/src/lib/components/ImageParamsPanel.svelte
+++ b/dashboard/src/lib/components/ImageParamsPanel.svelte
@@ -164,90 +164,92 @@
 <div class="border-b border-exo-medium-gray/30 px-3 py-2">
   <!-- Basic params row -->
   <div class="flex items-center gap-3 flex-wrap">
-    <!-- Size -->
-    <div class="flex items-center gap-1.5">
-      <span class="text-xs text-exo-light-gray uppercase tracking-wider"
-        >SIZE:</span
-      >
-      <div class="relative">
-        <button
-          bind:this={sizeButtonRef}
-          type="button"
-          onclick={() => (isSizeDropdownOpen = !isSizeDropdownOpen)}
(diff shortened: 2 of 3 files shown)


rltakashige:
Thank you for this; I've wanted to finish off this piece of code for a while but haven't got around to it. 

> This is my first contribution, so I'm unfamiliar with how to proceed.

Automated testing for this PR should go in `src/exo/worker/tests/unittests/test_mlx`. I want to write some of these tests myself in the near future, but - if you want to get started now - you can take a look at `src/exo/worker/tests/unittests/test_plan` as well as MLX LM's tests for an example of what to do.

Drifter4242:
In the process of fixing the CI issues, I found some problems with the KVCache code and fixed those too. Now it is consistently taking about 4 seconds TTFT even with a long conversation. Note that Kimi K2 has a think section which is automatically removed after the last prompt - this means the KVCache needs a small update (rather than having no work).

src/exo/worker/engines/mlx/cache.py
```diff
@@ -42,39 +63,76 @@ def get_kv_cache(
...
+            cached_length = _cache_length(self.caches[best_snapshot_index])
+            tokens_to_trim = cached_length - best_snapshot_length
+            if tokens_to_trim > 0:
+                trim_prompt_cache(prompt_cache, tokens_to_trim)
 
-        prefill(model, tokenizer, sampler, tokenized_prompt, prompt_cache)
+            # Prefill the remaining tokens (except the last one which stream_generate needs)
+            remaining_tokens = tokenized_prompt[best_snapshot_length:]
+            if len(remaining_tokens) > 1:
+                prefill(model, tokenizer, sampler, remaining_tokens[:-1], prompt_cache)
```
rltakashige: I know that I used to have the prefill function here, but I'm not sure it's good to have part of the generation logic in this file, now that you're doing this properly. I'd hope to just return the KV Cache and the remaining tokens that need to be prefilled.

src/exo/worker/engines/mlx/cache.py
```diff
@@ -5,34 +5,55 @@
...
 from mlx_lm.tokenizer_utils import TokenizerWrapper
 
 from exo.worker.engines.mlx import Model
-from exo.worker.engines.mlx.constants import KEEP_KV_SIZE, KV_BITS, KV_GROUP_SIZE
+from exo.worker.engines.mlx.constants import KV_BITS, KV_GROUP_SIZE
 from exo.worker.engines.mlx.utils_mlx import make_kv_cache
+from exo.worker.runner.bootstrap import logger
+
+# Type alias matching make_kv_cache return type
+KVCacheType = list[KVCache | RotatingKVCache | QuantizedKVCache]
```
rltakashige: This is quite nice, but this would be better as a shared type in Exo. Especially since this is not exhaustive and will need to be updated in several places.

rltakashige (commented):
Looks pretty good so far. Looking forward to what you come up with for the CI!

src/exo/worker/engines/mlx/cache.py
Drifter4242: Agreed.  Moved to mlx.py in shared.

src/exo/worker/engines/mlx/cache.py
Drifter4242: Agreed. Moved to generate.py.

Drifter4242:
I've also made some fixes to the CI (mainly regenerating cache and tokenizer pyi files) which means I removed some type ignores and the CI runs clean on my branch.
I haven't written any automated tests - is this needed for this PR?

Drifter4242:
I thought I'd ping this thread to see if there are any further changes or actions that are needed. Thanks.

rltakashige:
Really sorry about this. I was trying to rebase because the history was so out of date, and I pushed something with 0 commits, closing the PR. I've reopened your PR on https://github.com/exo-explore/exo/pull/1262 where I'll try fixing the history again...

testqer1234-hash (commented):
LGTM

PR #996, opened 2025-12-23
feat: add memory pressure awareness to placement decisions

suyashbhawsar:
## Summary

- Add memory pressure detection for macOS (sysctl/memory_pressure) and Linux (PSI metrics)
- Introduce `effective_available` property that returns conservative memory estimates under pressure
- Update placement logic to exclude nodes under critical pressure and reduce allocation on warned nodes

## Problem

The current placement system uses raw "memory available" from psutil, which doesn't reflect actual memory pressure. On systems with unified memory (Apple Silicon), the GPU shares RAM with the CPU. When the system is under memory pressure (compression active, swap thrashing), available memory reported by psutil is misleading and can cause OOM conditions.

## Solution

- **macOS**: Read `sysctl vm.memory_pressure` (kernel-level 0/1/2/4) and parse `memory_pressure` command output
- **Linux**: Parse `/proc/pressure/memory` PSI metrics with Facebook-derived thresholds
- **Pressure levels**: NORMAL (full capacity), WARN (50% conservative), CRITICAL (excluded from placement)
- **Performance**: Pressure checks run every 5 seconds to minimize subprocess overhead

## Test plan

- [x] 26 unit tests passing (13 new for memory pressure, 4 new for placement)
- [x] Verified on macOS - pressure detection working correctly
- [x] Backwards compatible - defaults to NORMAL when detection fails
- [x] Linting passes (ruff check/format)

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -29,7 +29,7 @@ dependencies = [
     "exo_pyo3_bindings", # rust bindings
     "anyio==4.11.0",
     "bidict>=0.23.1",
-    "mlx>=0.29.3",
+    "mlx>=0.30.1",
     "mlx-lm>=0.28.3",
     "tiktoken>=0.12.0", # required for kimi k2 tokenizer
     "hypercorn>=0.18.0",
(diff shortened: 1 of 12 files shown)


AlexCheema:

PR #1000, opened 2025-12-23
placement: generate per-node host lists for MLX ring backend

JakeHillion:
Pipeline + MLX Ring worked with 2 nodes but failed to initialize with
3 or more nodes. The MLX ring backend requires each node to know its
specific left and right neighbors in the ring, but the previous
implementation provided a single flat host list shared by all nodes.

With 2 nodes, a flat list [host0, host1] accidentally worked because
each node could find its only neighbor. With 3+ nodes, each node needs
a customized view:
- Rank 0: [self, right_neighbor, placeholder]
- Rank 1: [left_neighbor, self, right_neighbor]
- Rank 2: [placeholder, left_neighbor, self]

Changed MlxRingInstance from `hosts: list[Host]` to
`hosts_by_node: dict[NodeId, list[Host]]` with `ephemeral_port: int`.

Added `get_mlx_ring_hosts_by_node()` which generates per-node host
lists where:
- Self position uses 0.0.0.0 for local binding
- Left/right neighbors use actual connection IPs
- Non-neighbors use 198.51.100.1 (RFC 5737 TEST-NET-2 placeholder)

Also added IP prioritization (en0 > en1 > non-Thunderbolt > any) to
prefer stable network interfaces.

Fixed topology discovery recording loopback addresses (127.0.0.1) as
valid connections to remote nodes. The reachability check now verifies
node identity via HTTP GET /node_id rather than just checking if the
port is open.

Test plan:

- Built a DMG [0]
- Installed on all Macs and started cluster.
- Requested a 3 node Pipeline + MLX Ring Llama 3.3 70B (FP16).
- It started and I was able to send a few chat messages.

Eventually my instance seemed to get into a broken state and chat
stopped working, but this commit is a clear step forward.

[0] https://github.com/exo-explore/exo/actions/runs/20473983471/job/58834969418

### DIFF
--- a/src/exo/master/placement_utils.py
+++ b/src/exo/master/placement_utils.py
@@ -215,9 +215,11 @@ def get_mlx_ibv_devices_matrix(
                 continue
 
             # Find the IP J uses to talk to I
-            for connection_ip in _find_connection_ip(node_j, node_i, cycle_digraph):
+            for connection_ip, _ in _find_connection_ip(node_j, node_i, cycle_digraph):
                 # This is a local IP on I, which is attached to an interface: find that interface
-                if interface_name := _find_interface_name_for_ip(connection_ip, node_i):
+                if interface_name := _find_rdma_interface_name_for_ip(
+                    connection_ip, node_i
+                ):
                     matrix[i][j] = interface_name
                     logger.info(
                         f"Interface name for {connection_ip} on {node_i.node_id}: {interface_name}"
@@ -238,17 +240,17 @@ def _find_connection_ip(
     node_i: NodeInfo,
     node_j: NodeInfo,
     cycle_digraph: Topology,
-) -> Generator[str]:
-    """Find all IP addresses that connect node i to node j."""
+) -> Generator[tuple[str, bool]]:
+    """Find all IP addresses that connect node i to node j, with thunderbolt flag."""
     for connection in cycle_digraph.list_connections():
         if (
             connection.local_node_id == node_i.node_id
             and connection.send_back_node_id == node_j.node_id
         ):
-            yield connection.send_back_multiaddr.ip_address
+            yield connection.send_back_multiaddr.ip_address, connection.is_thunderbolt()
 
 
-def _find_interface_name_for_ip(
+def _find_rdma_interface_name_for_ip(
     ip_address: str,
     node_info: NodeInfo,
 ) -> str | None:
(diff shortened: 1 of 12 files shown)


JakeHillion:
Note for reviewers: this is a stacked PR. Only the top commit needs reviewing, the first is contained in #997. Please do not merge this PR without first merging #997.

src/exo/master/placement_utils.py
```diff
@@ -238,14 +238,14 @@ def _find_connection_ip(
...
     for connection in cycle_digraph.list_connections():
         if (
             connection.local_node_id == node_i.node_id
             and connection.send_back_node_id == node_j.node_id
         ):
-            yield connection.send_back_multiaddr.ip_address
+            yield connection.send_back_multiaddr.ip_address, connection.is_thunderbolt()
 
 
 def _find_interface_name_for_ip(
```
AlexCheema: wasn't this renamed `_find_rdma_interface_name_for_ip`?
maybe we should refactor this and `_find_general_interface_name_for_ip` combining them into one function and then prepending `rdma_`.

AlexCheema: Maybe a TODO for a future PR, but this condition is actually wrong:
`if interface.name not in ["en2", "en3", "en4", "en5", "en6", "en7"]:`

On MacBook, `en1` is also TB5.
We should probably have a better way of determining if an interface is TB5.

src/exo/master/placement_utils.py
```diff
@@ -269,6 +269,122 @@ def _find_interface_name_for_ip(
...
+    ]
+
+    en0_ip = next(
+        (
+            ip
+            for (ip, _), interface_name in zip(ips, interface_names, strict=True)
+            if interface_name == "en0"
+        ),
+        None,
+    )
```
AlexCheema: These are pretty verbose. Maybe extract into a function to select an interface by name.

src/exo/worker/utils/net_profile.py
```diff
@@ -1,41 +1,78 @@
...
 
 from anyio import create_task_group, to_thread
+from loguru import logger
 
 from exo.shared.topology import Topology
 from exo.shared.types.common import NodeId
 
 
-# TODO: ref. api port
 async def check_reachability(
```
AlexCheema, line 10: I know @Evanev7 had a different solution for this.
Here we're hitting the `/node_id` endpoint on the HTTP API on each node.
Evan's solution was lower level, I think checking if it's possible to open a TCP socket.
However, the tradeoff is that Evan's solution ignores loopback addresses which makes it so this ping wouldn't work for a second node that's running on the same machine. This is an annoying constraint if we can't run two exo's on on machine (super helpful for testing). However, maybe discovery would still work fine. It's a question for @Evanev7

src/exo/worker/utils/net_profile.py
```diff
@@ -1,41 +1,78 @@
...
+    if remote_node_id == self_node_id:
+        # Connected to ourselves via loopback - skip
+        return
+
+    if remote_node_id != expected_node_id:
+        logger.warning(
+            f"Discovered node with unexpected node_id; "
+            f"ip={target_ip}, expected_node_id={expected_node_id}, "
+            f"remote_node_id={remote_node_id}"
+        )
```
AlexCheema, line 50: Related to above, a lot of these warnings would currently come up if running exo with two nodes on the same machine

src/exo/worker/utils/net_profile.py
```diff
@@ -1,41 +1,78 @@
...
-    sock.settimeout(1)  # 1 second timeout
-    try:
-        result = await to_thread.run_sync(sock.connect_ex, (target_ip, 52415))
-    except socket.gaierror:
-        # seems to throw on ipv6 loopback. oh well
-        # logger.warning(f"invalid {target_ip=}")
+    """Check if a node is reachable at the given IP and verify its identity."""
+
+    def _fetch_remote_node_id() -> str | None:
+        connection = http.client.HTTPConnection(target_ip, 52415, timeout=1)
```
AlexCheema, line 19: Related to above, hardcoding 52415 here could be problematic since technically we give the ability to override the api port.

src/exo/worker/main.py
```diff
@@ -414,9 +414,14 @@ async def _poll_connection_updates(self):
...
             edges = set(self.state.topology.list_connections())
-            conns = await check_reachable(self.state.topology)
+            conns = await check_reachable(self.state.topology, self.node_id)
             for nid in conns:
                 for ip in conns[nid]:
+                    if "127.0.0.1" in ip or "localhost" in ip:
+                        logger.warning(
+                            f"Loopback connection should not happen: {ip=} for {nid=}"
+                        )
+
```
AlexCheema, line 424: This warning would appear if running two exo nodes on the same machine.

src/exo/master/placement_utils.py
```diff
@@ -238,14 +238,14 @@ def _find_connection_ip(
...
-    """Find all IP addresses that connect node i to node j."""
+) -> Generator[tuple[str, bool]]:
+    """Find all IP addresses that connect node i to node j, with thunderbolt flag."""
     for connection in cycle_digraph.list_connections():
         if (
             connection.local_node_id == node_i.node_id
             and connection.send_back_node_id == node_j.node_id
         ):
-            yield connection.send_back_multiaddr.ip_address
+            yield connection.send_back_multiaddr.ip_address, connection.is_thunderbolt()
```
AlexCheema, line 250: This was a little hacky to return `is_thunderbolt` here. Might consider refactoring (could return the connection objects instead for example.

AlexCheema (changes requested):

src/exo/master/placement_utils.py
Evanev7: Should be fixed in #928

src/exo/worker/utils/net_profile.py
Evanev7: Normal discovery should still work, the TCP socket opener should just add additional sources that aren't loopback. Also, I never merged the "ignore loopback pr"

rltakashige:
Looks like Alex's review is p comprehensive. Only other thing I'd want to suggest is a small test on tensor ring (with llama 1b on 3 nodes)

Evanev7:
I'll probably pick this up in a day or two.

src/exo/worker/utils/net_profile.py
Evanev7: Ok this is worse than I thought in some ways, better in others. I like pulling the node_id endpoint to verify rather than just checking an open socket, but tying it to the api port is a nightmare. We should run a different server on a different port for this style of discovery.

Evanev7:
I've addressed some of the earlier comments but this http server is non-tractable for single device testing. The good news is, it shouldn't matter and mdns should discover exo locally without pinging itself. That shouldn't is doing a lot of heavy lifting - the topology i get back from state is a bit weird - but we need a larger changeup of this code to get this style of http discovery functional.

Evanev7:
I've updated this branch in accordance with the review comments, but the ping-style connectivity is still a hack that needs a better solution imo. We should not depend on it for single node testing, and we shouldn't depend on it at all if possible. I could revert the changes here @AlexCheema but it's a less obvious fix than you're hoping.

Evanev7 (approved):

PR #1006, opened 2025-12-24
Separate out the Runner's behaviour into a "connect" phase and a "load" phase

Evanev7:
## Motivation

We should ensure all runners are connected before loading the model - this gives us finer grained control in the future for the workers planning mechanism over the runners state.

## Changes

- Introduced task ConnectToGroup, preceeding LoadModel
- Introduced runner statuses Idle, Connecting, Connected
- Separated out initialize_mlx from shard_and_load
- Single instances never go through the connecting phase

## Test Plan

# Automated Testing
Added a test for checking event ordering in a standard workflow.

# Manual testing
Tested Llama 3.2 1b and Kimi K2 Thinking loads and shuts down repeatedly on multiple configurations.
Not exhaustive, however.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -29,7 +29,7 @@ dependencies = [
     "exo_pyo3_bindings", # rust bindings
     "anyio==4.11.0",
     "bidict>=0.23.1",
-    "mlx>=0.30.1; sys_platform != 'darwin'",
+    "mlx>=0.30.1; sys_platform == 'darwin'",
     "mlx[cpu]>=0.30.1; sys_platform == 'linux'",
     "mlx-lm>=0.28.3",
     "tiktoken>=0.12.0", # required for kimi k2 tokenizer


src/exo/shared/types/tasks.py
```diff
@@ -57,5 +61,11 @@ class Shutdown(BaseTask):  # emitted by Worker
...
 
 Task = (
-    CreateRunner | DownloadModel | LoadModel | StartWarmup | ChatCompletion | Shutdown
+    CreateRunner
+    | DownloadModel
+    | LoadModel
+    | StartWarmup
+    | ChatCompletion
+    | Shutdown
+    | ConnectToGroup
```
rltakashige: Order for clarity

src/exo/worker/plan.py
```diff
@@ -183,17 +231,17 @@ def _ready_to_warmup(
         assert device_rank < world_size
         assert device_rank >= 0
 
-        # Rank != n-1
-        accepting_ranks_ready = device_rank != world_size - 1 and all(
+        # Rank != 0
```
rltakashige, line 239: Add a TODO to check whether this ordering is necessary (and whether we should do the same for chat completions.)

src/exo/worker/plan.py
```diff
@@ -117,50 +127,88 @@ def _model_needs_download(
...
+
+def _load_model(
+    runners: Mapping[RunnerId, RunnerSupervisor],
+    all_runners: Mapping[RunnerId, RunnerStatus],
+) -> LoadModel | None:
+    for runner in runners.values():
+        instance = runner.bound_instance.instance
+        shard_assignments = instance.shard_assignments
+
+        is_runner_waiting = isinstance(runner.status, RunnerConnected)
```
rltakashige, line 205: Single instance will never get loaded, maybe check `isinstance(runner.status, (RunnerConnected, RunnerIdle))`.

More fundamentally, however, I really do not like the fact that we have two statuses from which a task can happen.

src/exo/worker/engines/mlx/utils_mlx.py
```diff
@@ -5,6 +5,7 @@
 from pathlib import Path
 from typing import Any, Callable, cast
 
+from mlx.core.distributed import Group
```
rltakashige: Remember to resolve these type check issues - mx.distributed.Group isn't so bad to have anyway.

src/exo/worker/tests/unittests/test_runner/test_event_ordering.py
```diff
@@ -0,0 +1,209 @@
...
+        event_sender.join = nothin
+        task_receiver.close = nothin
+        task_receiver.join = nothin
+
+        mlx_runner.main(bound_instance, event_sender, task_receiver)
+
+        return event_receiver.collect()
+
+
+def test_events_processed_in_correct_order(patch_out_mlx: pytest.MonkeyPatch):
```
rltakashige, line 150: Test fails

Evanev7: aw what it worked earlier

src/exo/worker/plan.py
Evanev7: good catch - but also, I don't like the fact that we put the runner through a step it doesn't need!

joanwhitegt-art:
thanks for the follow up understood

virgilnsb817-collab:
appreciate the update on this

fsdfcdsafcsa-maker:
got it thanks for the confirmation

jessiaorshane-collab:
thanks for the explanation here

pyproject.toml
```diff
@@ -29,7 +29,8 @@ dependencies = [
     "exo_pyo3_bindings", # rust bindings
     "anyio==4.11.0",
     "bidict>=0.23.1",
-    "mlx>=0.30.1",
+    "mlx>=0.30.1; sys_platform != 'linux'",
```
rltakashige: `== 'darwin'`

rltakashige:
Also tested on one and two nodes.

PR #1018, opened 2025-12-26
Add Linux note to Run from Source section in README

MatiwosKebede:


Evanev7:
Could be worth adding, this is not a good update.

PR #1028, opened 2025-12-28
feat: remember last launch settings (model, sharding, instance type)

Drifter4242:
## Motivation

Saves the last launch settings, so that the next time you run exo it will default to the same launch settings.
This is just a small quality of life improvement.

## Changes

When you launch it saves the settings to the web browser local storage. When it fills out the model list, it reads the settings and sets the default.

I reviewed, tested and edited the code, but some of the code was written by Claude Opus. I hope that's ok.

## Why It Works

See above

### Manual Testing

I have two Macbook Studio M3 Ultras, each with 512Gb ram, connected with Thunderbolt 5. I ran Kimi K2 Thinking with MLX Ring and Tensor Split.
I ran exo multiple times to confirm that the default works.

### Automated Testing

No changes to automated testing.

### DIFF
--- a/src/exo/shared/types/worker/runners.py
+++ b/src/exo/shared/types/worker/runners.py
@@ -53,6 +53,10 @@ class RunnerRunning(BaseRunnerStatus):
     pass
 
 
+class RunnerShuttingDown(BaseRunnerStatus):
+    pass
+
+
 class RunnerShutdown(BaseRunnerStatus):
     pass
 
@@ -70,6 +74,7 @@ RunnerStatus = (
     | RunnerWarmingUp
     | RunnerReady
     | RunnerRunning
+    | RunnerShuttingDown
     | RunnerShutdown
     | RunnerFailed
 )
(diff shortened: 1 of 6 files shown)


Evanev7:
I agree with the feature, not the implementation. Will leave this open until we have an alternative, but I don't intend to merge this.

Drifter4242:
> I agree with the feature, not the implementation. Will leave this open until we have an alternative, but I don't intend to merge this.

Thanks for looking at my code.

Do you have any more information about what's wrong with the implementation or how I can fix it?

Evanev7:
On second review, I misunderstood the feature and actually this is completely fine. @AlexCheema can you take a look and see if it matches the svelte styling? I'm not familiar enough with that side of the codebase atm.

samiamjidkhan:
LGTM! @AlexCheema

AlexCheema:
EDIT: Worked for me now (forgot to rebuild dashboard, my bad!)

AlexCheema:
Also can we persist minimum nodes too? @Drifter4242

AlexCheema:
I noticed that you're persisting the settings only after you launch an instance.
I think it makes more sense to actually just persist the settings every time the settings are changed.

Drifter4242:
I've updated the branch as suggested.

AlexCheema:
Code looks good and works after testing. Thank you for the contribution!

PR #1044, opened 2025-12-30
Emit download progress on start, and change downloads to be keyed by model_id

AlexCheema:
## Motivation

We added a download page to the dashboard which shows the currently download status of each model on each node. Users have reported this to be extremely useful.

However, we don't currently fetch the download progress on start, so it doesn't show any model's download status.

## Changes

Fetch and emit model download status on start of worker, and periodically every 5 mins.
Also to support this, I changed download_status to be keyed by model_id instead of shard, since we want download_status of each model, not each shard.

## Why It Works

The dashboard already implements the correct functionality, we just weren't populating the download status in the state. Now it gets populated and shows correctly.

### Manual Testing
On a cluster of 2 x 512GB M3 Ultra Mac Studio, I launched an instance onto one node that hadn't been downloaded. I checked the download page and it showed the in progress download. I downloaded it to completion, restarted exo on both nodes, and then opened the download page and it showed the model as 100% downloaded and other models as 0% that hadn't been downloaded.

### DIFF
--- a/src/exo/worker/main.py
+++ b/src/exo/worker/main.py
@@ -448,8 +448,8 @@ class Worker:
             await anyio.sleep(10)
 
     async def _emit_existing_download_progress(self) -> None:
-        while True:
-            try:
+        try:
+            while True:
                 logger.info("Fetching and emitting existing download progress...")
                 async for (
                     _,
@@ -480,7 +480,6 @@ class Worker:
                         NodeDownloadProgress(download_progress=status)
                     )
                 logger.info("Done emitting existing download progress.")
-            except Exception as e:
-                logger.error(f"Error emitting existing download progress: {e}")
-            finally:
                 await anyio.sleep(5 * 60)  # 5 minutes
+        except Exception as e:
+            logger.error(f"Error emitting existing download progress: {e}")


src/exo/worker/main.py
```diff
@@ -444,3 +446,38 @@ async def _poll_connection_updates(self):
...
+                            )
+                    else:
+                        continue
+
+                    self.download_status[progress.shard.model_meta.model_id] = status
+                    await self.event_sender.send(
+                        NodeDownloadProgress(download_progress=status)
+                    )
+                logger.info("Done emitting existing download progress.")
+            except Exception as e:
```
Evanev7: Let's not catch-all errors please.

Evanev7 (dismissed):
not a fan but include wildcard downloads in here and I'll okay it. it's a regression in functionality but i think we've decided it's worth it for now.

AlexCheema:
> not a fan but include wildcard downloads in here and I'll okay it. it's a regression in functionality but i think we've decided it's worth it for now.

Did the wildcard change in #1055.

src/exo/worker/main.py
AlexCheema: What should we catch?

Evanev7:
I'd really rather you combined them, these changes should go together else this change doesn't make sense. I'll merge 1055 in here and run the formatter.

src/exo/worker/main.py
Evanev7: Realising that we raise plain `Exception`s in download_utils feels like a slight slap to the face, so fair enough. This loop can exit on exception, it's non-critical and catching all errors is a surefire way to GPU lock a device.

Evanev7 (approved):

PR #1045, opened 2025-12-30
log EXO_LIBP2P_NAMESPACE on start

AlexCheema:
## Motivation

Often when users send their logs, it's helpful to know what version they are running on and if they're running with a namespace.

## Changes

Logs `EXO_LIBP2P_NAMESPACE` on start

## Why It Works

Logs in output.

### Manual Testing
It logs it when running exo.


Evanev7:
Format it!

PR #1061, opened 2025-12-30
Deduplicate tasks in plan_step

AlexCheema:
## Motivation

We were getting duplicate tasks causing exo to get stuck.
Unclear what the root cause is.

## Changes

In `plan_step` check if we've seen a task with this ID before and if so skip.

## Why It Works

We ignore duplicate tasks.

## Test Plan

It would sometimes happen that a duplicate task would be received by some workers in an instance but not all, causing the instance to get stuck. The only way to recover from that was to kill exo, which would often cause a GPU lock.

This doesn't happen anymore.


Evanev7:
I'm going to reimplement this to fix the underlying network race condition, so closing this for now.

PR #1069, opened 2025-12-31
fix: prevent form submission during IME composition

ricky-chaoju:
## Problem
When typing in Chinese (or other IME-based languages like Japanese/Korean), pressing Enter to select a character from the IME candidate list would incorrectly submit the message instead of confirming the character selection.

## Solution
Added IME composition state detection in the `handleKeydown` function in `ChatForm.svelte`:
- Check `event.isComposing` to detect active IME composition
- Fallback to `event.keyCode === 229` for broader browser compatibility
- Return early when IME is active, allowing normal character selection

## Changes
- Modified `dashboard/src/lib/components/ChatForm.svelte` 
- Added IME composition check before Enter key handling

### DIFF
--- a/dashboard/src/lib/components/ChatForm.svelte
+++ b/dashboard/src/lib/components/ChatForm.svelte
@@ -139,6 +139,11 @@
 	}
 
 	function handleKeydown(event: KeyboardEvent) {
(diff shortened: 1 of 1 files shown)


AlexCheema:
Looks good. Thank you!

ricky-chaoju:
Thanks for the review and merge!

PR #1071, opened 2025-12-31
Run Sparkle updater in background to make macOS app work offline.

AlexCheema:
## Motivation

Currently the macOS app requires internet to start, since the sparkle update runs on start in the foreground.

## Changes

Run sparkle updater in background.

## Why It Works

Sparkle updater no longer runs in the foreground. It now runs in the background so if it fails it will fail gracefully.

### Manual Testing
We need to make a new build of the app and ensure it starts without internet.


JakeHillion:
I tested both EXO-latest.dmg and the EXO-1.0.61-alpha.0.dmg and both start fine on a Mac with no Internet connection. Will leave that comment on the issue too, but currently I can't reproduce the issue this is solving.

AlexCheema:
This doesn't seem to fix anything.

Issue #1070, opened 2025-12-31
[BUG] Deleting instance then creating new instance launches stale model


AlexCheema:
## Describe the bug

Deleting an instance, creating a new one, then sends a chat completion request with the old, stale instance.

## To Reproduce

1. Create instance of Qwen3-30B
2. Delete instance
3. Create instance of Llama 3.2 3B
4. Chat (model selector shows Llama 3.2 3B selected)
5. It sends a request to wrong model, Qwen3-30B

<img width="1197" height="874" alt="Image" src="https://github.com/user-attachments/assets/e0ef856e-5286-428b-a13e-ef7099a3393b" />

## Expected behavior

It *should* automatically select the most recently launched model.
The model selector already shows the correct model.
But it should actually use that model when chatting.

## Actual behavior

It uses the wrong model, the previously launched model.

## Environment

- macOS Version: 26.2
- EXO Version: c1be5184b2787bb11bc16c72789d0afd7b6b5510
- Hardware:
Any 
- Interconnection:
Any

PR #1076, opened 2026-01-01
resolve issue #1070

samiamjidkhan:
## Motivation

https://github.com/exo-explore/exo/issues/1070

## Changes

Added check in ChatForm.svelte to reset selectedChatModel when it no longer matches any running instance.

## Why It Works

The $effect now detects when the selected model is stale (not in availableModels()) and resets to the first available model.

### Manual Testing

1. Create instance of Model A → Delete it → Create instance of Model B → Chat
2. Verify request goes to Model B (not Model A)


AlexCheema:
I think the better way to fix this is to auto-select an instance after launching it.

If we delete one that is selected, then we should select the most recently launched model.

Most importantly, whatever is selected in the dropdown menu should be the one that we chat to, otherwise what we display to the user is not in line with the actual model running.

AlexCheema:
Tested this change.

Launching an instance does *not* select it in the model dropdown.

How to reproduce:
- Launch an instance of Qwen 0.6B 4-bit
- It gets auto-selected (correct)
- Chat with Qwen 0.6B 4-bit
- It returns correctly using Qwen 0.6B 4-bit (correct)
- Launch an instance of Llama 1B 4-bit
- It *does not* select Llama 1B 4-bit (wrong)

AlexCheema:
- Launch an instance of Qwen 0.6B 4-bit
- It gets auto-selected (correct)
- Chat with Qwen 0.6B 4-bit
- It returns correctly using Qwen 0.6B 4-bit (correct)
- Delete instance of Qwen 0.6B 4-bit
- Instance gets deleted and model option disappears next to chat input
- Chat again
- Response is `Error: API error: 404 - {"detail":"No instance found for model mlx-community/Qwen3-0.6B-4bit"}` (WRONG)

The correct behavior should be when you start exo from scratch with no instances and try to chat - it returns `Error: No model selected and no running instances available. Please launch an instance first.`
So the solution is when there are no instances it should behave like when you open the dashboard initially and no instances have been launched.

samiamjidkhan:
Added an else branch to clear the selection when no instances are running @AlexCheema

Issue #1025, opened 2025-12-28
High GPU usage when EXO Dashboard is open in browser (increases with node count)


andrewwutw:
### Description
The EXO Dashboard consumes significant GPU resources when active in a browser tab. The GPU load appears to scale linearly with the number of nodes in the network. I have verified this behavior on both **Safari** and **Chrome** with similar results.

### Environment
- **EXO Version:** 1.0.60
- **Device:** 2 x Mac Studio (M-series chips)
- **OS:** macOS
- **Monitoring Tool:** `macmon`

### Reproduce Steps

1. **Start EXO on the first machine:**
   Launch the EXO app (v1.0.60) on the first Mac Studio without running any model instances.

<img width="317" height="482" alt="Image" src="https://github.com/user-attachments/assets/4ecd2af4-06aa-4228-ae47-733aec12705a" />

   *GPU usage (via macmon) is idle at approximately 2%.*

<img width="1082" height="843" alt="Image" src="https://github.com/user-attachments/assets/c4376a7b-305d-46ad-a631-509a8a763873" />

2. **Open the Dashboard:**
   Access the dashboard at `http://localhost:8000`. 
   Observed that the first node's GPU usage immediately jumps to about 10%.

<img width="321" height="238" alt="Image" src="https://github.com/user-attachments/assets/ffa819e3-c9e2-4eac-87cb-73af806cc353" />

   `macmon` reflects this significant increase in GPU load.

<img width="1082" height="843" alt="Image" src="https://github.com/user-attachments/assets/43b0d892-f78a-42ea-a78f-4767f9315fc6" />

3. **Add a second node to the network:**
   Join a second EXO node to the network while keeping the first node's dashboard open.
   The GPU usage on the first node now spikes to over 20%.

<img width="287" height="501" alt="Image" src="https://github.com/user-attachments/assets/b5a75165-bd03-4d82-a0a4-3da24959607d" />

   `macmon` confirms the elevated GPU usage.

<img width="1082" height="843" alt="Image" src="https://github.com/user-attachments/assets/5b9e2b2c-c9bb-4d86-95d3-b06e2e23fc03" />

4. **Minimize the browser window:**
   Once the dashboard tab/window is minimized, the GPU usage immediately drops back to the idle state (~2%).

<img width="1082" height="843" alt="Image" src="https://github.com/user-attachments/assets/df802237-14ca-4aaa-8836-be38a0765050" />

### Additional Observations
[...]

PR #1078, opened 2026-01-01
resolve issue #1025

samiamjidkhan:
## Motivation

https://github.com/exo-explore/exo/issues/1025

## Changes

- Add JSON comparison in AppStore before state updates - only triggers Svelte reactivity when data actually changes
- Add visibility-based polling - skips fetches when browser tab is hidden
- Reduce poll interval from 1s to 2s
- Add hash-based render check in TopologyGraph to prevent unnecessary D3 re-renders
- Add will-change CSS hints and prefers-reduced-motion support
- Pause all CSS animations when page is hidden

## Why It Works

The root cause was Svelte reactivity triggering full re-renders on every 1s poll, even when data hadn't changed. By comparing JSON before state assignment, Svelte's reactivity system is only triggered when something actually changed. The TopologyGraph hash check provides an additional guard against unnecessary D3 SVG redraws.

### Manual Testing
Hardware: MacBook Pro
- Tested with **fake** nodes only, some more testing with real nodes would be nice @AlexCheema :)

### Automated Testing
N/A

### DIFF
--- a/dashboard/src/app.css
+++ b/dashboard/src/app.css
@@ -198,14 +198,34 @@
 	stroke: oklch(0.85 0.18 85 / 0.4);
 	stroke-width: 1.5px;
 	stroke-dasharray: 8, 8;
-	animation: flowAnimation 1s linear infinite;
+	animation: flowAnimation 1.5s linear infinite;
 	filter: drop-shadow(0 0 3px oklch(0.85 0.18 85 / 0.5));
+	/* GPU optimization - hint to browser this element will animate */
+	will-change: stroke-dashoffset;
 }
 
 .graph-link-active {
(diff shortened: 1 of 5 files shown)


AlexCheema:
This wasn't what I was expecting.

The whole point of the svelte compiler / svelte runes is to be able to only re-render relevant parts of the UI after a state change.

All these manual checks to see if something changed seems like an anti-pattern. Is there a better way to do this?

samiamjidkhan:
> This wasn't what I was expecting.
> 
> The whole point of the svelte compiler / svelte runes is to be able to only re-render relevant parts of the UI after a state change.
> 
> All these manual checks to see if something changed seems like an anti-pattern. Is there a better way to do this?

Ah I see. I'll make a new PR

PR #1079, opened 2026-01-01
Feat per request temperature, top_p, logit bias, frequency_penalty

Drifter4242:
## Motivation

I wanted to specify the temperature from the api. I added temperature along with top_p, logit bias, and frequency_penalty

## Changes

Modified mlx_generate() to create samplers and logits processors per-request rather than once at initialization
Added make_sampler_from_task() to map openai temperature and top_p to mlx_lm's make_sampler()
Added make_logits_processors_from_task() to handle:
logit_bias: Converts string token IDs (OpenAI format) to int keys (mlx_lm format)
frequency_penalty: Maps to repetition_penalty using formula 1.0 + (frequency_penalty * 0.25), clamped to [0.1, 2.0] for safety

## Why It Works

temperature and top_p map directly from openai to mlx
logit_bias just required a conversion from string to int
frequency_penalty is more tricky. The mapping is approximate but generally openai's frequency penalty has values -2,0,+2 which map to repetition_penalty of 0.5, 1.0 and 1.5

### Manual Testing
I have two Macbook Studio M3 Ultras, each with 512Gb ram, connected with Thunderbolt 5. I ran Kimi K2 Thinking with MLX Ring and Tensor Split.
For temperature, setting it to 0 causes the result to be deterministic
For top_p, setting to 0.95 created a varied output
A frequency penalty of 2.0 and a prompt or "Say the word 'hello' 20 times" caused a result that failed to repeat hello.
For logit bias, I set " yes" to -100, and had a prompt "Do you like dogs, yes or no" and it never used the word yes (but it did use Yes, (yes), "yes").

### Automated Testing
No automated testing was done for this feature.


Evanev7:
Great feature, we're going to focus our efforts on our testing infra before we merge, but absolutely plan to! Thanks for your work as always

Drifter4242:
I see that the latest has temperature, top_p. It doesn't have frequency, but I don't think we should keep this pull request just for that.

Issue #1002, opened 2025-12-24
Support multimodality (Image, PDF) input


RWL-Dittrich:
My usecase for EXO involves reading and parsing PDF's to extract data from it.

One of the requirements for this is multimodality with vision support. It would be great to have this supported in EXO through file upload.

AlexCheema:
Would you like support for any specific vision model?
Would this be something you want control over to choose which vision model is used to parse the PDF?

Evanev7:
Would be great to have!

rafipatel:
Is anyone assigned/working on it?, if not I can give it a try. Sounds interesting to me.

AlexCheema:
> Is anyone assigned/working on it?, if not I can give it a try. Sounds interesting to me.

Assigned. Best of luck.

rafipatel:
Hello all, let me know further steps!
@Evanev7 @AlexCheema

alexl4321:
qwen3-vl with exo rdma / spark would be great

wangaocheng:
> qwen3-vl 搭配 exo rdma / spark 就太好了

Yes, I need too

RWL-Dittrich:
Is there anyone still working on these PR's?
We really hope this feature will be added soon as this will fulfill our usecase for EXO

Evanev7:
we've decided to prioritise continuous batching as our next milestone feature (#1642) with vision models as afterwards on our internal roadmap.

PR #1080, opened 2026-01-01
 feat: Added Images attachment support,  generator updates and related changes (#1002)

rafipatel:
## Motivation

EXO was previously limited to text-only Large Language Models (LLMs). This change is needed to expand EXO's capabilities into the multimodal space, specifically allowing users to interact with Vision-Language Models (VLMs). It solves the problem of not being able to process image inputs, which is essential for modern AI applications like image description, visual reasoning, and multimodal chat.

#1002 

## Changes

- Dependency Management: Integrated mlx-vlm as a core dependency and added torch and torchvision to pyproject.toml to support the underlying vision processing metadata required by transformer-based VLMs.
- Vision Logic: Created src/exo/worker/engines/mlx/generator/generate_vlm.py which acts as the dedicated router for multimodal requests. It includes logic to extract Base64 encoded images from various message formats (dicts, Pydantic objects, strings).

Core Engine Updates:

- Updated utils_mlx.py to correctly load vision models using mlx_vlm.load() instead of standard mlx_lm to prevent hanging.
- Modified model detection to identify "Vision Processors" by checking for the absence of eos_token_id at the top-level object.
Template System: 
- Enhanced apply_chat_template to support OpenAI-style multimodal content structures, allowing a mix of text and image_url blocks in a single message.

Model Registry: 

- Registered mlx-community/Qwen2-VL-2B-Instruct-4bit in model_cards.py and mlx-community/Qwen3-VL-2B-Instruct-4bit
- Added supports_vision=True metadata flag.
- Added a special function in model_meta, found the cause of the error. Multimodal models like SmolVLM2 and Qwen3-VL nest their architectural settings (like layer count and hidden size) inside a text_config property in config.json (can be checked from hugging face), but Exo was only looking at the top level.

Dashboard: 
- Updated the Svelte-based dashboard frontend (app.svelte.ts) to correctly serialize image uploads into the multimodal payload structure required by the updated backend.

## Why It Works

The approach works by implementing a "Vision-Aware" routing layer. Instead of trying to force VLMs into the standard LLM generation pipeline (which causes crashes because VLMs expect different input tensors and processor handling), the system now detects the model type at load time. By identifying the model as a "Vision Processor," EXO knows to use the mlx-vlm generation logic.
The implementation of extract_images_from_messages ensures that even if a model is vision-capable but the current query is text-only, the system routes it through the correct processor bridge, preventing the "LanguageModelOutput object is not subscriptable" error typically seen when standard pipelines interact with VLM heads.

### Manual Testing

Hardware : Macbook AIr M3 16GB

API Functional Verification:

- Sent a POST request containing a 100x100 Red Square Base64 image to the Qwen2-VL model. Verified the model correctly identifies visual content.

Dashboard End-to-End: 

- Uploaded image file via the dashboard UI, added a text prompt "What is in this image?", and confirmed the model's response appeared correctly in the chat window.
- Follow-up Chat: Verified that after an image is processed, the model can still handle subsequent text-only follow-ups in the same session without crashing.
- Launched an instance with two VLMs (Qwen2 and Qwen3) and tested conversations by switching between the models in same chat.
- Tested with several .JPEG and .PNG format images

### Automated Testing

Warmup Logic Testing:

- Verified that the startup/warmup sequence no longer triggers AttributeError by running uv run exo with a pre-loaded vision model.
-  The fix ensures the system can extract the tokenizer from the vision processor during initial health checks.

Multimodal Extraction Tests: 

- Used check_multimodal.py (integration script) to ensure the image extraction logic is robust against different input schemas (OpenAI format vs. simplified dashboard format). (not committed)

Screenshot 2026-01-01 at 11 51 12

Note : UV lock file dependency updates needs a second look.


src/exo/worker/engines/mlx/generator/generate.py
```diff
@@ -65,19 +70,23 @@ def warmup_inference(
...
+            max_tokens=50,
+            sampler=sampler,
+            prompt_cache=cache,
+            prefill_step_size=65536,
+            kv_group_size=KV_GROUP_SIZE,
+            kv_bits=KV_BITS,
+        ):
+            tokens_generated += 1
+    except Exception as e:
+        logger.warning(f"Warmup failed (this is usually non-fatal): {e}")
```
AlexCheema: We shouldn't be catching an exception here.
If warmup fails, then it should be retried by the worker's reconciliation loop.

src/exo/worker/engines/mlx/generator/generate.py
```diff
@@ -40,9 +41,13 @@ def maybe_quantize_kv_cache(
 
 def warmup_inference(
     model: Model,
-    tokenizer: TokenizerWrapper,
+    tokenizer: Any,
```
AlexCheema: can we explicitly type this?

src/exo/worker/engines/mlx/generator/generate.py
```diff
@@ -40,9 +41,13 @@ def maybe_quantize_kv_cache(
...
     sampler: Callable[[mx.array], mx.array],
 ) -> int:
+    # Check if this is a vision model processor
+    is_vision_processor = not hasattr(tokenizer, "eos_token_id") and hasattr(tokenizer, "tokenizer")
+    actual_tokenizer = tokenizer.tokenizer if is_vision_processor else tokenizer
```
AlexCheema: make this a fn.
why is it `tokenizer. tokenizer ` if the "actual" tokenizer is not `tokenizer`....

src/exo/worker/engines/mlx/generator/generate.py
```diff
@@ -65,19 +70,23 @@ def warmup_inference(
...
-        prompt_cache=cache,
-        prefill_step_size=65536,
-        kv_group_size=KV_GROUP_SIZE,
-        kv_bits=KV_BITS,
-    ):
-        logger.info("Generated warmup token: " + str(_r.text))
-        tokens_generated += 1
+    try:
+        # If it's a vision model, we might want to skip text-only warmup or handle it specially
+        # For now, we try text-only warmup with the actual tokenizer
```
AlexCheema: We should just do the equivalent of a warmup for vision models too.

src/exo/worker/engines/mlx/utils_mlx.py
```diff
@@ -181,16 +181,45 @@ def load_mlx_items(
...
     logger.info("Created a sampler")
 
+    # Check if this is a vision model
+    model_id = str(bound_instance.bound_shard.model_meta.model_id)
+    is_vision_model = (
+        "vl" in model_id.lower() 
+        or "vision" in model_id.lower()
+        or getattr(bound_instance.bound_shard.model_meta, 'supports_vision', False)
+    )
+
```
AlexCheema, line 195: isn't `getattr(bound_instance.bound_shard.model_meta, 'supports_vision', False)` sufficient?

src/exo/shared/types/models.py
```diff
@@ -16,3 +16,4 @@ class ModelMetadata(CamelCaseModel):
     n_layers: PositiveInt
     hidden_size: PositiveInt
     supports_tensor: bool
+    supports_vision: bool = False
```
AlexCheema: don't make this a default. lets just explicitly set it for each model.

src/exo/worker/engines/mlx/generator/generate_vlm.py
```diff
@@ -0,0 +1,142 @@
...
+                
+                if url.startswith("data:image"):
+                    try:
+                        base64_data = url.split(",", 1)[1]
+                        image_bytes = base64.b64decode(base64_data)
+                        image = Image.open(io.BytesIO(image_bytes))
+                        images.append(image)
+                        logger.info(f"Successfully extracted image. Current count: {len(images)}")
+                    except Exception as e:
+                        logger.error(f"Failed to decode image: {e}")
```
AlexCheema: what's the point of try excepting an continuing? doesn't this mean it failed? the correct thing is to raise the exception

src/exo/worker/engines/mlx/generator/generate_vlm.py
```diff
@@ -0,0 +1,142 @@
...
+    max_tokens = task.max_tokens or 512
+    temperature = task.temperature or 0.7
+    
+    try:
+        # Use mlx_vlm's generate function
+        output = generate(
+            model,
+            processor,
+            prompt,
+            images[0] if images else None,  # Qwen2-VL takes single image
```
AlexCheema: Qwen2-VL takes a single image but what about other vision models? do they support multiple images?

src/exo/worker/engines/mlx/generator/generate_vlm.py
```diff
@@ -0,0 +1,142 @@
...
+    except Exception as e:
+        logger.error(f"Error during vision model generation: {e}")
+        import traceback
+        logger.error(traceback.format_exc())
+        # Yield error message
+        yield GenerationResponse(
+            text=f"[Error generating response: {str(e)}]",
+            token=0,
+            finish_reason="stop",
+        )
```
AlexCheema: I don't think this is the right place to be handling an exception like this.

src/exo/worker/engines/mlx/utils_mlx.py
```diff
@@ -282,37 +311,58 @@ def get_tokenizer(model_path: Path, shard_metadata: ShardMetadata):
...
-            {k: v for k, v in message.model_dump().items() if v is not None}  # type: ignore
-        )
-
-    prompt: str = tokenizer.apply_chat_template(  # type: ignore
+        # Construct the message dict carefully
+        msg_dict = {"role": message.role}
+        if content is not None:
+            msg_dict["content"] = content
+        if message.name:
+            msg_dict["name"] = message.name
```
AlexCheema: All of this can be extracted into a function?

src/exo/worker/engines/mlx/generator/generate_vlm.py
```diff
@@ -0,0 +1,142 @@
...
+                token=idx,
+                finish_reason=None,
+            )
+        
+        # Final token with finish reason
+        yield GenerationResponse(
+            text="",
+            token=len(tokens),
+            finish_reason="stop",
+        )
```
AlexCheema: This all seems quite hacky.
 shouldn't we be detokenizing and passing that in `text` field of `GenerationResponse`?

src/exo/worker/engines/mlx/generator/generate_vlm.py
```diff
@@ -0,0 +1,142 @@
...
+    prompt = apply_chat_template(processor, task)
+    
+    if not images:
+        logger.info("No images found in messages for vision model. Proceeding with text-only inference.")
+    else:
+        logger.info(f"Generating with prompt: {prompt[:100]}... and {len(images)} image(s)")
+    
+    # Generate response
+    max_tokens = task.max_tokens or 512
+    temperature = task.temperature or 0.7
```
AlexCheema: magic number defaults.
put these in `constants.py`

src/exo/worker/engines/mlx/generator/generate.py
rafipatel: Yes, on it.
and I see the confusion, in vision processor its wrapped in tokenizer.tokenizer, I will make it more understandable

src/exo/worker/engines/mlx/generator/generate_vlm.py
rafipatel: I have not tried with models that has multiple images, I will update you on this, although removing the indexing from the code should work for models supporting multiple images, but will throw error if they do not support, I am testing it

src/exo/worker/engines/mlx/generator/generate.py
rafipatel: Thought about it, doing a multimodal warmup, passing a dummy 32x32 RGB image then the normal text during warmup.

src/exo/worker/engines/mlx/utils_mlx.py
rafipatel: You are right, its unnecessary and extra

src/exo/worker/engines/mlx/generator/generate.py
rafipatel: "Done ✅ removed the try/except"

src/exo/worker/engines/mlx/generator/generate.py
rafipatel: Done ✅ Fixed in commit 98eb417

src/exo/worker/engines/mlx/generator/generate.py
rafipatel: Done ✅ Fixed in commit 98eb417

src/exo/worker/engines/mlx/utils_mlx.py
rafipatel: Done ✅ Fixed in commit 98eb417

src/exo/worker/engines/mlx/generator/generate_vlm.py
rafipatel: Done ✅ Fixed in commit 98eb417

src/exo/worker/engines/mlx/utils_mlx.py
rafipatel: Done ✅ Fixed in commit 98eb417

src/exo/worker/engines/mlx/generator/generate_vlm.py
rafipatel: Done ✅ Fixed in commit 98eb417

src/exo/worker/engines/mlx/generator/generate_vlm.py
rafipatel: Done ✅ Fixed in commit 98eb417

src/exo/worker/engines/mlx/generator/generate.py
rafipatel: Done ✅ Fixed in commit 98eb417

src/exo/shared/types/models.py
rafipatel: Done ✅ Fixed in commit 4f16e1f

src/exo/worker/engines/mlx/generator/generate_vlm.py
rafipatel: Tried by passing multiple images, seems to work, I think I need to test it with more robust models

Done ✅ Fixed in commit 98eb417

src/exo/worker/engines/mlx/generator/generate_vlm.py
rafipatel: Done ✅ Fixed in commit 98eb417

rafipatel:
## Summary of Changes

**1. Error Handling**
- Removed try/except block in vision model generation - exceptions now propagate to caller

**2. Type Safety**
- Added `Union[TextTokenizer, VisionProcessor]` type hints
- Created Protocol classes for both tokenizer types

**3. Model Metadata**
- Removed `supports_vision=False` default value
- Now requires explicit `supports_vision` value for each model in `MODEL_CARDS`

**4. Vision Model Integration**
- Unified vision and text-only generation into single `mlx_generate()` function
- Improved warmup to properly handle vision models with dummy images

Note : Deleted separate file for vision language model generate function, and moves image extraction function in utils_mlx, thought was a good approach and extra file was unnecessary

AlexCheema:
Edit: closed by mistake.

AlexCheema:
Sorry closed by mistake.

AlexCheema:
Do you mind fixing conflicts so we can get this merged?

Issue #957, opened 2025-12-22
[MEDIUM] Better placement algorithm for pipeline parallelism using memory bandwidth + latency


AlexCheema:
This is our placement algorithm for pipeline parallelism:
https://github.com/exo-explore/exo/blob/abaeb0323d4182f7bc4dd3775a8ba9209117d1cf/src/exo/master/placement_utils.py#L52-L100

It places a number of layers proportional to the memory available on each machine. This is not optimal.

In order to maximize TPS in the memory-bound regime (i.e. low batch_size), we should look at memory bandwidth and latency. The time spent computing on device `i` would be `C_i = M * (N_i / N)/B_i` where `M` is the total memory used by the entire model, `N` is the number of layers in the model, `N_i` is the number of layers on device `i` and `B_i` is the memory bandwidth of device `i`. The time spent communicating between device `i` and device `i + 1` is `L_i`. Then the total time for one token for `k` devices is sum from 1 to `k` of `C_i + L_i`. This sum is what we want to minimize. I'l leave the details of the algorithm up to the implementer.

abendrothj:
@AlexCheema I'll tackle this. Based on the issue description, the goal is to replace the current RAM-proportional split with a bandwidth-aware distribution to minimize total inference time.

Since the cycle (topology) is fixed during shard assignment, ∑Li​ is constant. Minimizing the total time ∑(Ci​+Li​) effectively reduces to minimizing ∑Ci​.

### **My plan**

Data Collection: I will add a lookup table mapping chip_id to memory bandwidth (e.g. 400GB/s for M3 Max) in NodePerformanceProfile, as this data is currently missing.

**Algorithm Update**: I will modify get_shard_assignments_for_pipeline_parallel in src/exo/master/placement_utils.py to use a greedy approach:

1. Reserve: Assign 1 layer to every node in the cycle (required to prevent IndexError in auto_parallel.py where networking layers are injected).
2. Sort: Sort nodes by bandwidth (Bi​).
3. Fill: calculate the "cost per layer" for each node, and saturate the fastest nodes with as many of the remaining layers (Ni​) as their RAM permits.
4. Finalize: Map these counts back to the original topology order to generate the final ShardAssignments.

I would start by adding the bandwidth specs for Apple Silicon. Let me know if you'd like any modifications to my approach to ensure a clean pr. :)

AlexCheema:
Happy for you to take this one.

Your approach looks good except it doesn’t take into account latency.

[...]

Issue #11, opened 2024-07-16
Example setups with benchmarks


AlexCheema:
Users want to know exactly which setups work, how to set them up, and what the benchmarks are.

Ref:

<img width="645" alt="Screenshot 2024-07-16 at 09 11 37" src="https://github.com/user-attachments/assets/b40d5252-cdfa-4c8f-996c-6f9eebfd3755">

AlexCheema:
Request: Raspberry Pi. We could bundle it with Coral USB TPU (https://coral.ai/products/) which could be super cost effective home ai inference.

<img width="324" alt="Screenshot 2024-07-16 at 10 24 06" src="https://github.com/user-attachments/assets/7fb7fa1c-fa9c-47fa-9695-0a24f0b802ca">

AlexCheema:
List of supported hardware:

<img width="409" alt="Screenshot 2024-07-16 at 13 46 40" src="https://github.com/user-attachments/assets/a322bcae-5612-429f-be6d-6e4a00fd5bba">

AlexCheema:
2x4090 on exo vs bunch of 4090s in one pc

![IMG_0076](https://github.com/user-attachments/assets/e2b223fd-fe23-46e8-87cf-d1517f462303)

audkar:
> We could bundle it with Coral USB TPU 

Coral is a little bit hard to get nowadays 

There is also this partnership https://www.raspberrypi.com/news/raspberry-pi-ai-kit-available-now-at-70/

AlexCheema:
The people want to know how fast it is.

![IMG_7523](https://github.com/user-attachments/assets/e8f18dd6-1130-4ac1-8bf9-4532d032afa5)

AlexCheema:
<img width="591" alt="Screenshot 2024-07-19 at 02 22 29" src="https://github.com/user-attachments/assets/68990555-df17-46ac-bd64-bfe8977a9efd">

AlexCheema:
![IMG_0093](https://github.com/user-attachments/assets/cfff1e8c-c872-4795-acff-f52c05ff4a8d)

foucist:
I assume it would be possible to run the Llama 3.1 405B Q4 model (234GB)  across 4 x 64GB Mac Studio or 2 x 128GB Mac Studio ?  would be interesting to see token speeds for sure.

jhgoodwin:
Yes, ideally, I'd like to know:
- Cluster details
- Model used, including bit size
[...]

PR #1088, opened 2026-01-03
feat: implement bandwidth-aware shard assignment for pipeline parallelism

abendrothj:
This PR implements bandwidth-aware shard assignment for pipeline parallelism, minimizing total inference time by assigning more layers to devices with higher memory bandwidth. This aligns with Issue #957.

## Changes
- **Profiling**: Added `memory_bandwidth` to `NodePerformanceProfile` and Apple Silicon bandwidth data.
- **Assignment**: Implemented greedy assignment algorithm in `placement_utils.py`.
- **Verification**: Added `test_get_shard_assignments_bandwidth_aware` to confirm optimal distribution.


src/exo/worker/utils/system_info.py
```diff
@@ -81,3 +81,35 @@ async def get_model_and_chip() -> tuple[str, str]:
...
+    "Apple M2 Ultra": 800_000_000_000,
+    # M3 Series
+    "Apple M3": 100_000_000_000,
+    "Apple M3 Pro": 150_000_000_000,
+    "Apple M3 Max": 400_000_000_000,
+    # M4 Series
+    "Apple M4": 120_000_000_000,
+    "Apple M4 Pro": 273_000_000_000,
+    "Apple M4 Max": 546_000_000_000,
+}
```
AlexCheema: Given we support linux too + Spark support is coming, we should not hardcode. We should profile this on each device exo is running on.

src/exo/master/placement_utils.py
```diff
@@ -53,43 +53,147 @@ def get_shard_assignments_for_pipeline_parallel(
...
+            can_take = max(0, max_layers_by_ram - assignments[original_idx])
+
+            take = min(can_take, remaining_layers)
+            assignments[original_idx] += take
+            remaining_layers -= take
+
+        # Distribute any leftovers (if RAM constrained everywhere) to highest bandwidth nodes again (or round robin)
+        # Assuming we just dump onto the fastest node even if it OOMs?
+        # Or better -> round robin among sorted nodes?
+        # The prompt says "saturate the fastest nodes...".
```
AlexCheema: These comments are a bit...verbose

src/exo/master/placement_utils.py
```diff
@@ -53,43 +53,147 @@ def get_shard_assignments_for_pipeline_parallel(
...
+            layers_assigned += node_layers
+
+        return ShardAssignments(
+            model_id=model_meta.model_id,
+            runner_to_shard=runner_to_shard,
+            node_to_runner=node_to_runner,
+        )
 
-        runner_id = RunnerId()
+    # 2. Bandwidth-aware logic
```
AlexCheema: Would definitely want to refactor this into seperate smaller functions.

AlexCheema (changes requested):
The next leap up for us is actual live profiling of devices and doing memory-bandwidth-aware placement based on that.

The code in this PR needs to be refactored as we have one massive function for placement. Placement will continue to get more complex as we introduce more factors e.g. latency, or model downloads, so it's important we keep this part of the code simple and clean.

src/exo/worker/utils/system_info.py
abendrothj: Using profiling in latest commit

src/exo/master/placement_utils.py
abendrothj: Removed

src/exo/master/placement_utils.py
abendrothj: Refactored

abendrothj:
Should be refactored into the new architecture @AlexCheema

AlexCheema:
Edit: Sorry, closed by mistake.

AlexCheema:
Sorry, closed by mistake.

AlexCheema:
Could you please have a look at merge. conflicts?

abendrothj:
> Could you please have a look at merge. conflicts?

Resolved

AlexCheema:
## Code Review — PR #1088: feat: implement bandwidth-aware shard assignment for pipeline parallelism

**CI status:** No status checks reported (empty `statusCheckRollup`). Cannot confirm CI passes.

---

### Overview

This PR adds memory-bandwidth-aware layer assignment for pipeline parallel shard placement. When bandwidth data is available for all nodes in a cycle, layers are distributed proportionally to memory bandwidth rather than proportionally to available RAM. The implementation profiles memory bandwidth at startup on macOS using MLX GPU operations, stores it in cluster state via the event-sourcing pipeline, and threads it through to the placement algorithm.

**Files changed (9):** `state.py`, `profiling.py`, `apply.py`, `info_gatherer.py`, `system_info.py`, `placement.py`, `placement_utils.py`, `main.py`, `test_placement_utils.py`

---

### Critical Issues

**1. `node_bandwidth` is not cleaned up on node timeout**

When a node is removed from the cluster via `NodeTimedOut`, `apply_node_timed_out` cleans up every per-node mapping (`node_memory`, `node_disk`, `node_system`, `node_network`, `node_thunderbolt`, `node_thunderbolt_bridge`, `node_rdma_ctl`) but the new `node_bandwidth` mapping is not cleaned up. This means stale bandwidth data for departed nodes will persist in state indefinitely.

```python
# src/exo/shared/apply.py — apply_node_timed_out is missing:
node_bandwidth = {
    key: value for key, value in state.node_bandwidth.items() if key != event.node_id
}
# ... and "node_bandwidth": node_bandwidth in the update dict
```

**2. Greedy assignment algorithm produces degenerate distributions**

The `_distribute_layers_by_bandwidth` function assigns layers greedily to the highest-bandwidth node first, constrained only by RAM. As the test itself demonstrates, with nodes of 400/200/100 GB/s bandwidth and 1 GB RAM each (far exceeding model size), the result is **A=28, B=1, C=1**. This is not "bandwidth-aware proportional assignment" — it is "dump everything on the fastest node." The entire benefit of distributed inference (spreading work across devices) is lost. A proportional allocation (e.g., `400:200:100 -> ~17:9:4` for 30 layers) would minimize total inference time while actually using the cluster, since total time is dominated by the slowest pipeline stage.

The greedy approach is only optimal if you assume zero network transfer cost, which is unrealistic for pipeline parallelism where every stage boundary incurs a network round-trip.

**3. `NodeMemoryBandwidth` in info_gatherer duplicates the name with the one in `apply.py`**

The PR defines `NodeMemoryBandwidth` in *two different modules* with different meanings:
- `src/exo/utils/info_gatherer/info_gatherer.py`: a `TaggedModel` subclass used as gathered info
- Referenced in `src/exo/shared/apply.py`: imported from info_gatherer as part of the `GatheredInfo` union

However, the `NodeMemoryBandwidth` import in `apply.py` (the `match info:` block) comes from `info_gatherer.py`, while `NodeBandwidth` (the state type) comes from `profiling.py`. This works but is confusing — having `NodeMemoryBandwidth` (gathered info), `NodeBandwidth` (state type), and `memory_bandwidth` (the int field) all floating around makes the naming inconsistent with the rest of the codebase's cleaner patterns.

---

### Significant Issues

**4. The profiling benchmark allocates 2 GB of GPU memory and is not gated on available memory**

```python
# src/exo/utils/info_gatherer/system_info.py
size_bytes = 2 * 1024 * 1024 * 1024
```

This allocates a 2 GB buffer on the Metal GPU (plus a second 2 GB for the copy). On devices with limited unified memory or when a model is already loaded, this could cause memory pressure or OOM. There is no check that sufficient memory is available before allocation, and no cleanup guarantee if the operation fails mid-way (MLX garbage collection is non-deterministic).

**5. Profiling runs synchronously in a thread at startup, blocking info gathering**

```python
# info_gatherer.py
bandwidth = await NodeMemoryBandwidth.gather()
await self.info_sender.send(bandwidth)
```

`NodeMemoryBandwidth.gather()` runs `profile_memory_bandwidth` via `to_thread.run_sync` with up to 5 retry attempts, each sleeping 1 second on failure. In the worst case this blocks the info gatherer's `run()` method for ~5+ seconds after all the other startup gathering completes. During this time, no other info gathering tasks run (though existing periodic tasks already started in the task group are unaffected). This is a minor startup latency concern but worth noting.

**6. `_distribute_layers_by_bandwidth` uses RAM-based capping that doesn't match the existing memory validation**

The bandwidth path computes `layer_size_bytes = model_card.storage_size.in_bytes / model_card.n_layers` and caps assignment based on `ram_available`. But the existing RAM-proportional path in `_allocate_and_validate_layers` does a post-hoc validation that is more rigorous. The two paths have inconsistent memory constraint logic, and the bandwidth path's "dump remaining layers on fastest nodes" fallback completely ignores memory constraints:

```python
if remaining_layers > 0:
    logger.warning("All nodes maxed out on RAM estimation, dumping remaining layers on fastest nodes.")
    for original_idx, _ in sorted_nodes:
        assignments[original_idx] += 1  # No RAM check!
```

**7. Bandwidth data is gathered only once at startup and never refreshed**

Unlike memory usage (polled every second), disk usage (polled every 30s), and other metrics, bandwidth is gathered exactly once. If conditions change (thermal throttling, different power state), the stale value persists for the lifetime of the process. This is acceptable for Apple Silicon where bandwidth is fixed, but the code has no comment explaining this design decision, and it won't generalize well to other hardware.

---

### Minor Issues

**8. `memory_bandwidth: int | None = None` in `NodeBandwidth` allows `None` to be stored in state**

```python
class NodeBandwidth(CamelCaseModel):
    memory_bandwidth: int | None = None
```

If profiling fails, `NodeBandwidth(memory_bandwidth=None)` is stored in state. The placement code must then defensively filter these out (which it does via the `if nb.memory_bandwidth is not None` check), but it would be cleaner to simply not emit the event if profiling fails.

**9. No bandwidth-aware assignment for CFG parallel sharding**

The bandwidth parameter is only threaded through to `_get_shard_assignments_for_pure_pipeline`. CFG parallel (which also does pipeline-style layer assignment) does not benefit from bandwidth awareness. This is not documented.

**10. f-string in logger call**

```python
logger.info(f"Memory bandwidth: {bandwidth / 1e9:.1f} GB/s")
```

This evaluates the f-string even when the log level is above INFO. Use lazy formatting: `logger.info("Memory bandwidth: {:.1f} GB/s", bandwidth / 1e9)`.

**11. Test uses `Memory.from_kb(300)` for storage_size with `n_layers=30`**

This gives 10 KB per layer, meaning each node's ~1 GB RAM can hold all layers easily. The test only validates the greedy behavior, not the RAM-capping behavior. A test with constrained RAM would be valuable.

---

### What's Good

- The event-sourcing integration follows the established pattern correctly (new `GatheredInfo` variant, new state field, `match` arm in `apply_node_gathered_info`).
- The `profile_memory_bandwidth` function uses sensible benchmarking technique (warm-up, multiple runs, best-of).
- Fallback to RAM-proportional assignment when bandwidth data is missing or incomplete is a good defensive design.
- The test for fallback behavior (`test_get_shard_assignments_falls_back_without_bandwidth`) verifies backward compatibility.
- The PR correctly gates bandwidth profiling to macOS only (`IS_DARWIN`).

---

### Verdict

The PR has the right structure and follows existing patterns, but has several issues that need addressing before merge:

1. **Must fix:** `node_bandwidth` cleanup in `apply_node_timed_out` (state leak).
2. **Must fix:** The greedy allocation algorithm is fundamentally flawed for pipeline parallelism — it should distribute layers proportionally to bandwidth, not greedily dump everything on the fastest node.
3. **Should fix:** Memory safety of the 2 GB GPU allocation during profiling.
4. **Should fix:** The RAM-bypass fallback in `_distribute_layers_by_bandwidth`.

Recommend **changes requested**.

abendrothj:
Thanks for the review @AlexCheema- I've fixed the issues, and request a re-review.
## Critical Fixes ✅
node_bandwidth cleanup - Added cleanup in `apply_node_timed_out()` to remove stale bandwidth data
Proportional distribution - Rewrote allocation algorithm: 400:200:100 GB/s now gives 16:9:5 layers (not 28:1:1). Balances work across cluster to minimize pipeline time
RAM-bypass removed - Dangerous fallback that ignored memory constraints is gone
## Significant Fixes ✅
Memory safety - Added ~4GB allocation warning in docstrings
Lazy logging - Fixed f-string: [logger.info("Memory bandwidth: {:.1f} GB/s", bandwidth / 1e9)](vscode-file://vscode-app/Applications/Visual%20Studio%20Code.app/Contents/Resources/app/out/vs/code/electron-browser/workbench/workbench.html)
Type annotations - Added explicit types (0 basedpyright errors)
## Minor Issues ✅
CFG parallel - Now bandwidth-aware! (commit 8ee0b6c6)
Single sample - Added comments explaining this is intentional for Apple Silicon's hardware-fixed bandwidth (commit 5cbf167d)

abendrothj:
Rebased onto current main (merge commit cc6244f6 — resolved conflicts with the zenoh migration, `node_backends` state, and the new `place_instance` signature) and addressed the remaining review items in 57cb2184:

- **#1 (state leak on `NodeTimedOut`)** — was already fixed in 43b450b1; `node_bandwidth` is pruned alongside the other per-node mappings.
- **#2 (greedy allocation)** — was already fixed in 43b450b1; layers are distributed proportionally to bandwidth (largest-remainder), RAM-capped, with proportional redistribution of any overflow. The test asserts a 16/9/5 split for 4:2:1 bandwidths.
- **#4 (unguarded 2 GB allocation)** — the profiling buffer is now sized to at most a quarter of currently available memory (working set ≤ half, since the copy needs 2× the buffer), and profiling is skipped below a 256 MB floor.
- **#5 (blocking startup)** — the bandwidth gather now runs as a concurrent task in the info-gatherer's task group, so its retries never delay other gathering.
- **#6 (RAM-bypass fallback)** — was already fixed in 43b450b1; the redistribution loop respects per-node capacity and errors out rather than over-assigning.
- **#8 (`None` stored in state)** — failed profiling now emits no event at all; `NodeBandwidth.memory_bandwidth` is non-optional.
- **#9 (CFG parallel)** — was already added in 8ee0b6c6.
- **#11 (RAM-capping untested)** — added `test_get_shard_assignments_bandwidth_aware_respects_ram_capacity`: a 4× faster node whose RAM fits only 3 of 10 layers gets capped at 3 with the excess flowing to the slower node.

basedpyright, ruff, and the master/shared/utils test suites all pass locally. On #3 (naming): `NodeMemoryBandwidth` (gathered info) → `NodeBandwidth` (state) follows the same two-type pattern as `RdmaCtlStatus` → `NodeRdmaCtlStatus`, so I've left it — happy to rename if you'd prefer.

One note for sequencing: this touches the same allocation functions as #2214 (manual layer allocation), so whichever lands first will leave the other a small conflict — I'll rebase the survivor.

PR #1093, opened 2026-01-05
fix: pass temperature to MLX sampler dynamically

chriscod3:


Evanev7:
we're still working on our testing but I think this is the golden case for our integration test server. once we have that up and running we'll get these merged

Evanev7:
Implemented with #1167

PR #1094, opened 2026-01-05
Fix mlx seed

chriscod3:


rltakashige:
LGTM

Tested locally and fixed merge conflicts. There are still some other arguments passed in `ChatCompletionTaskParams` that aren't handled yet, but this is a good start. Thanks for the contribution!

PR #1098, opened 2026-01-06
add a server for distributed testing in /tests until we work out a stable solution.

Evanev7:
## Motivation

Testing multiple devices simultaneously requires coordination, and we don't necessarily want to run a full EXO to test single components. We need a mid-scale integration testing framework for distributed tests.

## Changes

Add a simple python server + bash query that runs Jaccl and Ring tests without constructing a worker/master/networking. The query relies on all devices being accessible over tailscale, currently.

## Test Plan

uhhh lmao

### DIFF
--- a/src/exo/worker/runner/bootstrap.py
+++ b/src/exo/worker/runner/bootstrap.py
@@ -6,7 +6,7 @@ from exo.shared.types.events import Event, RunnerStatusUpdated
 from exo.shared.types.tasks import Task
 from exo.shared.types.worker.instances import BoundInstance, MlxJacclInstance
 from exo.shared.types.worker.runners import RunnerFailed
-from exo.utils.channels import MpReceiver, MpSender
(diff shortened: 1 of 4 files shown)


src/exo/worker/runner/bootstrap.py
```diff
@@ -42,8 +42,11 @@ def entrypoint(
             )
         )
     finally:
-        event_sender.close()
-        task_receiver.close()
+        try:
+            event_sender.close()
+            task_receiver.close()
+        except ValueError:
+            pass
```
JakeHillion: This is if the channels are already closed?

JakeHillion (approved):
LGTM. Can you move the scripts somewhere more reasonable than /tmp? I don't expect we'll replace these for a long time.

src/exo/worker/runner/bootstrap.py
Evanev7: Yup, multiprocessing queues raise a valueerror on double-close

PR #1099, opened 2026-01-06
exo-bench (Benchmark model pp & tg speed)

rltakashige:
## Motivation

This PR implements benchmarking in the style of llama-bench. The main difficulty here is the fact that exo is not a library - it exposes an endpoint. This means that benchmarking numbers will be inaccurate if the API is measured.

The solution assumes nodes are set up with uv run exo (or via the app), and then hits the new endpoint /bench/chat/completions to retrieve generation statistics directly from mlx_lm. 

This will allow us to release benchmarks for models and perform regression tests.

TODO: Performance benchmarking.

## Changes

- Adds /bench/chat/completions endpoint
- Adds BenchChatCompletion/Response
- Adds a logits processor to prevent response from ending early
- Adds a "Prompt Sizer" which downloads the tokenizer and dynamically adjusts the prompt of "a" to fit the desired prompt size.
- Reduce prefill step size to 2048 for now (in future, dynamically adjust this value)

### Manual Testing

Benchmarked Llama, Qwen, DeepSeek and Kimi models. Will require several fixes to run consistently on all configurations (to be done in the future). 
Manually tested the normal API to verify chat requests complete as expected.

### Automated Testing

Not really possible. Type checker passes.

### DIFF
--- a/bench/exo_bench.py
+++ b/bench/exo_bench.py
@@ -274,7 +274,9 @@ def main() -> int:
         description="Benchmark exo model throughput across placement previews.",
     )
     ap.add_argument("--host", default=os.environ.get("EXO_HOST", "localhost"))
-    ap.add_argument("--port", type=int, default=int(os.environ.get("EXO_PORT", "52415")))
+    ap.add_argument(
+        "--port", type=int, default=int(os.environ.get("EXO_PORT", "52415"))
+    )
     ap.add_argument("--model", required=True, help="Model short id or huggingface id")
     ap.add_argument(
         "--pp",
@@ -407,7 +409,7 @@ def main() -> int:
             str(p.get("sharding", "")),
             -nodes_used_in_instance(p["instance"]),
         ),
-        reverse=True
+        reverse=True,
     )
 
     logger.debug(f"exo-bench model: short_id={short_id} full_id={full_model_id}")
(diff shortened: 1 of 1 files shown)


src/exo/worker/engines/mlx/generator/generate.py
```diff
@@ -72,25 +77,47 @@ def warmup_inference(
...
+    return eos
+
+
 def mlx_generate(
     model: Model,
     tokenizer: TokenizerWrapper,
     sampler: Callable[[mx.array], mx.array],
     task: ChatCompletionTaskParams,
 ) -> Generator[GenerationResponse]:
+    mx.reset_peak_memory()
```
JakeHillion, line 122: Why?

src/exo/worker/engines/mlx/generator/generate.py
```diff
@@ -72,25 +77,47 @@ def warmup_inference(
...
         kv_group_size=KV_GROUP_SIZE,
         kv_bits=KV_BITS,
     ):
         logger.info("Generated warmup token: " + str(_r.text))
         tokens_generated += 1
 
     logger.info("Generated ALL warmup tokens")
-    mx_barrier()
+
+    # TODO: Do we want an mx_barrier?
```
JakeHillion, line 90: Might be worth adding this back for now, if it's not essential to this PR. There's an open discussion about whether we need more of these barriers.

JakeHillion (approved):
Couple of comments of bits of code that likely need commenting, but this looks great.

PR #1109, opened 2026-01-08
Use presigned URLs for bug report uploads

AlexCheema:
## Motivation

Previously we hardcoded AWS credentials into the app.
This is not good practice.

## Changes

Use presigned URLs instead.

## Why It Works

Presigned URLs are an S3 feature for this kind of thing. They provide an expiring presigned URL with certain permissions. In this case we have a presigned URL with `s3:PutObject` permission that expires after 5 minutes. The client uses this presigned URL to upload a bug report instead of using its own credentials to sign a request. This also simplifies a lot of the Swift code.

### Manual Testing
On a single MacBook, I downloaded the app and sent a bug report. It worked and appeared in the bucket.


JakeHillion (approved):
Nice.

PR #1110, opened 2026-01-08
ci: avoid uploading alpha appcasts

JakeHillion:
Currently alpha appcasts get uploaded. It turns out these overwrite the standard appcast, so even though no one will update to the alpha channel, everyone will miss regular updates while the latest build was an alpha one.

Ideally we should combine the source of truth for both the alpha and release channels, but as no one is using the alpha channel for yet let's stop uploading it for now.

Test plan:

eyes


AlexCheema (approved):
looks gud

Issue #1074, opened 2025-12-31
Support Tool Use via the API


lucasajackson:
It seems tool use is not supported when using the EXO API endpoint on the Mac app.  I'm running this on MacOS 26.2 with RDMA enabled between two Mac Studios.  When trying to configure Synology AI Console to use EXO I get an error that EXO does not support tool use when using the gpt-oss-120b model.  This model does work properly when configuring Synology AI Console with LM Studio.  I've also run into issues trying to use OpenCode with EXO using both the Qwen-Coder-3-480b and gpt-oss-120b models.  OpenCode is able to generate code; however, it will not create/edit files.  When trying gpt-oss-120b with LM Studio, OpenCode is able to create and edit files without issue.

PR #1118, opened 2026-01-09
Tool Calling

Evanev7:
ayo lets do it bigger

Closes: #1074


frenchtoasters:
@Evanev7 Is there a replacement PR for this one? Tool calling is something im working to implement locally but would much rather work off a base from someone who is deeper in the weeds than I am.

Issue #910, opened 2025-12-18
[TRACKING] Linux CPU Support


Evanev7:
Check PLATFORMS.md for up to date information about current support.

triko88:
I investigated the code from the `main` branch. And found the following:
- The tinygrad added as the inference backend is a compiled PYC bytecode, that too without an entry point. Which primarily blocks out Linux users.
- Tinygrad supports CPU, CUDA, ROCm and WebGPU as of now on Linux. Providing tinygrad entry point will cover most of the Linux support at once.

Follow [my fork's branch for more updates on Linux support](https://github.com/triko88/exo/tree/feature/linux-support)

Edit: removed the claim of PYC bytecode compatibility as it is incorrect. The lack of entry point is still true.

test-j-l:
Sorry, I'm a newbie and I'm not quite clear about it. I've read the platforms.md document. It seems that the Intel platform is not supported yet for the time being, but the plan is to support it and it's under development, right?

Evanev7:
Yup! This is probably our first non-apple platform we want to support.

test-j-l:
OK, thank you very much for your recovery. I also have a question. Do you have a rough completion time plan for the support for Linux at present? For example, December 2026? Or December 2027, etc.

Evanev7:
Definitely this year - I can't commit to a timeline but I should expect the next 3 months we should have CPU support down - presuming the upstream issues get fixed.

Evanev7:
Update: this is actually an env issue. Upstream mlx[cpu] is compiled with g++12, so you need g++12 to be your system default. With that change made, exo runs (very very slowly!) on linux CPU!

TazDevl74:
Think I have it working with gcc-11 & g++-11, al least that's when I stopped getting float errors

```
# Replace gcc
sudo apt clean
sudo apt autoremove --purge
sudo apt remove gcc-12 g++-12 g++-13 -y
sudo apt autoremove -y
sudo apt install gcc-11 g++-11 -y

# Make g++ point to g++-11
# You can create a symlink so g++ resolves to g++-11:
sudo update-alternatives --install /usr/bin/gcc gcc /usr/bin/gcc-11 100
sudo update-alternatives --install /usr/bin/g++ g++ /usr/bin/g++-11 100
sudo update-alternatives --config gcc
sudo update-alternatives --config g++
```

But... on a NUC without GPU this took ages:
[...]

PR #1119, opened 2026-01-09
nix mlx compilation for better portability

Evanev7:
we want to build exo with nix for reproducibility to avoid the g++ versioning issues we have been seeing in #910. We also need to build mlx for aarch64-linux systems like the DGX Spark.

## Changes

Added mlx.nix with the aim of upstreaming once its stable cross-platform.
Added xcode.nix as a shim for the build platforms Xcode.

### DIFF
--- a/nix/mlx.nix
+++ b/nix/mlx.nix
@@ -1,7 +1,9 @@
 { stdenv
 , lib
-, buildPythonPackage
 , fetchFromGitHub
+, fetchPypi
+, pyprojectHook
+, pypaInstallHook
 , replaceVars
 , fetchzip
 , setuptools
@@ -32,7 +34,7 @@ let
     hash = "sha256-7n2eI2lw/S+Us6l7YPAATKwcIbRRpaQ8VmES7S8ZjY8=";
   };
 
-  mlx = buildPythonPackage rec {
+  mlx = stdenv.mkDerivation rec {
     pname = "mlx";
     version = "0.30.1";
     pyproject = true;
@@ -98,6 +100,8 @@ let
     nativeBuildInputs = [
       cmake
       metal
+      pyprojectHook
+      pypaInstallHook
     ];
 
     buildInputs = [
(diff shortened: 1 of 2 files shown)


Evanev7:
Now compiles mlx metal for darwin.

nix/mlx.nix
```diff
@@ -0,0 +1,161 @@
...
+
+    # Run the mlx Python test suite.
+    nativeCheckInputs = [
+      numpy
+      pytestCheckHook
+    ];
+
+    enabledTestPaths = [
+      "python/tests/"
+    ];
```
JakeHillion, line 125: can we get a comment on why we're skipping some tests?

nix/mlx.nix
```diff
@@ -0,0 +1,161 @@
...
+, cmake
+, zsh
+  # buildInputs
+, nanobind
+, pybind11
+, nlohmann_json
+  # buildInputs (darwin)
+, apple-sdk_26
+, metal
+  # tests
```
JakeHillion: a bunch of commented out inputs here, and can we alphabeticise them?

nix/metalWrapper.nix
```diff
@@ -0,0 +1,19 @@
...
+  version = metalVersion;
+
+  __noChroot = true;
+  buildCommand = ''
+    DEVELOPER_DIR=${xcodeBaseDir}/Contents/Developer
+    SDKROOT=${xcodeBaseDir}/Contents/Developer/Platforms/MacOSX.platform/Developer/SDKs/MacOSX.sdk
+    export DEVELOPER_DIR SDKROOT
+    mkdir -p $out/bin && cd $out/bin
+    ln -s $(/usr/bin/xcrun -f metal)
+    ln -s $(/usr/bin/xcrun -f metallib)
```
JakeHillion: Can we make this error if "$out" doesn't have the expected files at the end? I'd rather this failed than the "pure" derivation.

Evanev7:
Superseded by #1285

PR #1127, opened 2026-01-12
Support for FLASH

samiamjidkhan:
## Motivation

Enables Exo to orchestrate FLASH MPI simulations across distributed Mac clusters. FLASH is a multiphysics astrophysics simulation code used for hydrodynamics, MHD, and radiation transport research.

 Previously, running distributed FLASH simulations required manual hostfile creation and coordination. This integration allows Exo to:
  - Automatically discover nodes via existing mDNS/gossipsub
  - Generate MPI hostfiles from topology
  - Launch and monitor simulations via REST API
  - Report status through the existing dashboard infrastructure

## Changes

New files:

- src/exo/worker/runner/flash_runner.py - MPI runner that generates hostfiles and spawns mpirun

Modified files:

- src/exo/shared/types/worker/instances.py - Added FLASHInstance type
- src/exo/shared/types/commands.py - Added LaunchFLASH and StopFLASH commands
- src/exo/master/placement.py - Added place_flash_instance() for node selection
- src/exo/master/main.py - Added command handlers for FLASH launch/stop
- src/exo/master/api.py - Added REST endpoints: POST /flash/launch, GET /flash/instances, DELETE /flash/{instance_id}
- src/exo/worker/plan.py - Added _plan_flash() for FLASH instance lifecycle
- src/exo/worker/runner/bootstrap.py - Added routing to flash_runner for FLASHInstance

## Why It Works

EXO already handles node discovery and topology management. This PR bridges EXO's topology to MPI by:

- Placement: Selects nodes from EXO topology, creates shard assignments (one runner per node)
- Hostfile generation: Converts node IPs to hostnames (for SSH config matching) and writes MPI hostfile
- Coordinator pattern: Rank 0 node runs mpirun --hostfile, which SSHs to other nodes to spawn processes
- Status reporting: Runners emit standard EXO events (RunnerRunning, RunnerFailed, etc.)

MPI handles actual inter-process communication during simulation - Exo just orchestrates the spawn.

### Manual Testing
Hardware: 2x Mac Studio M2 Ultra 512GB

What we tested:

- Started Exo on both nodes, verified discovery via /state endpoint (2 nodes visible)
- Launched FLASH Sod shock-tube simulation via POST /flash/launch
- Verified hostfile generated correctly: s14 slots=1 + james21-1 slots=1
- Simulation completed 151 timesteps across both nodes
- Output files created: sod_hdf5_chk_0000, sod_hdf5_chk_0001, sod_hdf5_plt_cnt_0000

### Automated Testing

N/A

### DIFF
--- a/src/exo/master/api.py
+++ b/src/exo/master/api.py
@@ -333,7 +333,9 @@ class API:
             # stdout/stderr pipe fds. Using PIPE would cause communicate() to hang
             # because the daemon child never closes them. Use DEVNULL instead.
             is_daemonize = "--daemonize" in cmd_str
-            out_mode = asyncio.subprocess.DEVNULL if is_daemonize else asyncio.subprocess.PIPE
+            out_mode = (
+                asyncio.subprocess.DEVNULL if is_daemonize else asyncio.subprocess.PIPE
+            )
 
             if needs_shell:
                 process = await asyncio.create_subprocess_shell(
--- a/src/exo/plugins/implementations/flash/placement.py
+++ b/src/exo/plugins/implementations/flash/placement.py
@@ -120,7 +120,7 @@ def place_flash_instance(
             # Prefer private network IPs (10.x, 192.168.x) over Tailscale CGNAT (100.64-127.x)
             chosen_ip: str | None = None
             for ip in candidate_ips:
-                if ip.startswith("10.") or ip.startswith("192.168."):
+                if ip.startswith(("10.", "192.168.")):
                     chosen_ip = ip
                     break
             if chosen_ip is None and candidate_ips:


samiamjidkhan:
Updated to remove the need for SSH entirely. It now uses a built-in remote execution system called exo-rsh, where each EXO node runs a small HTTP server for remote command execution and mpirun spawns processes without SSH, keys, or passwords. To use it, simply run EXO (`uv run exo`), and launch simulations via the API! @AlexCheema

samiamjidkhan:
Updated to simplify exo-rsh by consolidating its HTTP server into EXO's existing API. Instead of running a separate server on port 52416, remote command execution now uses the /execute endpoint on EXO's main API (port 52415). Same SSH-free functionality, one less server. @AlexCheema

samiamjidkhan:
# Update: Plugin Architecture & SLURM CLI

## Summary

Added a modular plugin system for exo. New capabilities can be added by dropping code into `src/exo/plugins/implementations/` without modifying core source code. FLASH is the first plugin.

## Plugin Interface

```python
class ExoPlugin(ABC):
    name: str                                    # Plugin identifier
    version: str                                 # Semantic version
    get_command_types() -> Sequence[type]        # Commands to handle
    get_instance_type() -> type                  # Instance type to manage
    get_api_routes() -> Sequence[tuple]          # HTTP endpoints
    process_command(...) -> Sequence[Event]      # Command processing
    plan_task(...) -> Task | None                # Worker planning
    create_runner(...)                           # Runner bootstrap
```

## SLURM-Compatible CLI

```bash
# Submit job
exo sbatch -J my_sim -N 2 --hosts s14,james21 -D /tmp /path/to/flash4

# View queue
exo squeue
JOBID     NAME    NODES  STATE
bc0ba007  my_sim  2      RUNNING

# Cancel job
exo scancel bc0ba007
```

## How It Works

1. `sbatch` calls `/flash/launch` API endpoint
2. Plugin places job across nodes
3. Coordinator runs `mpirun` with `exo-rsh` (no SSH needed)
4. `exo-rsh` spawns processes via each node's `/execute` endpoint

## Files Changed

**Added:** `src/exo/plugins/` (base, registry, context) + `implementations/flash/`

**Modified:** `api.py`, `main.py`, `plan.py`, `bootstrap.py` - added plugin hooks

**Removed:** FLASH-specific code from core files

## Tested

- 1-node and 2-node distributed FLASH simulations
- sbatch, squeue, scancel commands

samiamjidkhan:
## Update: Dynamic Type Registry for Plugin System

This change introduces a `TypeRegistry` pattern that enables plugins to register their types dynamically at import time.

What changed:
- New `TypeRegistry[T]` class with `@register` decorator and `deserialize()` method
- Commands, instances, events, and tasks now use registry-based deserialization
- FLASH plugin types properly inherit from BaseCommand/BaseInstance
  
Benefits:
- True plugin isolation - plugins are fully self-contained
- No circular imports between shared types and plugins
- Clean inheritance hierarchy (`FLASHInstance` → `BaseInstance` → `TaggedModel`)
- Extensible - new plugins just decorate their types with @registry.register

AlexCheema:
## Code Review: PR #1127 - Support for FLASH

Thanks for the contribution, @samiamjidkhan. This is an ambitious PR that adds a plugin system, a FLASH MPI simulation plugin, SLURM-compatible CLI commands, and a remote shell. I've read the full diff carefully. Below is a detailed review organized by severity.

---

### Critical: Remote Code Execution via `/execute` endpoint

**`src/exo/master/api.py` - `execute()` method (lines ~989-1054 in the diff)**

This is a **remote code execution endpoint with zero authentication**. Any HTTP client that can reach port 52415 can execute arbitrary commands on the node:

```python
async def execute(self, request: ExecuteRequest) -> ExecuteResponse:
    cmd_str = " ".join(request.command)
    # ...
    if needs_shell:
        process = await asyncio.create_subprocess_shell(cmd_str, ...)
    else:
        process = await asyncio.create_subprocess_exec(*request.command, ...)
```

This endpoint:
- Accepts arbitrary commands with no auth token, API key, or mTLS
- Uses `create_subprocess_shell` for commands containing `;|&$\``  -- full shell injection surface
- Is bound to 0.0.0.0 (the default for exo's API), not localhost-only
- Accepts caller-provided environment variables, allowing `LD_PRELOAD` injection and similar attacks

This **must not** be merged without authentication. At absolute minimum: a shared secret token passed via header, or restrict the endpoint to connections from known cluster peers only. Ideally this should use the existing libp2p gossipsub transport for remote execution rather than an unauthenticated HTTP endpoint.

---

### High: `event_apply` catch-all silently drops unknown events

**`src/exo/shared/apply.py`**

```python
case _:
    # Unknown event types from plugins are ignored
    logger.debug(f"Ignoring unknown event type: {type(event).__name__}")
    return state
```

Previously, `event_apply` had an exhaustive match over the `Event` union -- if a new core event type was added without updating the match, basedpyright would flag it as a type error. The `case _` catch-all defeats this safety net entirely. If someone adds a new core event type in the future and forgets to add it here, it will be silently dropped at runtime instead of caught at typecheck time.

**Suggestion:** Keep the exhaustive match for core events. Handle plugin events with an explicit check before the match statement, e.g.:

```python
def event_apply(event: BaseEvent, state: State) -> State:
    # Plugin events don't modify core state
    if type(event).__module__.startswith("exo.plugins"):
        return state

    # Exhaustive match for core events (pyright enforces completeness)
    match event:
        case InstanceCreated(): ...
        case InstanceDeleted(): ...
        # ... all core cases, NO case _ ...
```

---

### High: Breaking serialization change on gossipsub

**`src/exo/routing/topics.py`**

```python
# Before:
return t.model_dump_json().encode("utf-8")
# After:
return t.model_dump_json(by_alias=True, serialize_as_any=True).encode("utf-8")
```

`serialize_as_any=True` changes how subclass fields are serialized -- fields from derived classes that aren't on the base type annotation will now be included. This means the wire format of messages changes. If any node in a cluster is running the old code while another runs the new code, **deserialization will break or produce unexpected results** because the old code doesn't expect these extra fields.

Questions:
1. Is `by_alias=True` actually needed? The existing codebase uses `CamelCaseModel` with `alias_generator`, but the deserializer calls `model_validate_json()` which accepts both aliases and field names by default. What breaks without this?
2. Is there a cluster upgrade strategy? Could we add a protocol version to gossipsub messages?

---

### High: Plugin system complexity vs. value

The full plugin architecture (TypeRegistry, PluginRegistry, dynamic `field_validator` on every serialization boundary, `discover_plugins()` using `pkgutil.iter_modules`) is a substantial amount of infrastructure for a single plugin. Specific concerns:

1. **Performance on the hot path**: Every `ForwarderCommand`, `ForwarderEvent`, `IndexedEvent`, `BoundInstance`, `TaskCreated`, and `InstanceCreated` deserialization now goes through a `field_validator(mode="before")` that calls `TypeRegistry.deserialize()`. This includes iterating over all registered types in the flat-format fallback path. For the core event-processing loop that handles potentially thousands of events/sec, this is a non-trivial regression.

2. **`discover_plugins()` in subprocess**: Called in `bootstrap.py` for every runner subprocess. This re-imports all plugin modules, adding startup latency to every runner process spawn.

3. **Singleton state**: `PluginRegistry` and `TypeRegistry` use module-level singletons and class-level `_instance`. This makes testing harder and introduces hidden global state.

**Suggestion:** Consider whether FLASH could be added as a built-in instance type (like `MlxRingInstance` / `MlxJacclInstance`) with a FLASH-specific command handler in the master, without the full plugin system. The plugin system could be introduced later when there's a second use case to validate the abstraction.

---

### Medium: Dummy `ModelCard` for FLASH placement

**`src/exo/plugins/implementations/flash/placement.py`**

```python
flash_model_card = ModelCard(
    model_id=ModelId(command.simulation_name),
    storage_size=Memory(in_bytes=0),
    n_layers=1,
    hidden_size=1,
    supports_tensor=False,
    tasks=[],
)
```

This creates a fake `ModelCard` with nonsensical values to satisfy the `ShardAssignments` / `PipelineShardMetadata` interface. This pollutes the model card system -- the model_id namespace now contains simulation names, and storage_size=0 / n_layers=1 are lies. If any downstream code ever relies on model card values being accurate (e.g., memory estimation, download tracking), this will break silently.

**Suggestion:** Either make `ShardAssignments` generic enough that it doesn't require a `ModelCard`, or create a separate assignment type for non-ML workloads.

---

### Medium: Module-level crash on missing `exo-rsh`

**`src/exo/plugins/implementations/flash/runner.py`**

```python
_exo_rsh_path = shutil.which("exo-rsh")
if not _exo_rsh_path:
    raise RuntimeError("exo-rsh not found in PATH - this should be installed with exo")
EXO_RSH_PATH: str = _exo_rsh_path
```

This executes at **module import time**. If `exo-rsh` isn't on `PATH` (which is likely in CI, test environments, or when exo is imported as a library), importing this module crashes. Since `discover_plugins()` imports all plugin modules, this means **any environment without `exo-rsh` installed will fail to start exo**.

**Fix:** Move this check to the runner's `main()` function where it's actually needed, or use a lazy property.

---

### Medium: Hardcoded `network_interface: str = "en0"`

**`src/exo/plugins/implementations/flash/types.py`**

```python
network_interface: str = "en0"  # Network interface for MPI (e.g., en0, eth0)
```

`en0` is macOS-specific. On Linux (common in HPC), the interface is typically `eth0`, `ens3`, `ib0` (InfiniBand), etc. This should either be auto-detected or have no default (require explicit configuration).

---

### Medium: `resolve_host` blocking DNS calls in runner

**`src/exo/plugins/implementations/flash/runner.py` - `resolve_host()`**

```python
hostname, _, _ = socket.gethostbyaddr(host)
```

`socket.gethostbyaddr()` is a blocking call that can hang for 30+ seconds on networks without reverse DNS. This is called for every host in `generate_hostfile()`. Since the runner runs in an asyncio context (via the worker), blocking calls should use `asyncio.to_thread()` or `loop.run_in_executor()`.

---

### Medium: No tests

39 files changed, 2490 additions, 0 test files. The plugin registry, type registry, dynamic deserialization, FLASH placement logic, FLASH runner, CLI commands, and exo-rsh all lack tests. The PR description says "Automated Testing: N/A".

Given the scope of changes (modifying core serialization, event processing, and type system), automated tests are essential. At minimum:
- TypeRegistry: register, deserialize (tagged format), deserialize (flat format), error cases
- PluginRegistry: discover, route commands, route instances
- FLASH placement: node selection, hostfile generation, edge cases (not enough nodes, missing IPs)
- CLI: sbatch script parsing, squeue status mapping
- Serialization roundtrip: verify that the new `serialize_as_any` format deserializes correctly

---

### Minor: `scancel` prefix matching is O(n^2)

**`src/exo/cli/scancel.py`**

```python
for length in range(4, len(normalized) + 1):
    prefix = normalized[:length]
    if prefix not in id_map:
        id_map[prefix] = full_id
```

This pre-computes every possible prefix of every ID. For small job counts it works, but the correct approach is a simple prefix match at lookup time, which is both simpler and more efficient.

---

### Minor: `squeue` status mapping questionable

```python
"RunnerReady": "COMPLETING",
```

`RunnerReady` means the runner is ready to accept work -- mapping it to SLURM's `COMPLETING` (which means the job is finishing) is semantically incorrect. `RunnerReady` should probably map to `RUNNING` or a custom state.

---

### Minor: Numerous `pyright: ignore` comments

The CLI code has 15+ `# pyright: ignore[reportAny]` comments, mostly around `argparse` return types. Consider using a typed argument parsing library (e.g., `typer` or typed `dataclass` wrappers) to avoid the type-safety escape hatches.

---

### Minor: `salloc` and `sbatch` aren't registered as console scripts

The SLURM commands are accessed via `exo sbatch`, `exo squeue`, etc. by intercepting `sys.argv[1]` in `main()`. But they're not registered as standalone console scripts in `pyproject.toml` (unlike `exo-rsh` which is). If SLURM compatibility is a goal, users might expect `sbatch` to work directly.

---

### Architecture suggestion

This PR does too much at once, making it hard to review and risky to merge. Consider splitting into:

1. **Plugin system infrastructure** -- TypeRegistry, PluginRegistry, base classes, field validators, serialization changes. This can be reviewed and tested independently.
2. **FLASH plugin** -- FLASHInstance, placement, planning, runner. Depends on (1).
3. **exo-rsh remote shell** -- The `/execute` endpoint and `exo-rsh` client. This is the highest-risk piece and needs its own security review.
4. **SLURM CLI** -- `sbatch`, `squeue`, `scancel`, `salloc`. These are independent of the other three.

---

### What's good

- The plugin concept is sound -- exo should support workloads beyond LLM inference.
- Clean separation within the FLASH plugin (placement, planning, runner, API handlers).
- SLURM CLI compatibility is a nice touch for HPC users.
- The exo-rsh approach (replacing SSH with HTTP for MPI spawning) is clever.
- Good docstrings and comments throughout.
- The `PluginContext` dataclass is a clean way to give plugins access to state without coupling them to internal API components.

Issue #952, opened 2025-12-21
[MEDIUM] Fix Local Network permissions in MacOS App


rltakashige:
**Context:**
The MacOS app requires Local Network access (in System Settings -> Privacy & Security) in order to allow devices to discover each other.

**Issue:**
This is configured correctly upon install of the app, where a pop-up requests Local Network access.
However, upon restart, _devices no longer discover each other until turning Local Network for the Exo app off and on again_ (with the app closed).

DariusSasarman:
Hi! I saw your issue and I looked it up on google to see how other people dealt with it.

I found this link, which mentions a similar issue:

https://foldr.com/foldr-support/foldr-for-macos/macos-15-sequoia-local-network-privacy/

Turns out, it's an OS-level issue. Try checking your MacOS version and see if it's lower than 15.3. If so, update your OS and the issue should be solved.

rltakashige:
Thanks for a quick response. However, we've mainly been testing on MacOS 26.2, which shouldn't have this issue. Unless it's a regression?

Evanev7:
Yup! mDNS access - which we use for local discovery - is definitely intended to require a permission to access. We've googled it quite a bit! The issue is why we seem to lose this permission on restart.
I'd ask that unless you have an actual idea of what the problem is, please don't post the first few links you find - it doesn't particularly help us move forward.

southkorea2013:
> Thanks for a quick response. However, we've mainly been testing on MacOS 26.2, which shouldn't have this issue. Unless it's a regression?

We also encountered this issue in our tests. We were testing on macOS 26.2, and every time we rebooted the machines, we had to first revoke the "Local Network permission" for EXO and then re-grant it in order to discover other nodes.

Evanev7:
We aren't doing much weird or manual with the mDNS responder. Could be something to do with how we're embedding EXO in the macOS application. iroh may fix this, since it uses it's own mdns equivalent implementation, or it may not.

tinybluedev:
## Reproducing on macOS 26.2 with Thunderbolt 5

We are seeing the same behavior on a clean install of EXO.app v1.0.67 with two M3 Ultra Mac Studios connected via dual Thunderbolt 5 cables (80 Gb/s, RDMA enabled).

### Setup
- **STRaiNGE1** & **STRaiNGE2**: Mac Studio M3 Ultra, 512GB unified memory each
- **macOS**: 26.2 (Build 25C56)
[...]

PR #1136, opened 2026-01-12
fix local network warning

samiamjidkhan:
## Motivation

Local network warning banner was showing on fresh install even though mDNS was working. The check would fail before the user had a chance to grant permission via the macOS prompt.

## Changes

- Added `hasWorkedBefore` flag persisted in UserDefaults
- Only show warning if permission previously worked but now doesn't

## Why It Works

On fresh install, the check may fail (no permission yet), but `hasWorkedBefore` is false so no warning shows. Once the user grants permission and a check succeeds, we record it. Future failures (zombie permission after restart) will show the warning since `hasWorkedBefore` is now true.

### Manual Testing
Please test a new app build with this change @AlexCheema 

### Automated Testing
N/A

### DIFF
--- a/app/EXO/EXO/ContentView.swift
+++ b/app/EXO/EXO/ContentView.swift
@@ -56,12 +56,13 @@ struct ContentView: View {
     }
 
     private var shouldShowLocalNetworkWarning: Bool {
-        // Only show warning if:
-        // 1. Local network is not working
-        // 2. EXO is running
-        // 3. We've had a successful connection before (avoids false positive on fresh install)
+        // Show warning if local network is not working and EXO is running.
+        // The checker uses a longer timeout on first launch to allow time for
+        // the permission prompt, so this correctly handles both:
+        // 1. User denied permission on first launch
+        // 2. Permission broke after restart (macOS TCC bug)
         if case .notWorking = localNetworkChecker.status {
-            return controller.status != .stopped && localNetworkChecker.hasWorkedBefore
+            return controller.status != .stopped
         }
         return false
     }
--- a/app/EXO/EXO/Services/LocalNetworkChecker.swift
+++ b/app/EXO/EXO/Services/LocalNetworkChecker.swift
@@ -5,10 +5,8 @@ import os.log
 /// Checks if the app's local network permission is actually functional.
 ///
 /// macOS local network permission can appear enabled in System Preferences but not
-/// actually work after a restart. This service detects this by creating a UDP
-/// connection to the mDNS multicast address (224.0.0.251:5353).
-///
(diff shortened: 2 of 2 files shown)


AlexCheema (dismissed):
Is this the right way to fix it?
Can't we find out why this happens?
It's easy to reproduce. Just install exo from scratch and it will happen.

AlexCheema:
Also please run `nix flake check` and `nix fmt`

JakeHillion (changes requested):
Took a look at this, this is definitely an annoying issue. There are two cases we want this local network warning to appear in:
1. The user said "no" the first time they were asked for permissions.
2. The Mac has decided to break local network access, even though permissions were previously allowed.

The first case is important too and we definitely lose it with this approach.

Could we adjust the Swift to make a local network request that would make the popup appear, then only after that has happened perform our check? Or alternatively, coordinate with the Python to ensure it's already attempted discovery before we do the check. I think that would make this check effective.

lucasajackson:
Could there be a bug causing the message to show more often?  I've granted local network access on all of my nodes, but after restarting them I run into the message.  When the message shows the nodes are unable to discover each other. I'm constantly going back and turning local network access on and off after restarting my nodes - it kind of defats the purpose of opening the application on startup.  Sometimes it just requires me restart EXO to get things working again.

samiamjidkhan:
The changes introduce a 30-second timeout on first launch to allow enough time for the system permission prompt to appear and for the user to respond, while subsequent launches use a shorter 3-second timeout to provide quicker feedback. On a fresh install, the 30-second timeout ensures the user has sufficient time to click “Allow”, preventing false warnings. If the user explicitly selects “Don't Allow”, the app still waits up to 30 seconds, after which the connection fails and a warning is correctly shown.

This approach is the most practical solution imo because **macOS does not provide any API** to detect whether the permission prompt is currently displayed or what choice the user made; the only observable signal is the connection result, so timing is used to reliably handle both scenarios.

JakeHillion:
> Could there be a bug causing the message to show more often?  I've granted local network access on all of my nodes, but after restarting them I run into the message.  When the message shows the nodes are unable to discover each other. I'm constantly going back and turning local network access on and off after restarting my nodes - it kind of defats the purpose of opening the application on startup.  Sometimes it just requires me restart EXO to get things working again.  

Hi, the general issue is tracked in #952. As we don't know how to fix it, we're adding better detection to at least inform the user that's something gone wrong in the short term. This improves that detection.

The fundamental issue is much more annoying. It's quite easy to reproduce, but we can't figure out what about our app structure is causing this permission to show as approved when it's not. Or why turning it off and on again works.

Any debugging/input on the main issue would be appreciated!

PR #1145, opened 2026-01-13
nix: convert to flake-parts

JakeHillion:
Preparing to add a flake-parts module for Rust builds. The flake-utils library doesn't support the module system needed for cleanly separating the Rust build configuration.

Converted from flake-utils to flake-parts, switching to the treefmt-nix flakeModule import pattern. The devShell and formatter outputs remain functionally equivalent.

Test plan:
- Ran `nix flake check` successfully
- Verified `nix develop` provides the same environment

### DIFF
--- a/flake.nix
+++ b/flake.nix
@@ -3,13 +3,17 @@
 
   inputs = {
     nixpkgs.url = "github:NixOS/nixpkgs/nixos-unstable";
-    flake-utils.url = "github:numtide/flake-utils";
-    # Provides Rust dev-env integration:
+
+    flake-parts = {
+      url = "github:hercules-ci/flake-parts";
+      inputs.nixpkgs-lib.follows = "nixpkgs";
+    };
+
     fenix = {
       url = "github:nix-community/fenix";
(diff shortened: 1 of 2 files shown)


JakeHillion:
Note to reviewers: I'd recommend reviewing this with whitespace disabled in GitHub's diff. Whitespace isn't important in Nix but there's a lot of indentation change.

flake.nix
```diff
@@ -3,131 +3,131 @@
...
-    # nix community cachix
     extra-trusted-public-keys = "exo.cachix.org-1:okq7hl624TBeAR3kV+g39dUFSiaZgLRkLsFBCuJ2NZI=";
     extra-substituters = "https://exo.cachix.org";
   };
 
   outputs =
-    inputs:
-    let
+    inputs@{ flake-parts, ... }:
+    flake-parts.lib.mkFlake { inherit inputs; } {
```
Evanev7: can we fully expand inputs or just use inputs.flake-parts?

PR #1149, opened 2026-01-13
chore(deps): nix lock file maintenance

JakeHillion:
Update nix flake inputs. Add a second input as Swift is currently broken
in nixpkgs on Linux for `swift-format` as we want `nix fmt` to continue
being reproducible everywhere.


AlexCheema (approved):
looks fine.

PR #1153, opened 2026-01-13
feat: add continuous batching for concurrent request processing

AlexCheema:
## Motivation

Enable the runner to process multiple concurrent inference requests efficiently. Previously, requests were processed sequentially - one had to complete before the next could start. With continuous batching, multiple requests can be processed in parallel, significantly improving throughput.

## Changes

- **New `BatchGenerationEngine`** (`batch_engine.py`): Wraps mlx_lm's `BatchGenerator` to manage continuous batching. Handles request insertion, token generation, and response routing.
- **New `distributed_sync.py`**: Implements `share_object()` for syncing batch state across distributed ranks using `mx.distributed.all_sum()`.
- **Modified `runner.py`**: Changed from blocking sequential loop to non-blocking pattern that polls for new tasks between decode steps.
- **Modified `plan.py`**: Updated to forward tasks to runners in `RunnerRunning` state (not just `RunnerReady`).
- **Modified `runner_supervisor.py`**: Fixed race condition by keeping tasks in `pending` until completion (not just acknowledgement).
- **Updated `RunnerRunning`**: Added `active_requests` field to track batch size.

## Why It Works

mlx_lm's `BatchGenerator` handles the complexity of continuous batching internally:
1. New prompts are prefilled while existing requests decode
2. KV caches are concatenated when new requests join the batch
3. All active requests generate one token per decode step

The runner now:
1. Drains all pending tasks non-blocking (`receive_nowait()`)
2. Runs one batch decode step (~1-5ms)
3. Loops back to check for new tasks

This ensures new requests can join the batch mid-generation.

### Manual Testing

Hardware: MacBook Pro M4 Max (Apple Silicon)

### Automated Testing

- Added `test_continuous_batching.py` with unit tests for the new batching logic
- Updated `test_event_ordering.py` to work with the new batch engine

### DIFF
--- a/src/exo/shared/apply.py
+++ b/src/exo/shared/apply.py
@@ -184,10 +184,19 @@ def apply_instance_created(event: InstanceCreated, state: State) -> State:
 
 
 def apply_instance_deleted(event: InstanceDeleted, state: State) -> State:
+    deleted_instance = state.instances.get(event.instance_id)
     new_instances: Mapping[InstanceId, Instance] = {
         iid: inst for iid, inst in state.instances.items() if iid != event.instance_id
     }
-    return state.model_copy(update={"instances": new_instances})
+    runner_ids_to_remove: set[RunnerId] = set()
+    if deleted_instance is not None:
+        runner_ids_to_remove = set(
+            deleted_instance.shard_assignments.runner_to_shard.keys()
+        )
+    new_runners: Mapping[RunnerId, RunnerStatus] = {
+        rid: rs for rid, rs in state.runners.items() if rid not in runner_ids_to_remove
+    }
+    return state.model_copy(update={"instances": new_instances, "runners": new_runners})
 
 
 def apply_runner_status_updated(event: RunnerStatusUpdated, state: State) -> State:
--- a/src/exo/shared/tests/test_apply/test_apply_instance_deleted.py
+++ b/src/exo/shared/tests/test_apply/test_apply_instance_deleted.py
@@ -1,0 +1,142 @@
+from exo.shared.apply import apply_instance_deleted
+from exo.shared.models.model_cards import ModelId
+from exo.shared.tests.conftest import get_pipeline_shard_metadata
+from exo.shared.types.common import NodeId
+from exo.shared.types.events import InstanceDeleted
+from exo.shared.types.state import State
+from exo.shared.types.worker.instances import InstanceId, MlxRingInstance
+from exo.shared.types.worker.runners import (
+    RunnerId,
+    RunnerReady,
+    ShardAssignments,
+)
+from exo.shared.types.worker.shards import ShardMetadata
+from exo.worker.tests.constants import (
+    INSTANCE_1_ID,
+    INSTANCE_2_ID,
+    MODEL_A_ID,
+    MODEL_B_ID,
+    NODE_A,
+    NODE_B,
+    RUNNER_1_ID,
+    RUNNER_2_ID,
+)
+
+
+def _make_instance(
+    instance_id: InstanceId,
+    model_id: ModelId,
+    node_to_runner: dict[NodeId, RunnerId],
+    runner_to_shard: dict[RunnerId, ShardMetadata],
+) -> MlxRingInstance:
+    return MlxRingInstance(
+        instance_id=instance_id,
+        shard_assignments=ShardAssignments(
+            model_id=model_id,
+            node_to_runner=node_to_runner,
+            runner_to_shard=runner_to_shard,
+        ),
+        hosts_by_node={},
+        ephemeral_port=50000,
+    )
+
+
+def test_instance_deleted_removes_runners():
+    """Deleting an instance must also remove its runner entries from state."""
+    shard = get_pipeline_shard_metadata(MODEL_A_ID, device_rank=0)
+    instance = _make_instance(
+        INSTANCE_1_ID,
+        MODEL_A_ID,
+        {NODE_A: RUNNER_1_ID},
+        {RUNNER_1_ID: shard},
+    )
+    state = State(
+        instances={INSTANCE_1_ID: instance},
+        runners={RUNNER_1_ID: RunnerReady()},
+    )
+
+    new_state = apply_instance_deleted(
+        InstanceDeleted(instance_id=INSTANCE_1_ID), state
+    )
+
+    assert INSTANCE_1_ID not in new_state.instances
+    assert RUNNER_1_ID not in new_state.runners
+
+
+def test_instance_deleted_removes_only_its_runners():
+    """Deleting one instance must not remove runners belonging to another."""
+    shard_a = get_pipeline_shard_metadata(MODEL_A_ID, device_rank=0)
+    shard_b = get_pipeline_shard_metadata(MODEL_B_ID, device_rank=0)
+    instance_1 = _make_instance(
+        INSTANCE_1_ID,
+        MODEL_A_ID,
+        {NODE_A: RUNNER_1_ID},
+        {RUNNER_1_ID: shard_a},
+    )
+    instance_2 = _make_instance(
+        INSTANCE_2_ID,
+        MODEL_B_ID,
+        {NODE_B: RUNNER_2_ID},
+        {RUNNER_2_ID: shard_b},
+    )
+    state = State(
+        instances={INSTANCE_1_ID: instance_1, INSTANCE_2_ID: instance_2},
+        runners={RUNNER_1_ID: RunnerReady(), RUNNER_2_ID: RunnerReady()},
+    )
+
+    new_state = apply_instance_deleted(
+        InstanceDeleted(instance_id=INSTANCE_1_ID), state
+    )
+
+    assert INSTANCE_1_ID not in new_state.instances
+    assert RUNNER_1_ID not in new_state.runners
+    # Instance 2 and its runner must remain
+    assert INSTANCE_2_ID in new_state.instances
+    assert RUNNER_2_ID in new_state.runners
+
+
+def test_instance_deleted_multi_node_removes_all_runners():
+    """Deleting a multi-node instance removes all of its runners."""
+    shard1 = get_pipeline_shard_metadata(MODEL_A_ID, device_rank=0, world_size=2)
+    shard2 = get_pipeline_shard_metadata(MODEL_A_ID, device_rank=1, world_size=2)
+    instance = _make_instance(
+        INSTANCE_1_ID,
+        MODEL_A_ID,
+        {NODE_A: RUNNER_1_ID, NODE_B: RUNNER_2_ID},
+        {RUNNER_1_ID: shard1, RUNNER_2_ID: shard2},
+    )
+    state = State(
+        instances={INSTANCE_1_ID: instance},
+        runners={RUNNER_1_ID: RunnerReady(), RUNNER_2_ID: RunnerReady()},
+    )
+
+    new_state = apply_instance_deleted(
+        InstanceDeleted(instance_id=INSTANCE_1_ID), state
+    )
+
+    assert INSTANCE_1_ID not in new_state.instances
+    assert RUNNER_1_ID not in new_state.runners
+    assert RUNNER_2_ID not in new_state.runners
+    assert len(new_state.runners) == 0
+
+
+def test_instance_deleted_unknown_id_is_noop_for_runners():
+    """Deleting a non-existent instance should not affect runners."""
+    shard = get_pipeline_shard_metadata(MODEL_A_ID, device_rank=0)
+    instance = _make_instance(
+        INSTANCE_1_ID,
+        MODEL_A_ID,
+        {NODE_A: RUNNER_1_ID},
+        {RUNNER_1_ID: shard},
+    )
+    unknown_id = InstanceId("99999999-9999-4999-8999-999999999999")
+    state = State(
+        instances={INSTANCE_1_ID: instance},
+        runners={RUNNER_1_ID: RunnerReady()},
+    )
+
+    new_state = apply_instance_deleted(InstanceDeleted(instance_id=unknown_id), state)
+
+    # Everything should remain untouched
+    assert INSTANCE_1_ID in new_state.instances
+    assert RUNNER_1_ID in new_state.runners


Evanev7:
its open!

src/exo/worker/engines/mlx/generator/batch_engine.py
```diff
@@ -0,0 +1,302 @@
...
+                    peak_memory_usage=Memory.from_gb(mx.get_peak_memory() / 1e9),
+                )
+
+                if raw_finish_reason == "stop":
+                    finish_reason = "stop"
+                elif raw_finish_reason == "length":
+                    finish_reason = "length"
+                else:
+                    logger.warning(f"Unknown finish_reason: {raw_finish_reason}")
+                    finish_reason = "stop"
```
Evanev7: tool_calls are a valid finish reason we need to handle - this seems like it'd be better phrased as

```suggestion
                if raw_finish_reason in get_args(FinishReason):
                    finish_reason = raw_finish_reason
                else:
                    logger.warning(f"Unknown finish_reason: {raw_finish_reason}")
                    finish_reason = "stop"
```

src/exo/worker/engines/mlx/generator/distributed_sync.py
```diff
@@ -0,0 +1,30 @@
...
+
+import pickle
+from typing import TypeVar, cast
+
+import mlx.core as mx
+
+T = TypeVar("T")
+
+
+def share_object(obj: T | None, rank: int, group: mx.distributed.Group) -> T | None:
```
Evanev7:
```suggestion
from typing import cast

import mlx.core as mx


def share_object[T](obj: T | None, rank: int, group: mx.distributed.Group) -> T | None:
```

src/exo/worker/engines/mlx/generator/distributed_sync.py
```diff
@@ -0,0 +1,30 @@
...
+    """Broadcast object from rank 0 to all ranks. Two-phase: size then data."""
+    if rank == 0:
+        if obj is None:
+            mx.eval(mx.distributed.all_sum(mx.array([0]), group=group))
+            return None
```
Evanev7: If I understand your intent correctly, this should fail hard (`assert obj is not None`), and mx_barrier should be used instead.

src/exo/worker/engines/mlx/generator/distributed_sync.py
```diff
@@ -0,0 +1,30 @@
...
+        data = mx.array(list(pickle.dumps(obj)), dtype=mx.uint8)
+        mx.eval(mx.distributed.all_sum(mx.array([data.size]), group=group))
+        mx.eval(mx.distributed.all_sum(data, group=group))
+        return obj
+    else:
+        size = int(mx.distributed.all_sum(mx.array([0]), group=group).item())
+        if size == 0:
+            return None
```
Evanev7: & then here, we should treat size=0 as an error imo

AlexCheema:
Moving back to draft. Needs some further work.

Evanev7:
ok - i had some comments, ill submit them but I didnt make it to the runner.
looks good though!

src/exo/worker/engines/mlx/generator/distributed_sync.py
Evanev7: can we use the new syntax for these

src/exo/worker/engines/mlx/generator/batch_engine.py
AlexCheema: Done — now using `get_args(FinishReason)` so all valid finish reasons (including `tool_calls`) are accepted. Unknown reasons still fall back to `"stop"` with a warning.

src/exo/worker/engines/mlx/generator/distributed_sync.py
AlexCheema: Done — switched to new Python type syntax (`def share_object[T]`), removed the `TypeVar` import.

src/exo/worker/engines/mlx/generator/distributed_sync.py
AlexCheema: Done — rank 0 now asserts `obj is not None` with a message pointing to `mx_barrier()` as the alternative for syncing without data. Return type narrowed from `T | None` to `T`.

src/exo/worker/engines/mlx/generator/distributed_sync.py
AlexCheema: Done — `size == 0` now raises `RuntimeError` instead of silently returning `None`.

src/exo/worker/engines/mlx/generator/distributed_sync.py
AlexCheema: Done — using `def share_object[T](...)` with the new PEP 695 syntax.

AlexCheema:
## Code Review — PR #1153: feat: add continuous batching for concurrent request processing

**CI Status**: All checks passing (typecheck, build on aarch64-darwin, x86_64-linux, aarch64-linux).

---

### Overview

This PR replaces the sequential one-request-at-a-time text generation path in the MLX runner with a continuous batching architecture based on `mlx_lm.BatchGenerator`. The key additions are:

- **`BatchGenerationEngine`** (`batch_engine.py`): Wraps `BatchGenerator`, manages request lifecycle (queue, insert, step, complete), handles distributed sync.
- **`distributed_sync.py`**: Two-phase `share_object()` for broadcasting data from rank 0 via `all_sum`.
- **`TimeBudget`** (`time_budget.py`): Iterator that controls generation loop duration, with distributed timing sync.
- **Rewritten `runner.py` main loop**: Switches from blocking `for task in tasks` to a two-phase loop: (1) run generation steps with `TimeBudget`, polling for new tasks mid-cycle, (2) process non-generation tasks.
- **`plan.py` / `runner_supervisor.py`**: Allow task forwarding to `RunnerRunning` runners; keep tasks in `pending` until fully complete.
- **`apply.py`**: `apply_instance_deleted` now cleans up associated runners from state.
- **Tool call & model-specific handling re-implemented**: `ToolCallTracker`, `GptOssTracker`, Kimi/GLM patches, thinking model support are all re-implemented for the batch path.

Since AlexCheema's last CTO review, the critical issues he identified appear to have been addressed. The generation loop now exists (via `_run_generation_steps` + `TimeBudget`), tasks are not marked complete prematurely, and tool calls are handled. I will focus on what remains.

---

### Critical Issues

**1. `_narrow_finish_reason` silently drops `tool_calls` finish reason**

In `runner.py` lines 112-118 (`_narrow_finish_reason`), the function narrows the finish reason to `Literal["stop", "length", "content_filter"] | None`. If `batch_engine.py:step()` returns `finish_reason="tool_calls"` (which it correctly preserves via `get_args(FinishReason)`), the `_narrow_finish_reason` function in `runner.py` silently converts it to `"stop"`. This means the `TokenChunk` emitted for a tool-call completion will report `finish_reason="stop"` instead of `"tool_calls"`, which could confuse downstream OpenAI-compatible clients expecting the correct reason.

The `ToolCallTracker.process_token()` at line 158 calls `_narrow_finish_reason(response.finish_reason)` when creating `TokenChunk` for tool call parse failures, so this is not merely theoretical — the narrowing is applied in the tool call code path. If a `ToolCallChunk` is emitted the finish reason from the batch engine is discarded, but if parsing fails the narrowed (incorrect) reason is propagated.

```python
def _narrow_finish_reason(
    raw: str | None,
) -> Literal["stop", "length", "content_filter"] | None:
    if raw is None or raw in ("stop", "length", "content_filter"):
        return raw
    return "stop"
```

Consider adding `"tool_calls"` to the accepted set, or at minimum logging when an unknown reason is narrowed.

**2. Per-request sampler parameters: "last request wins" for the entire batch**

In `batch_engine.py` lines 160-166, `sync_and_insert_pending` updates the sampler from the last insert's parameters. Since `BatchGenerator` uses a single sampler for all requests in the batch, this means if request A has `temperature=0.0` and request B has `temperature=1.0`, and they're inserted in the same batch, ALL requests (including A) will use `temperature=1.0`. This is a semantic correctness issue — users sending requests with different sampling parameters will get unexpected behavior.

```python
# Update sampler from per-request parameters (last request wins for batch)
last = inserts_to_process[-1]
self.batch_gen.sampler = make_sampler(
    temp=last.temperature if last.temperature is not None else 0.7,
    top_p=last.top_p if last.top_p is not None else 1.0,
    top_k=last.top_k if last.top_k is not None else 0,
)
```

This should at minimum be documented prominently (e.g., in the API response or logs), or ideally the batch engine should refuse to batch requests with incompatible sampling parameters.

---

### Significant Issues

**3. `KVPrefixCache` removed with no replacement**

The old code used `KVPrefixCache` for efficient handling of repeated/similar prompts. This is entirely removed with no equivalent in the batch path. For workloads with shared system prompts (common in production), this is a meaningful performance regression. The `BatchGenerator` may have its own internal caching, but this should be verified and documented.

**4. `share_object` does not guard against non-rank-0 callers passing non-None data**

In `distributed_sync.py`, the function asserts that rank 0 provides non-None data, but does not assert that non-rank-0 callers pass None. If a non-rank-0 caller accidentally passes real data, `all_sum` would silently add those bytes to the rank-0 data, corrupting the result:

```python
if rank == 0:
    assert obj is not None, "Rank 0 must provide data..."
    data = mx.array(list(pickle.dumps(obj)), dtype=mx.uint8)
    ...
else:
    size = int(mx.distributed.all_sum(mx.array([0]), group=group).item())
    ...
```

Add `assert obj is None` in the `else` branch for defensive programming.

**5. Detokenizer instance sharing across requests**

In `batch_engine.py` line 181, each `ActiveRequest` is assigned `self.tokenizer.detokenizer`. If `detokenizer` is a property that returns a shared instance (not a new copy), all concurrent requests would share the same detokenizer state, corrupting text output. If it returns a new instance each time, this is fine. Worth verifying with `StreamingDetokenizer`'s behavior.

**6. `emit_task_completion` flag is always True for non-TextGeneration tasks**

In the runner main loop, `emit_task_completion` is set to `True` at line 1130 and never set to `False` within the match/case block. This means every non-TextGeneration task (ConnectToGroup, LoadModel, StartWarmup, Shutdown) will emit both `TaskStatusUpdated(Complete)` AND `RunnerStatusUpdated` at the end of the for-loop. Looking at the old code, this matches the previous behavior, but the variable name suggests it was intended to be conditionally set to `False` for some cases. The Shutdown case in particular already transitions to `RunnerShutdown` status inside the match block, then the outer `emit_task_completion` block re-emits the same status.

**7. Error handling in `_run_generation_steps` re-raises after cleanup**

In `_run_generation_steps` (runner.py around line 460-485), when `batch_engine.step()` throws an exception, the code sends `ErrorChunk` and `TaskStatusUpdated(Complete)` for all in-flight tasks, then re-raises the exception. This will crash the runner process. In the old code, the exception was also re-raised, but only for a single request. With batching, a single exception now kills ALL in-flight requests AND the entire runner process. Consider whether the runner should attempt to recover (e.g., reset the batch engine) rather than crashing.

---

### Minor Issues

**8. `conftest.py` at repo root**

A new `conftest.py` is added at the repo root with `collect_ignore = ["tests/start_distributed_test.py"]`. This is fine functionally, but the comment in the PR description doesn't mention it. Is this meant to be temporary?

**9. AGENTS.md changes bundled into this PR**

The PR adds documentation about model storage, API testing, distributed testing, and logs to `AGENTS.md`. While useful, these are unrelated to continuous batching and would be cleaner in a separate commit/PR.

**10. `TimeBudget` distributed sync uses `all_sum` for timing**

The `TimeBudget` class uses `mx.distributed.all_sum` to aggregate timing across ranks, then divides by `group.size() * sync_frequency`. This computes an average loop time across ranks and sync periods. However, the slowest rank determines the actual throughput — using an average may cause faster ranks to overshoot their iteration count. Consider using `all_max` (if available) instead of `all_sum / count`.

**11. Hardcoded default `temp=0.7`**

The default temperature of 0.7 in `BatchGenerationEngine.__init__` (line 75) and in `sync_and_insert_pending` (line 163) should be documented or pulled from a shared constant. The old code path did not have a hardcoded default — it came from the task parameters.

**12. `_process_generation_results` uses `MpSender.send` (synchronous)**

The function signature takes `MpSender[Event]` but uses synchronous `.send()` calls. In the test harness this works via `EventCollector`, but in production the synchronous send to a multiprocessing pipe could block if the pipe buffer is full. The old code also used synchronous sends in the runner process, so this is not a regression, but worth noting.

---

### What's Good

- **Architecture is sound**: The two-phase main loop (generate, then process tasks) with inline TextGeneration handling during generation is well-designed. It ensures new requests can join the batch without breaking the decode loop while non-generation tasks (Shutdown, etc.) cause a clean exit.
- **Comprehensive test coverage**: Two new test files with 10+ test cases covering single requests, multiple concurrent requests, tool calls, staggered completions, edge cases (interrupted tool calls, mixed finish reasons, 10 simultaneous completions). The `ScriptedBatchEngine` pattern is elegant.
- **Tool call and model-specific handling re-implemented**: `ToolCallTracker`, `GptOssTracker`, Kimi token filtering, GLM tokenizer patching, and thinking model support are all present in the batch path — addressing a major gap from the earlier revision.
- **`apply_instance_deleted` cleanup**: The fix to clean up runner entries when an instance is deleted is correct, well-tested, and orthogonal to the batching work but valuable.
- **Supervisor `pending` fix**: Keeping tasks in `pending` until `TaskStatusUpdated(Complete)` prevents duplicate dispatch — this is the right fix for continuous batching where acknowledgment and completion are separated.
- **Clean separation**: `BatchGenerationEngine`, `distributed_sync`, and `TimeBudget` are cleanly separated into focused modules with clear responsibilities.

---

### Verdict

This is a substantial and well-structured PR. The core continuous batching architecture is solid, and the major gaps from the earlier revision (no generation loop, premature task completion, missing tool call handling) have been addressed. The remaining issues are: (1) the `_narrow_finish_reason` silently dropping `tool_calls`, (2) the "last request wins" sampler behavior needing documentation or guards, and (3) the removed `KVPrefixCache` as a performance consideration. None are blockers, but (1) should be fixed before merge.

rltakashige:
Closing in favour of #1642

PR #1163, opened 2026-01-15
feat: add KV prefix caching for prompt reuse

AlexCheema:
## Motivation

Avoid re-processing common prompt prefixes across requests. When multiple requests share a prefix (e.g., system prompt), reuse the cached KV state instead of recomputing.

## Changes

- Rewrote `KVPrefixCache` in `cache.py` with LRU eviction (max 10 entries)
- Integrated prefix cache into `mlx_generate()` - reuses cached KV when prefix matches
- Runner creates and passes cache instance to generation

## Why It Works

On each request, we tokenize the prompt and search for the longest matching prefix in cache. If found, we deepcopy the cached KV state, trim if needed, and only prefill the remaining tokens. After generation, we store the full KV cache for future reuse.

### Manual Testing
- TODO: Test with repeated prompts, verify faster prefill on second request

### Automated Testing
- Added 13 unit tests covering prefix matching, LRU eviction, and cache operations
- All tests pass: `uv run pytest src/exo/worker/tests/unittests/test_mlx/test_prefix_cache.py -v`

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -30,14 +30,17 @@ uv run pytest src/exo/shared/tests/test_election.py
 # Run a specific test function
 uv run pytest src/exo/shared/tests/test_election.py::test_function_name
 
-# Type checking (strict mode)
-uv run basedpyright
+# Type checking (strict mode) - MUST pass before committing
+uv run basedpyright --project pyproject.toml
 
 # Linting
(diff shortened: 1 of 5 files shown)


AlexCheema:
Closing in favor of #991. Will add tests in #991

PR #1164, opened 2026-01-15
up http timeout to 3 seconds and retry on BadStatusLine

Evanev7:
we're seeing a lot of network churn - perhaps this is a connection timing out issue? lets also re-try after a second

## testing
none yet

### DIFF
--- a/src/exo/worker/utils/net_profile.py
+++ b/src/exo/worker/utils/net_profile.py
@@ -19,7 +19,7 @@ async def check_reachability(
     """Check if a node is reachable at the given IP and verify its identity."""
 
     # TODO: use an async http client
-    def _fetch_remote_node_id(*, attempt: int = 0) -> NodeId | None:
+    def _fetch_remote_node_id(*, attempt: int = 1) -> NodeId | None:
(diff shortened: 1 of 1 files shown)


src/exo/worker/utils/net_profile.py
```diff
@@ -15,8 +18,9 @@ async def check_reachability(
 ) -> None:
     """Check if a node is reachable at the given IP and verify its identity."""
 
-    def _fetch_remote_node_id() -> NodeId | None:
-        connection = http.client.HTTPConnection(target_ip, 52415, timeout=1)
+    # TODO: use an async http client
+    def _fetch_remote_node_id(*, attempt: int = 0) -> NodeId | None:
```
JakeHillion:
```suggestion
    def _fetch_remote_node_id(*, attempt: int = 1) -> NodeId | None:
```

This is off by one at the minute

JakeHillion (approved):
Rest is good, there's an off by one error though

src/exo/worker/utils/net_profile.py
AlexCheema: fixed

PR #1167, opened 2026-01-16
feat: add Claude Messages API and OpenAI Responses API support

AlexCheema:
## Motivation

Add support for Claude Messages API and OpenAI Responses API to allow users to interact with exo using these popular API formats. This enables broader compatibility with existing tooling and SDKs that expect these API formats.

## Architecture

Adapter logic lives exclusively in the API layer (`src/exo/master/adapters/`). On the way in, each adapter converts its API-specific request type (`ChatCompletionRequest`, `ClaudeMessagesRequest`, `ResponsesRequest`) into `TextGenerationTaskParams`. On the way out, each adapter converts the `TokenChunk` stream back into its API-specific response format. Everything inside the application — commands, worker, runner, event sourcing — only sees `TextGenerationTaskParams` and `TokenChunk`. No API-specific types cross the boundary.

```
                          API layer                         │  Application internals
                                                            │
Chat Completions → [adapter] → TextGenerationTaskParams ──→ │ ──→ TextGeneration command → Runner → TokenChunk ──→ │ ──→ [adapter] → ChatCompletionResponse
Claude Messages  → [adapter] → TextGenerationTaskParams ──→ │ ──→ TextGeneration command → Runner → TokenChunk ──→ │ ──→ [adapter] → ClaudeMessagesResponse
Responses API    → [adapter] → TextGenerationTaskParams ──→ │ ──→ TextGeneration command → Runner → TokenChunk ──→ │ ──→ [adapter] → ResponsesResponse
```

### New Files
- `src/exo/shared/types/claude_api.py` - Pydantic types for Claude Messages API
- `src/exo/shared/types/openai_responses.py` - Pydantic types for OpenAI Responses API
- `src/exo/shared/types/text_generation.py` - Shared `TextGenerationTaskParams` internal type
- `src/exo/master/adapters/chat_completions.py` - Chat Completions adapter (streaming/non-streaming)
- `src/exo/master/adapters/claude.py` - Claude Messages adapter (streaming/non-streaming)
- `src/exo/master/adapters/responses.py` - OpenAI Responses adapter (streaming/non-streaming)

### Modified Files
- `src/exo/master/api.py` - Refactored to use adapters uniformly for all endpoints; extracted `_resolve_and_validate_text_model` helper to deduplicate model validation across all text endpoints; removed ad-hoc `try/except ValueError` blocks from non-streaming paths

### New Endpoints
- `POST /v1/messages` - Claude Messages API (streaming and non-streaming)
- `POST /v1/responses` - OpenAI Responses API (streaming and non-streaming)

## Why It Works

All APIs are implemented as pure conversion adapters at the edge of the application:
1. Adapter functions in `src/exo/master/adapters/` convert incoming requests to `TextGenerationTaskParams`
2. `api.py` wraps the params in a `TextGeneration` command and sends it through the existing command/event flow
3. The worker, runner, and event sourcing layers only handle `TextGenerationTaskParams` and `TokenChunk` — they have no awareness of Chat Completions, Claude, or Responses API formats
4. On response, adapter functions convert the `TokenChunk` stream back to the caller's expected format
5. Model validation is handled by a single shared helper (`_resolve_and_validate_text_model`), mirroring the existing `_validate_image_model` pattern for image endpoints

No changes to core inference logic were needed.

### Streaming Formats
- **Chat Completions**: Uses `data: {...}\n\n` with `[DONE]` terminator
- **Claude**: Uses event types `message_start`, `content_block_start`, `content_block_delta`, `content_block_stop`, `message_delta`, `message_stop`
- **OpenAI Responses**: Uses event types `response.created`, `response.in_progress`, `response.output_item.added`, `response.content_part.added`, `response.output_text.delta`, `response.output_text.done`, `response.content_part.done`, `response.output_item.done`, `response.completed`

### Manual Testing
Hardware: MacBook Pro M3 Max

**Non-streaming tests:**
```bash
# Chat Completions API
  -d '{"model": "llama-3.2-1b", "messages": [{"role": "user", "content": "Hello"}], "max_tokens": 20}'

# Claude Messages API
curl -X POST http://localhost:52415/v1/messages \
  -d '{"model": "llama-3.2-1b", "max_tokens": 50, "messages": [{"role": "user", "content": "Hello"}]}'

# OpenAI Responses API
curl -X POST http://localhost:52415/v1/responses \
  -d '{"model": "llama-3.2-1b", "input": "Hello", "max_output_tokens": 20}'
```

**Streaming tests:**
```bash
# Chat Completions API (streaming)
curl -N -X POST http://localhost:52415/v1/chat/completions \
  -d '{"model": "llama-3.2-1b", "messages": [{"role": "user", "content": "Hello"}], "stream": true, "max_tokens": 20}'

# Claude Messages API (streaming)
curl -N -X POST http://localhost:52415/v1/messages \
  -d '{"model": "llama-3.2-1b", "max_tokens": 50, "messages": [{"role": "user", "content": "Hello"}], "stream": true}'

# OpenAI Responses API (streaming)
curl -N -X POST http://localhost:52415/v1/responses \
  -d '{"model": "llama-3.2-1b", "input": "Hello", "stream": true, "max_output_tokens": 20}'
```

All endpoints tested successfully with proper response formats and streaming events.

### Automated Testing
- Tests in `src/exo/master/tests/` all pass (85 tests)
- Type checker (basedpyright) passes with 0 errors
- Linter (ruff) passes

### DIFF
--- a/bench/exo_bench.py
+++ b/bench/exo_bench.py
@@ -5,16 +5,97 @@ from __future__ import annotations
 import argparse
 import contextlib
 import http.client
+import itertools
 import json
 import os
+import sys
 import time
 from collections.abc import Callable
+from pathlib import Path
 from statistics import mean
 from typing import Any
 from urllib.parse import urlencode
 
 from loguru import logger
 from transformers import AutoTokenizer
+
+# Monkey-patch for transformers 5.x compatibility
+# Kimi's tokenization_kimi.py imports bytes_to_unicode from the old location
+# which was moved in transformers 5.0.0rc2
+try:
+    import transformers.models.gpt2.tokenization_gpt2 as gpt2_tokenization
+    from transformers.convert_slow_tokenizer import bytes_to_unicode
+
+    if not hasattr(gpt2_tokenization, "bytes_to_unicode"):
+        gpt2_tokenization.bytes_to_unicode = bytes_to_unicode  # type: ignore[attr-defined]
+except ImportError:
+    pass  # transformers < 5.0 or bytes_to_unicode not available
+
+
+def load_tokenizer_for_bench(model_id: str) -> Any:
+    """
+    Load tokenizer for benchmarking, with special handling for Kimi models.
+
+    Kimi uses a custom TikTokenTokenizer that transformers 5.x can't load via AutoTokenizer.
+    This function replicates the logic from utils_mlx.py for bench compatibility.
+    """
+    model_id_lower = model_id.lower()
+
+    if "kimi-k2" in model_id_lower:
+        import importlib.util
+        import types
+
+        from huggingface_hub import snapshot_download
+
+        # Download/get the model path
+        model_path = Path(
+            snapshot_download(
+                model_id,
+                allow_patterns=["*.json", "*.py", "*.tiktoken"],
+            )
+        )
+
+        sys.path.insert(0, str(model_path))
+
+        # Load tool_declaration_ts first (tokenization_kimi imports it with relative import)
+        tool_decl_path = model_path / "tool_declaration_ts.py"
+        if tool_decl_path.exists():
+            spec = importlib.util.spec_from_file_location(
+                "tool_declaration_ts", tool_decl_path
+            )
+            if spec and spec.loader:
+                tool_decl_module = importlib.util.module_from_spec(spec)
+                sys.modules["tool_declaration_ts"] = tool_decl_module
+                spec.loader.exec_module(tool_decl_module)
+
+        # Load tokenization_kimi with patched source (convert relative to absolute import)
+        tok_path = model_path / "tokenization_kimi.py"
+        source = tok_path.read_text()
+        source = source.replace("from .tool_declaration_ts", "from tool_declaration_ts")
+        spec = importlib.util.spec_from_file_location("tokenization_kimi", tok_path)
+        if spec:
+            tok_module = types.ModuleType("tokenization_kimi")
+            tok_module.__file__ = str(tok_path)
+            sys.modules["tokenization_kimi"] = tok_module
+            exec(compile(source, tok_path, "exec"), tok_module.__dict__)  # noqa: S102
+            TikTokenTokenizer = tok_module.TikTokenTokenizer  # noqa: N806
+        else:
+            from tokenization_kimi import TikTokenTokenizer  # type: ignore[import-not-found]  # noqa: I001
+
+        hf_tokenizer: Any = TikTokenTokenizer.from_pretrained(model_path)
+
+        # Patch encode to use internal tiktoken model directly
+        # transformers 5.x has a bug in the encode->pad path for slow tokenizers
+        def _patched_encode(text: str, **kwargs: object) -> list[int]:
+            # Pass allowed_special="all" to handle special tokens like <|im_user|>
+            return list(hf_tokenizer.model.encode(text, allowed_special="all"))
+
+        hf_tokenizer.encode = _patched_encode
+
+        return hf_tokenizer
+
+    # Default: use AutoTokenizer
+    return AutoTokenizer.from_pretrained(model_id, trust_remote_code=True)
 
 
 class ExoHttpError(RuntimeError):
@@ -24,7 +105,7 @@ class ExoHttpError(RuntimeError):
 
 
 class ExoClient:
-    def __init__(self, host: str, port: int, timeout_s: float = 600.0):
+    def __init__(self, host: str, port: int, timeout_s: float = 7200.0):
         self.host = host
         self.port = port
         self.timeout_s = timeout_s
@@ -180,14 +261,7 @@ def parse_int_list(values: list[str]) -> list[int]:
             part = part.strip()
             if part:
                 items.append(int(part))
-
-    seen: set[int] = set()
-    out: list[int] = []
-    for x in items:
-        if x not in seen:
-            out.append(x)
-            seen.add(x)
-    return out
+    return items
 
 
 def resolve_model_short_id(client: ExoClient, model_arg: str) -> tuple[str, str]:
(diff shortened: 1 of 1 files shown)


Evanev7:
do we really want to funnel everything through ChatCompletions?

AlexCheema:
> do we really want to funnel everything through ChatCompletions?

Maybe not. Would the alternative be to introduce a new intermediary type and have adapters for OpenAI completions, OpenAI responses and Claude APIs?

Evanev7:
After thinking about it a bit, the responses API is probably the thing to centralise on since its the most featureful

AlexCheema:
> After thinking about it a bit, the responses API is probably the thing to centralise on since its the most featureful

Agreed. I changed it now so we use responses as the canonical API / types

Evanev7:
Just some thoughts in it's current state - this feels like a lot of boilerplate to add features like #1181. We could perhaps improve that boilerplate with some dedicated from() and to() methods, rather than reimplementing the streams themselves. I'd like to think on this some more before we ship this - perhaps something like leo's old proxy server may be more apt in the short term.

AlexCheema:
Have you had a chance to think about this as that is the bottleneck as far as I can tell @Evanev7

Evanev7 (approved):

src/exo/worker/runner/runner.py
```diff
@@ -703,14 +706,48 @@ def parse_tool_calls(
...
         if in_tool_call:
             tool_call_text_parts.append(response.text)
+            last_buffered_response = response
+            if response.finish_reason is not None:
+                last_finish_reason = response.finish_reason
             continue
         # fallthrough
         yield response
 
+    # If the generator exhausted while inside an unclosed tool call (model
```
rltakashige: I think this doesn't happen, as there's fallthrough logic in the tool call parsing. Only thing to check is for GPT OSS, which is handled entirely separately.

rltakashige (commented):
Am happy for this to be merged ASAP if you've tested Claude and OpenCode (via Responses and ChatCompletions API individually -- just comment out the endpoint, change the config and rerun exo).

Also there'll be a few merge conflicts to sort with recent API changes.

src/exo/worker/runner/runner.py
```diff
@@ -779,7 +816,7 @@ def patch_glm_tokenizer(tokenizer: TokenizerWrapper):
 
     _func_name_regex = re.compile(r"^(.*?)<arg_key>", re.DOTALL)
     _func_arg_regex = re.compile(
-        r"<arg_key>(.*?)</arg_key>(?:\\n|\s)*<arg_value>(.*?)</arg_value>",
+        r"<arg_key>(.*?)</arg_key>(?:\n|\s)*<arg_value>(.*?)(?:</arg_value>|(?=<arg_key>)|$)",
```
rltakashige: Do we really want to be supporting malformed functions? The \n does seem like an actual bug, but the case to not handle a close token seems wrong.

src/exo/worker/engines/mlx/generator/generate.py
```diff
@@ -161,21 +155,18 @@ def eos_ids_from_tokenizer(tokenizer: TokenizerWrapper) -> list[int]:
...
-    is_bench: bool = isinstance(task, BenchChatCompletionTaskParams)
-
-    # Currently we support chat-completion tasks only.
-    logger.debug(f"task_params: {task}")
 
     if task.seed is not None:
         mx.random.seed(task.seed)
 
     # Do not use the prefix cache if we are trying to do benchmarks.
+    is_bench = (task.metadata or {}).get("bench") == "true"
```
rltakashige: Please don't have a "true" string here, also this is moved unnecessarily

src/exo/worker/engines/mlx/generator/generate.py
```diff
@@ -201,15 +192,25 @@ def mlx_generate(
...
+        top_k=task.top_k if task.top_k is not None else 0,
+    )
+
+    # Normalize stop sequences to a list
+    stop_sequences: list[str] = (
+        ([task.stop] if isinstance(task.stop, str) else task.stop)
+        if task.stop is not None
+        else []
     )
+    max_stop_len = max((len(s) for s in stop_sequences), default=0)
```
rltakashige, line 206: I don't get this

AlexCheema: We accumulate text response up to the max stop sequence length so we can do comparisons to support `stop` parameter in the api which is used e.g. here https://github.com/openai/evals/blob/4bfc1f58a821685acb40bb28ba54131740535046/evals/solvers/providers/openai/openai_solver.py#L141

src/exo/worker/engines/mlx/generator/generate.py
AlexCheema: This is consistent with how the OpenAI / claude APIs work with their metadata field.

  - https://platform.openai.com/docs/api-reference/responses                                                                                    
  - https://community.openai.com/t/whats-the-format-of-the-metadata-field-in-assistant-api/624266

src/exo/worker/runner/runner.py
AlexCheema: good question — so the main scenario here is max_tokens truncation. if the model starts a \`<tool_call>\` block and then max_tokens cuts it off before \`</tool_call>\`, the finish_reason token gets buffered inside the tool call block and never yielded. without this fix the stream just hangs forever which is obviously bad.

we're not really "supporting" malformed tool calls — we're flushing the buffered text back as regular content (not as a parsed tool call). so the client just sees the raw text + finish_reason like normal. this is consistent with how openai handles it — if you get truncated mid-tool-call you get \`finish_reason: "length"\` with partial content.

we confirmed this actually happens in practice too — tested locally with qwen3 + low max_tokens and it triggered multiple times in a single session. also saw it in production logs from jan 29.

moved this to a separate pr btw: #1344

PR #1168, opened 2026-01-16
Fix exo bench for transformers 5.x

rltakashige:
## Motivation
Prompt Sizer was broken as transformers 5.x tokenizers create BatchEncodings which are essentially a dictionary of {input_ids: []} instead of the list of input ids.

### Manual Testing
Tested that exo bench runs as expected.

### DIFF
--- a/bench/exo_bench.py
+++ b/bench/exo_bench.py
@@ -241,6 +241,9 @@ class PromptSizer:
             ids = tokenizer.apply_chat_template(
                 messages, tokenize=True, add_generation_prompt=True
             )
+            # Fix for transformers 5.x
+            if hasattr(ids, "input_ids"):
+                ids = ids.input_ids
             return int(len(ids))
 
         return count_fn


Evanev7:
is the "add tool calling" commit meant to be there?

src/exo/worker/runner/runner.py
```diff
@@ -261,6 +269,98 @@ def parse_gpt_oss(
...
+
+        if response.finish_reason is None or not pending_tool_calls:
+            yield response
+        else:
+            yield response.model_copy(
+                update={
+                    "finish_reason": "tool_calls",
+                    "tool_calls": pending_tool_calls if pending_tool_calls else None,
+                }
+            )
```
JakeHillion: This code is a little odd, we know that `pending_tool_calls` is equivalent to True, and yet we check it again in the second pass. In Rust we'd write:

```
let pending_tool_calls = if let Some(tc) = pending_tool_calls {
  tc
} else {
  yield response; // pseudocode
};
```

To early return in that case. Is there a nice way to express this in Python?

rltakashige: Sorry, based off wrong branch. This is coming later!

PR #1170, opened 2026-01-16
switch from synchronous threaded pinging to an async implementation

Evanev7:
still seeing churn in our networking - lets properly rate limit it

## changes

added an httpx client with max connections with a persistent AsyncClient

## testing

deployed on cluster, discovery VASTLY more stable (the only deleted edges were those discovered by mdns)

### DIFF
--- a/app/EXO/EXO/ContentView.swift
+++ b/app/EXO/EXO/ContentView.swift
@@ -56,6 +56,11 @@ struct ContentView: View {
     }
 
     private var shouldShowLocalNetworkWarning: Bool {
+        // Show warning if local network is not working and EXO is running.
+        // The checker uses a longer timeout on first launch to allow time for
+        // the permission prompt, so this correctly handles both:
+        // 1. User denied permission on first launch
+        // 2. Permission broke after restart (macOS TCC bug)
         if case .notWorking = localNetworkChecker.status {
             return controller.status != .stopped
         }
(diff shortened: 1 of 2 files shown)


src/exo/worker/utils/net_profile.py
```diff
@@ -13,48 +12,41 @@
...
+            r = await client.get(url)
+            if r.status_code != 200:
+                continue
 
-            # Strip quotes if present (JSON string response)
-            if body.startswith('"') and body.endswith('"') and len(body) >= 2:
-                body = body[1:-1]
+            body = r.text.strip().strip('"')
+            remote_node_id = NodeId(body) if body else None
+            break
```
JakeHillion, line 41: It seems a little odd that the "this didn't return a valid node id" case is a `break` instead of a continue. I know this is a reachability test, but this still seems a little dirty... Might be better to treat this as a failure here and rename `BAD_STATUSLINE_ATTEMPTS` to `REACHABILITY_ATTEMPTS`?

src/exo/worker/utils/net_profile.py
```diff
@@ -13,48 +12,41 @@
...
-        except OSError:
-            return None
-        except http.client.BadStatusLine:
+        except (
+            httpx.ConnectError,
+            httpx.ConnectTimeout,
+            httpx.ReadTimeout,
+            httpx.RemoteProtocolError,
+        ):
             if attempt >= BAD_STATUSLINE_ATTEMPTS:
```
JakeHillion: This condition is never hit, the `for attempt in range(...):` isn't inclusive.

JakeHillion:
Forgot to mention, the uv.lock changes are huge in this, seemingly because it changed the version of the lockfile. This isn't an issue, just curious why it happened now and not in other PRs.

src/exo/worker/utils/net_profile.py
Evanev7: the break is actually intended to be the success case here, not the failure - though youre right the body being None should continue instead

PR #1172, opened 2026-01-16
Add GitHub releases with Sparkle release notes integration

### LINKED ISSUE #1140: Show patch notes in sparkle releases
Sparkle supports a `<description>` tag in the appcast.
It supports HTML.
We should include our patch notes in each release.

See https://sparkle-project.org/documentation/publishing/

AlexCheema:
## Motivation

Closes #1140

Currently releases are uploaded to S3 for Sparkle updates but there's no GitHub Release created, and Sparkle update dialogs don't show release notes. Users have no visibility into what changed.

## Changes

- Added release workflow documentation comment at top of `build-app.yml`
- Added "Fetch release notes for Sparkle" step that converts markdown from draft GitHub release to HTML
- Added "Inject release notes into appcast" step that embeds HTML in appcast.xml with CDATA
- Added "Publish GitHub Release" step that attaches DMG and publishes the draft

## Why It Works

- Sparkle's `<description>` tag supports HTML wrapped in CDATA for rendering in update dialogs
- GitHub's markdown API (`/markdown`) converts the release notes to HTML with proper formatting
- Draft releases allow writing polished notes before the build, then the workflow publishes them automatically
- The workflow fails if no draft release exists, ensuring release notes are always provided

### Manual Testing
1. Create a draft GitHub release for a new tag with markdown release notes
2. Push the tag to trigger the workflow
3. Verify the GitHub release is published with DMG attached
4. Download appcast.xml from S3 and verify `<description><![CDATA[...]]></description>` contains HTML
5. Test Sparkle update dialog on macOS to confirm release notes appear

### Automated Testing
No automated tests added - this is CI workflow configuration.

### DIFF
--- a/src/exo/worker/utils/net_profile.py
+++ b/src/exo/worker/utils/net_profile.py
@@ -22,7 +22,6 @@ async def check_reachability(
     url = f"http://{target_ip}:52415/node_id"
 
     remote_node_id = None
-
     last_error = None
 
     for _ in range(REACHABILITY_ATTEMPTS):
@@ -40,24 +39,24 @@ async def check_reachability(
             remote_node_id = NodeId(body)
             break
 
+        # expected failure cases
         except (
-            httpx.ConnectError,
-            httpx.ConnectTimeout,
-            httpx.ReadTimeout,
-            httpx.RemoteProtocolError,
-        ) as e:
+            httpx.TimeoutException,
+            httpx.NetworkError,
+        ):
+            await anyio.sleep(1)
+
+        # other failures should be logged on last attempt
+        except httpx.HTTPError as e:
             last_error = e
             await anyio.sleep(1)
 
-    else:
-        if last_error is not None:
-            logger.warning(
-                f"connect error {type(last_error).__name__} from {target_ip} after {REACHABILITY_ATTEMPTS} attempts; treating as down"
-            )
-        else:
-            logger.warning(
-                f"malformed response from {target_ip} after {REACHABILITY_ATTEMPTS} attempts; treating as down"
-            )
+    if last_error is not None:
+        logger.warning(
+            f"connect error {type(last_error).__name__} from {target_ip} after {REACHABILITY_ATTEMPTS} attempts; treating as down"
+        )
+
+    if remote_node_id is None:
         return
 
     if remote_node_id != expected_node_id:


Evanev7:
Just checking I understand the workflow: this expects runs on a vXXX draft tag, then auto publishes the draft release once the workflow has completed?

Evanev7 (approved):
Assuming tested locally, this looks good to me

AlexCheema:
> Just checking I understand the workflow: this expects runs on a vXXX draft tag, then auto publishes the draft release once the workflow has completed?

Yes exactly. Reason being I can write up nice release notes first, then make the tag / publish.

.github/workflows/build-app.yml
```diff
@@ -304,6 +312,55 @@ jobs:
             $CHANNEL_FLAG \
             .
 
+      - name: Fetch release notes for Sparkle
+        if: github.ref_type == 'tag'
```
JakeHillion: This isn't quite strict enough, but it might be okay. We don't currently do anything with Sparkle for -alpha releases but this is still going to expect a draft release with notes.

.github/workflows/build-app.yml
```diff
@@ -304,6 +312,55 @@ jobs:
...
+          # Get release body (markdown) from draft release
+          NOTES=$(gh release view "$GITHUB_REF_NAME" --json body -q '.body' 2>/dev/null || echo "")
+
+          if [[ -n "$NOTES" ]]; then
+            # Convert markdown to HTML using GitHub's API
+            echo "$NOTES" | gh api /markdown -f text=@- -f mode=gfm > output/release_notes.html
+            echo "Converted release notes to HTML"
+          else
+            echo "WARNING: No release notes found in draft release"
+            echo "<p>See <a href=\"https://github.com/${{ github.repository }}/releases/tag/$GITHUB_REF_NAME\">release page</a> for details.</p>" > output/release_notes.html
```
JakeHillion: We should probably move this HTML creation step to be one of the first in the workflow & make it fail if someone has forgotten to create the draft with the release notes - it's quite an unusual step to have to remember.

AlexCheema: Yep, added a check for release notes and it will fail early if there aren't any.

.github/workflows/build-app.yml
AlexCheema: Changed it so release notes are optional for -alpha

AlexCheema:
@JakeHillion ptal

PR #1173, opened 2026-01-16
Return error responses for Chat Completions

rltakashige:
- Error chunks
- Use error handling in exo_bench.py

## Motivation

Return when an error occurs so that generation stops. Adding timeouts is a separate TODO for model loading and chat completions. 

## Changes

- Return HTTP exceptions as JSON responses in an OpenAI compatible format.
- Context manager for generation to catch and return error messages.
- Use error handling in exo_bench.py.

### Manual Testing
Manually tested that exo_bench returns on failures within and outside generation

### DIFF
--- a/.github/workflows/build-app.yml
+++ b/.github/workflows/build-app.yml
@@ -1,5 +1,16 @@
 name: Build EXO macOS DMG
 
+# Release workflow:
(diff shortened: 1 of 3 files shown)


JakeHillion (approved):
This makes sense to me, and I'm glad we could find a compatible format.

PR #1174, opened 2026-01-16
Fix draft release detection in build workflow

AlexCheema:
## Motivation

Fixes the draft release detection query that was broken due to `head -1` truncating multi-line JSON.

## Changes

- Changed jq query to use `first // empty` instead of piping to `head -1`
- Simplified query to match only on `.name` since draft releases don't have `tag_name` set

## Why It Works

Draft releases on GitHub have an empty `tag_name` field - the tag is only set when published. We match on `.name` (the release title) which the user sets to the version like `v1.0.63`.

## Test Plan

Tested locally:
```bash
GITHUB_REF_NAME="v1.0.63"
gh api repos/exo-explore/exo/releases --jq "[.[] | select(.draft == true) | select(.name == \"$GITHUB_REF_NAME\")] | first // empty"
# Returns the full draft release JSON
```

### DIFF
--- a/.github/workflows/build-app.yml
+++ b/.github/workflows/build-app.yml
@@ -1,5 +1,16 @@
 name: Build EXO macOS DMG
 
+# Release workflow:
+# 1. Create a draft GitHub Release with the tag name (e.g. v1.0.0) and write release notes in markdown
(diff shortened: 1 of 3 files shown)


AlexCheema:
closing as stale.

PR #1177, opened 2026-01-16
Handle model timeouts

rltakashige:
- Add eval with a timeout.
- Add fast synch flag

## Motivation

Because of the experimental FAST SYNCH flag, some models may not work. This PR catches when this occurs and allows users to specify a run without fast synch

## Changes

- Adds a flag to enable or disable fast synch  (--fast-synch and --no-fast-synch)
- Adds a heuristic timeout
- Reduces exo_bench default timeout to 10 minutes.

## Why It Works

Heuristic timeout assumes normal loading times on Mac devices (60 + model size in gb / 5: e.g. DeepSeek takes up to 120 seconds to load on tensor parallel, and timeout is set to 60 + 120 = 180s. 

We could raise this value if necessary.

### Manual Testing
Catches that GPT OSS fails to load in Tensor RDMA
Can launch with --no-fast-synch flag to launch GPT OSS.

**GPT OSS 20B**
TP with fast synch
image

TP without fast synch
image
[Note: the performance is really not great as fast synch is off]

(As a sanity check)
PP with fast synch
image

PP without fast synch
image

PP without RDMA
image

TP without RDMA
image

### DIFF
--- a/.github/workflows/build-app.yml
+++ b/.github/workflows/build-app.yml
@@ -1,5 +1,16 @@
 name: Build EXO macOS DMG
 
+# Release workflow:
+# 1. Create a draft GitHub Release with the tag name (e.g. v1.0.0) and write release notes in markdown
+# 2. Push the tag: git tag v1.0.0 && git push origin v1.0.0
+# 3. This workflow builds, signs, and notarizes the DMG
+# 4. Release notes are embedded in appcast.xml for Sparkle (rendered as markdown)
+# 5. DMG and appcast.xml are uploaded to S3
+# 6. The draft GitHub Release is published with the DMG attached
+#
+# For alpha releases (e.g. v1.0.0-alpha.1): draft release and notes are optional.
+# If no draft exists, a release is auto-created with generated notes.
+
 on:
   workflow_dispatch:
   push:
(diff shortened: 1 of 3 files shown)


src/exo/worker/engines/mlx/utils_mlx.py
```diff
@@ -252,7 +298,14 @@ def shard_and_load(
             logger.info(f"loading model from {model_path} with pipeline parallelism")
             model = pipeline_auto_parallel(model, group, shard_metadata)
 
-    mx.eval(model.parameters())
+    # Estimate timeout based on model size
+    model_size_gb = get_weights_size(shard_metadata).in_bytes / (1024**3)
+    timeout_seconds = 60 + model_size_gb / 5
```
JakeHillion: Am I reading this right that we expect the SSD to load in at 5GB/s, else we terminate the program? (as model size tends to infinity anyway)

I think this is way too optimistic, especially if someone is doing something else on their machine at the time. Honestly, 1GB/s seems a smarter choice for killing the worker. We should put more effort into not asking the worker for things it can't do and use these timeouts as a last resort IMO.

rltakashige: My reasoning was that, if you're running a huge model, you likely have a decent SSD. For any small/medium sized models, 60 seconds would load a 60GB model fine even at 1GB/s. I could make the baseline 120?

JakeHillion: > My reasoning was that, if you're running a huge model, you likely have a decent SSD. For any small/medium sized models, 60 seconds would load a 60GB model fine even at 1GB/s. I could make the baseline 120?

We've discussed recently adding an environment variable `HF_MODEL_PATH` that lets you make a bunch of models available in the HF format to EXO, and I think it's very likely someone will point that at a NAS over 10Gbps Ethernet.

Perhaps we could make the baseline an environment variable? At least that would give us a way out if the numbers we've picked are too small without cutting a new release.

rltakashige: Mm that's a good idea. Will add

PR #1178, opened 2026-01-16
Handle download plan correctly

rltakashige:
## Motivation

Downloads must all be complete before LoadModel. 

### Manual Testing
Verified that Load does not happen until after all models are downloaded

### Automated Testing
Made a failing test that I fixed


rltakashige:
This is just wrong.

PR #1179, opened 2026-01-17
Fix placement pipeline edge case

rltakashige:
## Motivation
Issue discovered by some users. See https://discord.com/channels/1262421488506769518/1305414037453930496/1461864134831505716

## Changes

Fixed the placement algorithm to handle cases such as a (97.5GB, 1.25GB, 1.25GB) and for a model with 20 layers.
This used to be given (20, 1, -1) layers.

Also addressed an issue where prompt size > prefill step size caused GPU timeout errors (more common on uneven placements). Unknown if this fixes the issue for all models and configurations (requires a bunch of E2E testing), but this is a good step in the right direction. **This patch may be unnecessary if we decide to not all gather except for sequence length 1.**

Finally addresses PipelineFirstLayer and PipelineLastLayer composition when there is only one layer assigned to a shard.

### Placement algorithm
Simple algorithm that
- Allocates floor of each fraction to all layers.
- Distributes remainders greedily
- Allocates every machine with at least one layer by stealing layers from machines with the most layers.
- Min node memory > memory of a single layer 

### Distributed patch
mlx_lm's prefill loop only evaluates cache state, not logits.
When prompt > prefill_step_size, the all_gather is never evaluated, causing GPU timeout. 

### Layer composition
Initial implementation assumed no layer composition would happen for CustomMlxLayer. The fix uses `@property` to define original layer without handling getattr.

### Manual Testing

Needs testing on lopsided setup such as a MacBook and a Mac Studio.

### Automated Testing
Added failing test cases and then resolved them. Added a test for tensor parallelism to ensure that remains unbroken after the patch.


rltakashige:
Some code duplication with https://github.com/exo-explore/exo/pull/1195 -- we'll need to figure out which solution is better.

rltakashige:
All of these changes are now merged. Closing as completed.

PR #1180, opened 2026-01-17
feat: add uncertainty visualization with token-level logprobs

AlexCheema:
## Motivation

Adds uncertainty visualization to the chat interface, allowing users to see token-level confidence scores and regenerate responses from any point in the generation. This enables users to:
- Understand model confidence at each token
- Explore alternative completions by regenerating from uncertain tokens
- Debug and analyze model behavior

### Uncertainty Visualization
- Add `TokenHeatmap` component showing token-level probability coloring
- Toggle uncertainty view per message with bar chart icon
- Display tooltip with probability, logprob, and top alternative tokens on hover

### Regenerate from Token
- Add "Regenerate from here" button in token tooltip
- Use `continue_final_message` in chat template to continue within same turn (no EOS tokens)
- Add `continue_from_prefix` flag to `ChatCompletionTaskParams`

### Request Cancellation
- Add `AbortController` to cancel in-flight requests when regenerating mid-generation
- Handle `BrokenResourceError` server-side when client disconnects gracefully

### Additional APIs
- Add Claude Messages API support (`/v1/messages`)
- Add OpenAI Responses API support (`/v1/responses`)

## Why It Works

- **Proper continuation**: Using `continue_final_message=True` instead of `add_generation_prompt=True` keeps the assistant turn open, allowing the model to continue naturally from the prefix without end-of-turn markers
- **Clean cancellation**: AbortController aborts the HTTP request, and server catches `BrokenResourceError` to avoid crashes
- **Stable hover during generation**: TokenHeatmap tracks hover by index (stable across re-renders) with longer hide delay during generation

### Manual Testing

- Send a message and verify logprobs are collected
- Enable uncertainty view and verify token coloring based on probability
- Hover over tokens to see tooltip with alternatives
- Click "Regenerate from here" on a token mid-response
- Verify the response continues naturally from that point
- Verify aborting mid-generation and regenerating works without server crash

### Automated Testing
- Added tests for Claude Messages API adapter
- Added tests for OpenAI Responses API adapter


AlexCheema:
It works:

https://github.com/user-attachments/assets/1d069d80-4207-49cc-8a16-8f09cd8eff51

src/exo/master/adapters/chat_completions.py
```diff
@@ -92,24 +111,33 @@ async def generate_chat_stream(
...
-            yield "data: [DONE]\n\n"
+            if chunk.finish_reason is not None:
+                logger.info(
+                    f"generate_chat_stream yielding [DONE] for finish_reason={chunk.finish_reason}: {command_id}"
+                )
+                yield "data: [DONE]\n\n"
+                logger.info(f"generate_chat_stream returning: {command_id}")
+                return
+    finally:
+        logger.info(f"generate_chat_stream finally block: {command_id}")
```
Evanev7: seems like debug info

src/exo/worker/engines/mlx/generator/generate.py
```diff
@@ -161,7 +216,6 @@ def mlx_generate(
         sampler=sampler,
         logits_processors=logits_processors,
         prompt_cache=caches,
-        # TODO: Dynamically change prefill step size to be the maximum possible without timing out.
```
Evanev7: we probably dont want to remove this comment

src/exo/worker/engines/mlx/utils_mlx.py
```diff
@@ -416,12 +416,26 @@ def apply_chat_template(
...
+            add_generation_prompt=False,
+            tools=task_params.tools,
+        )
+    else:
+        prompt = tokenizer.apply_chat_template(
+            formatted_messages,
+            tokenize=False,
+            add_generation_prompt=True,
+            tools=task_params.tools,
+        )
```
Evanev7: would something like
```suggestion
    prompt = tokenizer.apply_chat_template(
        formatted_messages,
        tokenize=False,
        continue_final_message = task_params.continue_from_prefix
        add_generation_prompt = not task_params.continue_from_prefix
        add_generation_prompt=True,
        tools=task_params.tools,
    )
```
be accurate?

Evanev7 (commented):
Just some thoughts

Evanev7:
seems like a cool feature, need to read through the extract logprobs function more carefuly but should be an easy merge otherwise

src/exo/worker/engines/mlx/generator/generate.py
AlexCheema: I put it back. thanks for catching that.

src/exo/worker/engines/mlx/utils_mlx.py
AlexCheema: yup this isn't needed any more. simplified

src/exo/master/adapters/chat_completions.py
AlexCheema: removed.

PR #1181, opened 2026-01-17
feat: add prefill progress bar for long prompts

AlexCheema:
## Motivation

Users processing long prompts have no visibility into when token generation will start. This feature adds a progress bar showing prefill progress, giving users real-time feedback during prompt processing.

### Backend
- Added `PrefillProgress` event type with `command_id`, `processed_tokens`, `total_tokens`
- Added `PrefillProgressResponse` type (though now using direct callback approach)
- Wired `prompt_progress_callback` through MLX's `stream_generate()`
- Progress events sent directly from callback for real-time updates (not batched)
- API generates SSE named events: `event: prefill_progress\ndata: {...}`
- Added `PrefillProgressData` dataclass and `StreamEvent` union type in API

### Dashboard
- Added `PrefillProgress` interface to store
- Updated SSE parsing to handle `event:` lines (named events)
- Created `PrefillProgressBar.svelte` with animated progress bar
- Shows "Processing prompt: X/Y tokens" with percentage
- Progress bar disappears when first token arrives

## Why It Works

MLX's `stream_generate()` accepts a `prompt_progress_callback(processed, total)` that's called after each prefill chunk. By sending events directly from this callback (rather than yielding from the generator), progress updates are sent in real-time during prefill.

Using SSE named events (`event: prefill_progress`) maintains full OpenAI/Claude API compatibility - standard clients ignore named events they don't recognize, while the exo dashboard explicitly listens for them.

### Manual Testing
- Hardware: MacBook Pro M3 Max
- Set `prefill_step_size=256` for more frequent updates
- Tested with long prompts (pasted large documents)
- Verified progress bar updates incrementally during prefill
- Confirmed progress bar disappears when generation starts
- Tested with curl - standard `data:` events still work normally

Here is it working:

https://github.com/user-attachments/assets/5cc6f075-c5b2-4a44-bb4d-9efb246bc5fe

### Automated Testing
- Type checker passes (0 errors)
- All 192 tests pass
- Dashboard builds successfully

### API Compatibility
- Named SSE events are ignored by OpenAI SDK clients
- Regular token data uses standard `data: {...}` format
- `[DONE]` sentinel works as expected

---

**Note:** `prefill_step_size` is temporarily set to 256 for testing. Should be changed back to 2048 before merging for production performance.

### DIFF
--- a/src/exo/shared/types/chunks.py
+++ b/src/exo/shared/types/chunks.py
@@ -83,4 +83,6 @@ class PrefillProgressChunk(BaseChunk):
     total_tokens: int
 
 
-GenerationChunk = TokenChunk | ImageChunk | ToolCallChunk | ErrorChunk | PrefillProgressChunk
+GenerationChunk = (
+    TokenChunk | ImageChunk | ToolCallChunk | ErrorChunk | PrefillProgressChunk
+)
--- a/bench/eval_tool_calls.py
+++ b/bench/eval_tool_calls.py
@@ -170,7 +170,10 @@ def validate_nested_args(
         return False, f"'{array_key}' is empty"
     for i, item in enumerate(arr):
         if not isinstance(item, dict):
-            return False, f"'{array_key}[{i}]' is not an object (got {type(item).__name__})"
+            return (
+                False,
+                f"'{array_key}[{i}]' is not an object (got {type(item).__name__})",
+            )
         missing = [k for k in required_item_keys if k not in item]
         if missing:
             return False, f"'{array_key}[{i}]' missing keys: {missing}"
--- a/src/exo/worker/engines/mlx/utils_mlx.py
+++ b/src/exo/worker/engines/mlx/utils_mlx.py
@@ -510,7 +510,9 @@ def apply_chat_template(
         if isinstance(original_template, str):
             patched_template = _patch_lossy_chat_template(original_template)
             if patched_template is not None:
-                logger.info("Patched lossy chat template (removed inner_type length guard)")
+                logger.info(
+                    "Patched lossy chat template (removed inner_type length guard)"
+                )
 
     prompt: str = tokenizer.apply_chat_template(
         formatted_messages,
--- a/src/exo/worker/runner/runner.py
+++ b/src/exo/worker/runner/runner.py
@@ -625,7 +625,9 @@ def parse_gpt_oss(
                 prefix = "functions."
                 if current_tool_name.startswith(prefix):
                     current_tool_name = current_tool_name[len(prefix) :]
-                logger.info(f"parse_gpt_oss yielding tool call: name={current_tool_name!r}")
+                logger.info(
+                    f"parse_gpt_oss yielding tool call: name={current_tool_name!r}"
+                )
                 yield ToolCallResponse(
                     tool_calls=[
                         ToolCallItem(


src/exo/worker/engines/mlx/generator/generate.py
```diff
@@ -216,9 +217,11 @@ def mlx_generate(
         sampler=sampler,
         logits_processors=logits_processors,
         prompt_cache=caches,
-        prefill_step_size=2048,
+        # TODO: Dynamically change prefill step size to be the maximum possible without timing out.
+        prefill_step_size=256,  # Temporarily reduced from 2048 for testing progress bar
```
AlexCheema: should we keep at 256 for usability? how much slower is this?

src/exo/master/adapters/chat_completions.py
```diff
@@ -108,11 +108,22 @@ def chunk_to_response(
...
+    """Generate Chat Completions API streaming events from StreamEvents.
+
+    Handles both TokenChunks (token generation) and PrefillProgressData (prefill progress).
+    """
     try:
-        async for chunk in chunk_stream:
+        async for event in event_stream:
+            if isinstance(event, PrefillProgressData):
+                # Send prefill progress as a named SSE event
+                progress_json = f'{{"processed":{event.processed_tokens},"total":{event.total_tokens}}}'
```
Evanev7: Should be a model_dump

src/exo/master/api.py
```diff
@@ -432,6 +437,74 @@ async def _token_chunk_stream(
...
+            if isinstance(event, TokenChunk):
+                yield event
+
+    async def _generate_chat_stream(
+        self, command_id: CommandId
+    ) -> AsyncGenerator[str, None]:
+        """Generate chat completion stream as JSON strings."""
+
+        async for event in self._stream_events(command_id):
+            if isinstance(event, PrefillProgressData):
```
Evanev7: can we make this a match case?

src/exo/worker/engines/mlx/generator/generate.py
```diff
@@ -75,7 +75,7 @@ def warmup_inference(
         max_tokens=50,
         sampler=sampler,
         prompt_cache=cache,
-        prefill_step_size=2048,
+        prefill_step_size=256,  # Temporarily reduced from 2048 for testing progress bar
```
Evanev7: bug in comment

src/exo/worker/engines/mlx/generator/generate.py
Evanev7: (weirdly this didnt reply to the original comment)

we shouldn't keep prefill step size at 256 (we also probably shouldnt assume time is linear in tokens) but we want to saturate the gpu. 256 is doable for huge models, but for say GPT-OSS (which ive heard is a good model for summarization) we're looking more like 1024 tokens with some napkin maths

Evanev7 (changes requested):
this one actually looks pretty good

AlexCheema:
## Code Review: feat: add prefill progress bar for long prompts

Overall, this is a well-designed feature with clean separation of concerns across the stack (event type -> runner callback -> API streaming -> frontend component). The architecture choices are sound. However, there is one critical issue with where the callback is wired that likely makes the feature non-functional on the current `main` branch, plus several medium-priority items.

---

### Critical: Prefill progress callback is wired to the wrong `stream_generate` call

In `generate.py`, the `on_prefill_progress` callback is passed to the `stream_generate` call at line ~346 that handles **token generation**, not the one that handles **prompt prefill**:

```python
# This stream_generate only gets `last_token` (2 tokens) with prefill_step_size=1
for completion_tokens, out in enumerate(
    stream_generate(
        model=model,
        tokenizer=tokenizer,
        prompt=last_token,        # <-- only 2 tokens!
        max_tokens=max_tokens,
        ...
        prefill_step_size=1,
        prompt_progress_callback=on_prefill_progress,  # <-- callback here
    ),
```

The actual long-prompt prefill happens earlier in `prefill()` (which internally calls `stream_generate` with `prefill_step_size=8192` and the full `prompt_tokens[:-1]`). That function already has its own `progress_callback` that logs to debug -- but it does NOT call the new `on_prefill_progress` callback.

Since `last_token` is only 2 tokens, the `on_prefill_progress` callback will fire trivially (for 2 tokens) and then generation starts. For long prompts, users will see no meaningful progress updates during the actual prefill phase.

**Fix:** The `on_prefill_progress` callback needs to be threaded into the `prefill()` function (or chained with its existing `progress_callback`). Something like:

```python
def prefill(
    model, tokenizer, sampler, prompt_tokens, cache,
    on_progress: Callable[[int, int], None] | None = None,  # <-- add parameter
) -> tuple[float, int, list[CacheSnapshot]]:
    ...
    def progress_callback(processed: int, total: int) -> None:
        # existing debug logging
        ...
        if on_progress is not None:
            on_progress(processed, total)
```

Then in `mlx_generate`:
```python
prefill_tps, prefill_tokens, ssm_snapshots_list = prefill(
    model, tokenizer, sampler, prompt_tokens[:-1], caches,
    on_progress=on_prefill_progress,
)
```

---

### Medium Issues

**1. `prefill_step_size` changed from 2048 to 1024 in `warmup_inference`**

The PR description notes: "prefill_step_size is temporarily set to 256 for testing. Should be changed back to 2048 before merging." The diff shows the warmup going from 2048 to 1024 (not 256, and not reverted to 2048). This is a behavioral change unrelated to the progress bar feature. Was this intentional? If the warmup step size matters for performance, it should be in a separate commit with justification.

**2. Claude and Responses streaming endpoints do NOT get prefill progress**

In `api.py`, only the Chat Completions streaming path uses `_stream_events()`:
```python
# Chat Completions - gets full event stream including prefill progress
generate_chat_stream(command.command_id, self._stream_events(command.command_id))
```

But Claude Messages and OpenAI Responses streaming paths use `_chunk_stream()` which filters out `PrefillProgressData`:
```python
# Claude - filters out prefill progress
generate_claude_stream(command.command_id, payload.model, self._chunk_stream(command.command_id))

# Responses - filters out prefill progress
generate_responses_stream(command.command_id, payload.model, self._chunk_stream(command.command_id))
```

If this is intentional (for API compatibility), it should be documented. If the dashboard uses the Chat Completions endpoint exclusively, this is fine -- but it would be good to add a comment explaining the asymmetry.

**3. `PrefillProgressResponse` type is unused**

`PrefillProgressResponse` is defined in `runner_response.py` but never imported or used anywhere. The PR description mentions "(though now using direct callback approach)" -- this looks like dead code from an earlier iteration. It should be removed to keep the codebase clean.

**4. `StreamEvent` type alias is unused and incomplete**

`StreamEvent = TokenChunk | PrefillProgressData` is defined in `chunks.py` but:
- It's never imported or used anywhere
- It doesn't include `ErrorChunk` or `ToolCallChunk`, which are also stream events
- The actual union used in `api.py` is `TokenChunk | ErrorChunk | ToolCallChunk | PrefillProgressData`

This should either be removed or made complete and actually used.

**5. SSE custom event type compatibility**

Using `event: prefill_progress` as a named SSE event is a reasonable choice. As the PR description correctly notes, standard clients ignore named events they don't subscribe to. However, some SSE client libraries parse `event:` lines and may include them in their event dispatch. Worth noting in an API doc or comment that this is an exo extension to the OpenAI streaming protocol.

**6. No frontend integration visible in this diff**

`PrefillProgressBar.svelte` is created and `PrefillProgress` is defined in `app.svelte.ts`, but I don't see where the component is actually rendered in the chat UI. The SSE listener that parses `event: prefill_progress` and feeds data to the component isn't shown in this diff either. Is the frontend integration in a separate PR or missing from this diff?

---

### Minor / Positive

- The `match`/`case` refactor in `chat_completions.py` is cleaner than the if/elif chain. Good improvement.
- Only rank 0 emitting progress events avoids duplicates in multi-device setups. Correct approach.
- Token count formatting with `k` suffix (e.g., "12.5k") is a nice UX touch in `PrefillProgressBar.svelte`.
- The `ValueError` catch wrapping non-streaming `collect_*` calls with `HTTPException(500)` is good defensive coding.
- The `BrokenResourceError` handling in `_apply_state` for prefill progress mirrors the existing pattern for `ChunkGenerated`. Consistent.
- Clean component design in `PrefillProgressBar.svelte` with proper Svelte 5 `$props()` and `$derived` usage.

---

### Summary

The feature design is solid and the code is well-structured, but the critical callback wiring issue means the progress bar won't actually show meaningful progress during long prompt prefill on the current codebase. Once that's fixed (threading the callback into the `prefill()` function), this will be a great UX improvement. The unused types (`PrefillProgressResponse`, `StreamEvent`) should be cleaned up, and the `prefill_step_size` change in warmup should be either reverted or justified separately.

src/exo/master/api.py
```diff
@@ -527,19 +530,23 @@ async def delete_instance(self, instance_id: InstanceId) -> DeleteInstanceRespon
...
             self._text_generation_queues[command_id], recv = channel[
-                ErrorChunk | ToolCallChunk | TokenChunk
+                TokenChunk | ErrorChunk | ToolCallChunk | PrefillProgressData
             ]()
 
             with recv as token_chunks:
                 async for chunk in token_chunks:
                     yield chunk
+                    if isinstance(chunk, PrefillProgressData):
+                        continue
```
Evanev7: this intentional? seems weird

src/exo/master/adapters/chat_completions.py
```diff
@@ -19,7 +19,12 @@
     ToolCall,
     Usage,
 )
-from exo.shared.types.chunks import ErrorChunk, TokenChunk, ToolCallChunk
+from exo.shared.types.chunks import (
+    ErrorChunk,
+    PrefillProgressData,
```
Evanev7: i think we should call it a chunk unfortunately

src/exo/master/adapters/chat_completions.py
```diff
@@ -197,6 +216,9 @@ async def collect_chat_response(
     last_usage: Usage | None = None
 
     async for chunk in chunk_stream:
+        if isinstance(chunk, PrefillProgressData):
```
Evanev7: given we have isinstance checks for every kind of chunk here, should this be a match case?

src/exo/master/api.py
Evanev7: though revisiting this function, it should probably be renamed or similar since it seems to now just be the cancellation handling for each stream. refactor needed in future i guess.

AlexCheema:
## Prefill Progress E2E Test

Prefill progress callback was not wired to the actual prefill() function. Fixed wiring and dashboard integration.

### Test Setup
- **Model**: `mlx-community/Meta-Llama-3.1-8B-Instruct-4bit` (single node, Pipeline MLX Ring)
- **Prompt**: ~3k tokens (150 repeated sections to trigger visible prefill)
- **Branch**: `alexcheema/prefill-progress` @ latest HEAD

### Results: PASS

The prefill progress bar displays correctly in the dashboard during prompt processing:

- **"Processing prompt"** label shown with token count: `2.0k / 3.0k tokens`
- **Yellow progress bar** fills proportionally (68% shown mid-prefill)
- **Percentage indicator** updates in real-time (`68%`)
- Progress bar appears in assistant message area before first token arrives
- Progress bar clears automatically when token generation begins
- Status indicator shows `PROCESSING` during prefill phase

### What was fixed
1. **Critical callback wiring**: `on_prefill_progress` was passed to the decode `stream_generate` (which only processes 2 tokens), not to `prefill()`. Rewired to thread the callback into `prefill()` via `progress_callback`.
2. **Dashboard SSE parsing**: Progress events sent as SSE comments (`: prefill_progress {...}`) so third-party OpenAI clients ignore them. Dashboard `_readStream` parses the comment format and updates `prefillProgress` state.
3. **PrefillProgressBar.svelte**: Component renders with token count formatting (`2.0k` suffix) and percentage display.

### Additional review fixes committed locally
- Removed unused `PrefillProgressResponse` and `StreamEvent` types
- Fixed incorrect docstring in `prefill()` return type
- Lowered `prefill_step_size` from 4096 to 1024 per Evan's review (better GPU saturation for large models)
- All checks pass: basedpyright 0 errors, ruff clean, 200 tests pass, nix fmt clean

Bug report ID: prefill-1181-20260218

src/exo/shared/types/chunks.py
```diff
@@ -77,3 +77,13 @@ def __repr_args__(self) -> Generator[tuple[str, Any], None, None]:
...
+
+
+class PrefillProgressChunk(TaggedModel):
+    """Data class for prefill progress events during streaming."""
+
+    processed_tokens: int
+    total_tokens: int
+
+
+StreamEvent = TokenChunk | PrefillProgressChunk
```
Evanev7: nitting once more: just use a uniform streamevent or chunk event imo. no need to separate the events as far as I can see.

Evanev7 (approved):
approving this but found another nit

AlexCheema:
## Prefill Progress Bar

\Prefill progress bar showing Processing prompt 2.0k / 3.0k tokens at 68%

src/exo/shared/types/chunks.py
rltakashige: I think it used to be implemented such that there was a good distinction, but I agree this is not necessary anymore.

Issue #1010, opened 2025-12-25
Add download and delete download button in download page


AlexCheema:
In the dashboard, there is the download page which shows all the models downloaded onto each node in your cluster.
It would be great to extend this to not just read but also write - being able to download a model onto a specific node and also delete a node on a specific node.

JakeHillion:
Closed by #1237

PR #1183, opened 2026-01-17
Feat: added delete model feature to the downloads page

saifxyzyz:
## Motivation
Ease of customization and reducing friction
This PR allows users to delete models from the downloads page

Partially solves: #1010 
## Changes

Created a new delete model function and a filter function to reflect on the deletion of the model

## Why It Works

Created a new delete model function and a filter function to reflect on the deletion of the model

## Test Plan  
Locally tested model deletion on my system:

https://github.com/user-attachments/assets/5453004d-7c59-4225-b396-4258758bf81c

### Manual Testing
Hardware: Intel i7
OS: Linux Pop_OS 24.04


src/exo/master/api.py
```diff
@@ -575,6 +577,51 @@ async def get_models(self) -> ModelList:
             ]
         )
 
+    async def delete_model(self, model_id: str) -> dict:
+        """Delete a downloaded model from disk."""
+        from exo.worker.download.download_utils import delete_model
```
AlexCheema: put import at top

src/exo/master/api.py
```diff
@@ -575,6 +577,51 @@ async def get_models(self) -> ModelList:
...
+                raise HTTPException(
+                    status_code=404, detail=f"Model {model_id} not found on disk"
+                )
+
+            await self._filter_stale_downloads()
+
+            return {"message": f"Model {model_id} deleted successfully"}
+        except Exception as e:
+            logger.error(f"Error deleting model {model_id}: {e}")
+            raise HTTPException(status_code=500, detail=str(e)) from e
```
AlexCheema: do we need the try except around this whole thing? wouldn't we want the 404 to be a 404 not wrapped in a 500?
also isn't 500 the default status code if this throws another kind of Exception anyway?

src/exo/master/api.py
```diff
@@ -575,6 +577,51 @@ async def get_models(self) -> ModelList:
...
+
+    async def _filter_stale_downloads(self):
+        """Remove download entries for models that no longer exist on disk."""
+        from exo.shared.types.state import State
+        from exo.worker.download.download_utils import ensure_models_dir
```
AlexCheema: imports at top

src/exo/master/api.py
```diff
@@ -575,6 +577,51 @@ async def get_models(self) -> ModelList:
...
+                )
+
+            await self._filter_stale_downloads()
+
+            return {"message": f"Model {model_id} deleted successfully"}
+        except Exception as e:
+            logger.error(f"Error deleting model {model_id}: {e}")
+            raise HTTPException(status_code=500, detail=str(e)) from e
+
+    async def _filter_stale_downloads(self):
```
AlexCheema: This fn seems like bit of a hack, probably because the way downloads work isn't great.

We should probably work towards reworking downloads rather than adding hacks around it like this. This also violates event sourcing principles, i.e. makes it so State isn't the left fold of events over apply.

saifxyzyz: You're right, I'm trying to find a way to make it work as it acts a little misleading without this

saifxyzyz: I got rid of the filter function, I'll try to tweak the the downloads logic to adapt to this feature, but until then i dont think this is worth merging, so will reopen when i figure out the structure so that the ui updates correctly without the hack

saifxyzyz:
Hey @AlexCheema PTAL, I created a delete model event and made the requested changes

AlexCheema (changes requested):
This currently doesn't do what I would expect.
It deletes the model only on the node running the API that you hit.

The correct way would be to route a Command to the node that you want to delete the model on, then the node would receive that Command and delete the model, then we should emit a `NodeDownloadProgress` event.

JakeHillion:
Hi, this should be covered by #1237. Thanks for the contribution and let me know if it doesn't do what you need.

PR #1188, opened 2026-01-18
Add speculative decoding support with draft models

AlexCheema:
## Summary

This PR adds support for speculative decoding using draft models to accelerate inference. Draft models are smaller, faster models that generate candidate tokens which are then verified by the main model in parallel.

### Key Features

- **Draft model configuration**: Set `draft_model` and `num_draft_tokens` when creating an instance
- **Runtime updates**: Change or clear draft model on running instances via API or dashboard
- **Parallel downloads**: Draft model downloads in parallel with main model for faster startup
- **Dashboard UI**: Draft model icon on instance cards to configure speculative decoding

### Changes

**Types & Events**
- `Instance`: Added `draft_model` and `num_draft_tokens` fields
- `SetDraftModel` task: Load/clear draft models on running instances
- `InstanceDraftModelUpdated` event: Propagate draft model changes to state
- `SetInstanceDraftModel` command: API command for runtime updates

**Worker**
- `plan.py`: Download draft models at `RunnerIdle` (parallel with main model)
- `runner.py`: Load draft model during `LoadModel` phase, handle `SetDraftModel` task
- `main.py`: Handle `SetDraftModel` task and update bound instance

**API & Master**
- `api.py`: `PUT /instance/{id}/draft_model` endpoint
- `main.py`: Handle `SetInstanceDraftModel` command
- `placement.py`: Pass draft model config when creating instances

**Dashboard**
- Draft model icon button on instance cards (both main view and chat view)
- Modal to select draft model and configure token count
- Visual indicator when draft model is active (cyan highlight)

## Test plan

- [ ] Create instance with draft model configured at launch
- [ ] Create instance without draft model, add via dashboard
- [ ] Clear draft model from running instance
- [ ] Verify draft model shows in both instances panels
- [ ] Verify parallel download of draft and main models
- [ ] Run inference with speculative decoding enabled

### DIFF
--- a/dashboard/src/lib/components/ModelPickerModal.svelte
+++ b/dashboard/src/lib/components/ModelPickerModal.svelte
@@ -806,6 +806,7 @@
                 isFavorite={favorites.has(group.id)}
                 {selectedModelId}
                 {canModelFit}
+                {getModelFitStatus}
                 onToggleExpand={() => toggleGroupExpanded(group.id)}
                 onSelectModel={handleSelect}
                 {onToggleFavorite}


AlexCheema:
Note the reason for `SetDraftModel` is that we need to be able to set the draft model for an already launched instance.

Initially I had written this so that you have to specify the draft model up front, but then I realized that the UX is not great if you need to specify this ahead of time, and what's much better is you first launch an instance then choose to add a draft model to it.

AlexCheema:
## Code Review — PR #1188: Add speculative decoding support with draft models

**CI status**: All checks passing (typecheck, Build and check on aarch64-darwin, x86_64-linux, aarch64-linux)

### Overview

This PR introduces the type-level foundations for speculative decoding with draft models. It adds `draft_model` and `num_draft_tokens` fields to `BaseInstance`, new event/command/task types (`InstanceDraftModelUpdated`, `SetInstanceDraftModel`, `DownloadDraftModel`, `SetDraftModel`), and a `load_draft_model()` utility. The master handles `SetInstanceDraftModel` commands and the apply layer propagates draft model state changes.

---

### Critical Issues

**1. Incomplete implementation — dead code everywhere**

The PR description claims changes to `worker/plan.py`, `worker/main.py`, `runner.py`, and `api.py`, but none of these files are actually modified. The result is:

- `DownloadDraftModel` task type is defined but **never emitted** by any planner or worker.
- `SetDraftModel` task type is defined but **never created** as a task by the master and **never handled** by any runner.
- `SetInstanceDraftModel` command type is handled by the master (emitting an `InstanceDraftModelUpdated` event), but there is **no API endpoint** (`PUT /instance/{id}/draft_model`) to issue this command.
- `load_draft_model()` in `utils_mlx.py` is defined but **never called** by any runner.
- The dashboard changes (draft model icon, modal) are mentioned in the description but **only a 2-line TypeScript interface change** is actually in the diff.

This means the entire feature is inert — the state can record draft model info, but nothing actually downloads, loads, or uses a draft model during inference. If this is intentional (first PR in a series), it should be documented as such.

**2. Heavily stale branch — 186 commits behind main**

The PR is based on commit `618cee52`, which is 186 commits behind `origin/main`. Several files have diverged significantly:

- `instances.py` on main now imports `MetaInstanceId` and has a `meta_instance_id` field on `BaseInstance`. The PR's diff is based on an older version without these. This will cause **merge conflicts**.
- `events.py` on main has `InstanceRetrying`, `MetaInstanceCreated`, `MetaInstanceDeleted`, `MetaInstancePlacementFailed`, `JacclSideChannelData`, `JacclSideChannelGathered` — none of which exist in the PR's base. The Event union insertion will conflict.
- `commands.py` on main has `CreateMetaInstance`, `DeleteMetaInstance`, `TaskCancelled` etc.
- `apply.py` has substantial new handlers for meta-instances and retries.

The branch needs a rebase before it can be merged.

**3. Master emits event but never creates a task for the worker**

In `master/main.py`, the `SetInstanceDraftModel` command handler immediately emits an `InstanceDraftModelUpdated` event — updating the state to record the draft model. But it **never creates a `SetDraftModel` task** for the worker to actually load/clear the draft model on the running runner process. The state says "draft model = X" but no worker ever acts on it. The task type exists but is unused.

---

### Significant Issues

**4. `DownloadDraftModel.model_id` uses `str` instead of `ModelId`**

```python
class DownloadDraftModel(BaseTask):
    model_id: str  # HuggingFace model ID
```

The codebase convention (and the `BaseInstance.draft_model` field in the same PR) uses `ModelId` (a `NewType`/`Id` subclass). Using bare `str` here is inconsistent and bypasses type safety.

Similarly, `SetDraftModel.model_id` is typed as `str | None` instead of `ModelId | None`.

**5. `load_draft_model` discards the tokenizer**

```python
def load_draft_model(model_id: ModelId) -> nn.Module:
    model_path = build_model_path(model_id)
    draft_model, _ = load_model(model_path, strict=True)
    ...
    return draft_model
```

Speculative decoding typically requires the draft model's tokenizer to align with the main model's tokenizer (or at least to verify they share the same vocabulary). Silently discarding the tokenizer may hide vocabulary mismatches that would produce incorrect draft tokens.

**6. No validation that draft model is compatible with main model**

There is no check anywhere that the draft model uses the same tokenizer/vocabulary as the main model. Running speculative decoding with mismatched vocabularies would produce garbage tokens. Even a simple model-family check would be valuable.

**7. `num_draft_tokens` default of 4 is duplicated in 4 places**

The default value `num_draft_tokens: int = 4` appears in:
- `BaseInstance`
- `PlaceInstance`
- `SetInstanceDraftModel`
- `SetDraftModel`

This should be a single constant (e.g., `DEFAULT_NUM_DRAFT_TOKENS = 4`) to avoid silent divergence if the default needs to change.

---

### Minor Issues

**8. `load_draft_model` uses f-string in logger.info (Loguru anti-pattern)**

```python
logger.info(f"Loaded draft model from {model_path}")
```

Loguru recommends `logger.info("Loaded draft model from {}", model_path)` for lazy formatting and structured logging.

**9. No `@final` decorator on new event/command types**

The existing codebase uses `@final` on newer event types (e.g., `MetaInstancePlacementFailed`, `InstanceRetrying`, `TracesMerged`). The new `InstanceDraftModelUpdated` is missing `@final`, breaking the pattern.

**10. `placement.py` passes `draft_model`/`num_draft_tokens` only for MlxJaccl and MlxRing but not as keyword arguments in the current main**

On the current main, `MlxJacclInstance` and `MlxRingInstance` don't have `draft_model`/`num_draft_tokens` fields (they inherit from `BaseInstance` which also doesn't have them on main). The diff adds these to `BaseInstance`, but the placement code must be rebased to work with the current `meta_instance_id` field that was added since.

---

### What's Good

- The type-level design follows the existing event-sourcing pattern well: command -> event -> apply.
- The `InstanceDraftModelUpdated` event and its apply function are cleanly written and follow the same immutable-state pattern as other apply functions.
- Adding draft model support at the `BaseInstance` level (rather than per-subclass) is the right architectural choice.
- The PR description is well-structured and the author's comment about runtime UX is a good design insight.

---

### Verdict

**Not ready to merge.** The PR is a skeleton — it defines types and state transitions but includes none of the actual worker/runner/API/dashboard code needed to make speculative decoding functional. The branch is 186 commits behind main and will have extensive merge conflicts across nearly all changed files. The implementation gaps (no task creation for workers, no API endpoint, no runner handling, dead `load_draft_model` function) mean the feature is completely inert even if merged.

**Recommended next steps:**
1. Rebase onto current main
2. Either scope this explicitly as "types-only foundation PR" (removing claims of worker/API/dashboard changes from the description), or include the actual worker, runner, API, and dashboard code
3. Add at least one unit test for `apply_instance_draft_model_updated`
4. Use `ModelId` consistently (not `str`) for draft model identifiers
5. Add a `DEFAULT_NUM_DRAFT_TOKENS` constant instead of duplicating `4` across types

PR #1193, opened 2026-01-18
Feat: Tool calling + thinking tokens

frenchtoasters:
## Motivation

This change is to enable tool calling within chats and produce thinking tokens. 

## Changes

- Added new API field tests for chat completion in the master component to verify thinking, tool_calls, and content handling in streamed completions.
- Added multiple unit tests for parsing logic in the worker component under test_parsing, covering chunk parsing with tools, reasoning, and minimax prefix logic.
- Added parser configuration tests for the runner component.
- Included a previously missed test that validates edge cases in chat completion message construction.
- Overall improvement of test coverage and validation for the tool calling and parsing features introduced earlier.

## Why It Works

This adds the parser system which will enable it to actually parse the `<tool>args</tool>` tokens, or ones similar, that are produced when a user tries to invoke a tool. They are parsed using the new https://github.com/exo-explore/exo/compare/main...frenchtoasters:tool_calling?expand=1#diff-411f9e51dde01e45ea0b31b98247a38037bdd293eeb5a1b8dee35e266bfdbda7R212 function and returned as structured tool calls based on the model type. We are also now creating thinking tokens in the `ChatCompletionMessage` to be parsed as well. 

### Manual Testing

Hardware: 2 M3 Mac Studio nodes
Using MCPHub and CodeCompanion I was able to invoke tools 
image
**further manual testing of a wider range of models and chats could be beneficial**

### Automated Testing

Tests were added in this commit 54afa9e3267ff953ddd93abebb4cf81cfcb3c618 which cover the changes that were added here.

### DIFF
--- a/src/exo/worker/runner/runner.py
+++ b/src/exo/worker/runner/runner.py
@@ -1,4 +1,5 @@
 import base64
+import dataclasses
 import time
 from collections.abc import Generator
 from functools import cache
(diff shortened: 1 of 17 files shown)


src/exo/worker/parsing/parsers/qwen3_moe.py
```diff
@@ -0,0 +1,11 @@
+from __future__ import annotations
+
+from .hermes import HermesReasoningParser, HermesToolParser
+
+
+class Qwen3MoEReasoningParser(HermesReasoningParser):
```
rltakashige, line 6: Is there a reason some parsers are inherited from HermesReasoningParser and others are aliases for it?

src/exo/worker/runner/runner.py
```diff
@@ -223,25 +350,63 @@ def on_model_load_timeout() -> None:
                         if isinstance(model, GptOssModel):
                             mlx_generator = parse_gpt_oss(mlx_generator)
 
-                        # TODO: Add tool call parser here
+                    parser = ChunkParser(_get_parser_config(shard_metadata.model_meta.model_id))
```
rltakashige: We load the model later, so a parser can be created at that point using the model's class rather than the model id. See how `parse_gpt_oss` is implemented

rltakashige (commented):
Thanks for this PR! However, I do have several concerns about landing something like this.

I'll take a closer look another time, but at first glance:
Why this solution over the example in mlx lm? Do we want this code inside EXO - should it be implemented as a proxy?

frenchtoasters:
> Thanks for this PR! However, I do have several concerns about landing something like this.
> 
> I'll take a closer look another time, but at first glance: Why this solution over the example in mlx lm? Do we want this code inside EXO - should it be implemented as a proxy?

Ah that is a very good point I was not aware there was parser logic in the mlx lm, I could revise things to use the upstream parser logic from there which is probably a better solution long term. 

The main thing I would like to see is the feature enabled when exo is stood up, without anything other than exo needing to be managed on the host. The exact implementation details are not something that really matter to me.

src/exo/worker/parsing/parsers/qwen3_moe.py
frenchtoasters: That was a design decision that I made in for future updates, having a file per distinct model type leaves run for if there are changes needed in the future things are already logically separated. Though based on the information you posted above about using the upstream mlx lm parsers this will probably be OBE from that.

src/exo/worker/runner/runner.py
frenchtoasters: I will take a look at that in a bit and follow the similar pattern.

Evanev7:
Heads up @rltakashige is also working on this so we may close one in favour of the other

frenchtoasters:
> Heads up @rltakashige is also working on this so we may close one in favour of the other

No issues there, I just pushed 

> Thanks for this PR! However, I do have several concerns about landing something like this.
> 
> I'll take a closer look another time, but at first glance: Why this solution over the example in mlx lm? Do we want this code inside EXO - should it be implemented as a proxy?

@rltakashige just pushed a change that moves away from standing up the parsers in exo in favor of upstream plus found the `pre-commit` checks and got those passing.

PR #1195, opened 2026-01-18
Resolve fast synch model loading hangs

rltakashige:
## Motivation

Some models hang in model loading with fast synch enabled. Handling loading by forcing each layer to be evaled seems to resolve this issue.

### Manual Testing
Fast synch works both enabled and disabled for gpt oss 20b. 
image

vs what used to be the case
image


Evanev7:
Is this closed by #1211?

rltakashige:
Completed in other PRs.

PR #1200, opened 2026-01-19
Fix several issues with placement

rltakashige:
## Motivation

Uneven placements were causing issues for some users with lopsided setups. While fixing, I ran into another issue with impossible allocation of memory. 

## Changes

- Allocate at least 1 layer per device.
- Catch overallocation of memory with an error.

### Manual Testing
Tested that GPT OSS is placed correctly.

### Automated Testing
Added breaking tests in the first commit. Resolved with new placement algorithm in the second one.


JakeHillion (approved):
Seems like an improvement, nice.

PR #1206, opened 2026-01-19
Wrap pipeline models for explicit mx.depends between cache and logits

rltakashige:
## Motivation

GPU timeouts often when prompt size > profile_step_size. It also happens for seemingly random models.

## Changes

Add mx.depends for cache on the logits.
All gather at the model level rather than the layer level, reducing the amount of data sent.

## Why It Works

mlx_lm's prefill loop only evaluates cache state, not logits.
When prompt > prefill_step_size, the all_gather is never evaluated, causing GPU timeout.

### Automated Testing
Added failing test cases and then resolved them.

### DIFF
--- a/src/exo/worker/tests/unittests/test_mlx/conftest.py
+++ b/src/exo/worker/tests/unittests/test_mlx/conftest.py
@@ -1,4 +1,8 @@
 # type: ignore
+import json
+import os
+import tempfile
+import traceback
 from dataclasses import dataclass
 from pathlib import Path
 from typing import Any, cast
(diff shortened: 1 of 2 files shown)


src/exo/worker/tests/unittests/test_mlx/conftest.py
```diff
@@ -156,14 +155,38 @@ def run_gpt_oss_tensor_parallel_device(
...
 
-    from exo.worker.engines.mlx.auto_parallel import tensor_auto_parallel
+    from exo.shared.types.api import ChatCompletionMessage
+    from exo.shared.types.memory import Memory
+    from exo.shared.types.models import ModelId, ModelMetadata
+    from exo.shared.types.tasks import ChatCompletionTaskParams
+    from exo.shared.types.worker.shards import TensorShardMetadata
+    from exo.worker.engines.mlx import Model
+    from exo.worker.engines.mlx.generator.generate import mlx_generate
+    from exo.worker.engines.mlx.utils_mlx import shard_and_load
```
Evanev7: Can be moved out of the function

Evanev7 (approved):

PR #1210, opened 2026-01-19
Fix macmon metric failures and MLX ring port selection

androolloyd:
## What changed
- Make macmon failures non-fatal so nodes stay online and continue publishing memory + identity metrics.
- Fix ephemeral port selection for MLX ring instances (avoid collapsing to 52414), preventing runner port collisions.
- Add model cards for `glm-4.7-flash-4bit` and `glm-4.7-reap-50-mxfp4`.
- Pin `mlx-lm==0.30.4` (and updated lockfile) to support `glm4_moe_lite` models.

## Why
- On m1-karo, macmon could fail even when the binary existed, which previously caused `start_polling_node_metrics` to exit and destabilize the node.
- The ephemeral port logic could resolve to 52414 for many instances, leading to runners never successfully listening/binding.

## Verification
- `pytest` via the project venv: `/Users/androolloyd/exo/.venv/bin/pytest -q` (98 passed).
- Manual: created a `glm-4.7-flash-4bit` instance and confirmed `/v1/chat/completions` works both directly and through HAProxy.


src/exo/master/placement.py
```diff
@@ -36,8 +34,7 @@
 
 
 def random_ephemeral_port() -> int:
-    port = random.randint(49153, 65535)
-    return port - 1 if port <= 52415 else 52414
+    return random.randint(49153, 65535)
```
Evanev7, line 37: ahh! original code was meant to be

```suggestion
	port = random.randint(49153, 65535)
    return port - 1 if port <= 52415 else port
```

src/exo/master/placement_utils.py
```diff
@@ -313,20 +298,17 @@ def _find_ip_prioritised(
...
+        _find_interface_name_for_ip(ip, other_node): ip for ip, _ in ips
     }
 
-    en0_ip = iface_map.get("en0")
+    en0_ip = other_iface_to_ip.get("en0")
     if en0_ip:
         return en0_ip
 
-    en1_ip = iface_map.get("en1")
+    en1_ip = other_iface_to_ip.get("en1")
```
Evanev7, line 311: This change doesnt seem to do anything

src/exo/worker/engines/mlx/utils_mlx.py
```diff
@@ -251,27 +193,28 @@ def load_mlx_items(
...
-    on_timeout: TimeoutCallback | None = None,
 ) -> tuple[nn.Module, TokenizerWrapper]:
     model_path = build_model_path(shard_metadata.model_meta.model_id)
 
-    model, _ = load_model(model_path, lazy=True, strict=False)
+    model, _ = load_model(
+        model_path,
+        lazy=True,
+        strict=False,
+    )
```
Evanev7, line 217: This will probably fail the formatter

src/exo/worker/utils/profile.py
Evanev7, line 1: All these changes should have just been made irrelevant - just landed a big pile of fixes, so you can probably drop this

pyproject.toml
```diff
@@ -8,22 +8,33 @@ dependencies = [
...
+    "textual>=5.3.0",
     "exo_pyo3_bindings", # rust bindings
     "anyio==4.11.0",
-    "mlx==0.30.1; sys_platform == 'darwin'",
-    "mlx[cpu]==0.30.1; sys_platform == 'linux'",
-    "mlx-lm @ git+https://github.com/AlexCheema/mlx-lm.git@fix-transformers-5.0.0rc2",
+    "bidict>=0.23.1",
+    "mlx>=0.30.1; sys_platform == 'darwin'",
+    "mlx[cpu]>=0.30.1; sys_platform == 'linux'",
+    "mlx-lm==0.30.4",
```
Evanev7, line 34: Updating mlx-lm should be a separate PR

AlexCheema:
Thanks for the contribution.
Unfortunately, this was bad timing - we just refactored all the macmon / information gathering code!
Here is the new code (which does correctly try except on macmon errors): https://github.com/exo-explore/exo/blob/main/src/exo/utils/info_gatherer/info_gatherer.py

Evanev7:
oh hi alex

PR #1211, opened 2026-01-19
Load layers individually

### LINKED ISSUE #1205: [BUG] Failed to shard gpt-oss-120b-MXFP4-Q8 across 2+ nodes in v1.0.63
## Describe the bug

In version 1.0.63, the `mlx-community/gpt-oss-120b-MXFP4-Q8` model fails to load correctly when distributed across 2 or more nodes. This appears to be a regression or related to the issues described in #1130, where the model fails to initialize during sharding.

## To Reproduce

1. Select model: `mlx-community/gpt-oss-120b-MXFP4-Q8`.
2. Set Sharding to "Tensor".
3. Set Instance Type to "MLX RDMA" (or "MLX Ring").
4. Select 2, 3, or 4 nodes for loading.
5. Observed the status changes in the dashboard.

## Expected behavior

The model should be partitioned and successfully loaded across all selected nodes (up to 4 nodes), as expected for this hardware configuration.

## Actual behavior

The loading process fails depending on the instance type and node count:

*   **Instance Type: MLX RDMA**
    *   **1 Node:** Successfully reaches **READY**.
    *   **2 Nodes:** Status is stuck at **LOADING** indefinitely.
    *   **3 or 4 Nodes:** Status cycles repeatedly between **PREPARING**, **LOADING**, **FAILED**, and **SHUTDOWN**.

*   **Instance Type: MLX Ring**
    *   **1 or 2 Nodes:** Successfully reaches **READY**.
    *   **3 or 4 Nodes:** Status cycles repeatedly between **PREPARING**, **LOADING**, **FAILED**, and **SHUTDOWN**.

## Environment

- **EXO Version:** 1.0.63 (commit: `5e28664c412fcb87821d7c441ae2ebee6f7d0382`)
- **Hardware:** 4x M3 Ultra Mac Studio (512GB RAM each)
- **Interconnection:** Fully connected network via 6x Thunderbolt cables
- **Sharding:** Tensor

## Additional context

Image

rltakashige:
## Motivation

Certain models hang at model loading in tensor parallel. 

Hopefully closes #1205 

## Changes

- Load layer by layer for tensor parallel sharding
- Move eval_with_timeout to auto_parallel.py to resolve circular import.

## Why It Works

The naive way to fix this is to use load model with lazy = False and then shard in tensor parallel. However, this requires the entire model to be loaded into memory.

Instead, we can load layer by layer and shard after loading. There is a very small memory footprint to this, but it is negligible.

I tried loading layer by layer after the sharding, and this allowed model loading but got stuck at warming up.

### Manual Testing
GPT OSS loads with TP and FAST SYNCH. Kimi does too.

### Automated Testing
We need to run a suite of exo_bench before merging this!

### DIFF
--- a/dashboard/src/lib/stores/app.svelte.ts
+++ b/dashboard/src/lib/stores/app.svelte.ts
@@ -71,44 +71,46 @@ export interface Instance {
 	};
 }
 
-interface RawNodeProfile {
+// Granular node state types from the new state structure
+interface RawNodeIdentity {
 	modelId?: string;
 	chipId?: string;
 	friendlyName?: string;
-	networkInterfaces?: Array<{
-		name?: string;
-		ipAddress?: string;
-		addresses?: Array<{ address?: string } | string>;
-		ipv4?: string;
-		ipv6?: string;
-		ipAddresses?: string[];
-		ips?: string[];
-	}>;
-	memory?: {
-		ramTotal?: { inBytes: number };
-		ramAvailable?: { inBytes: number };
-		swapTotal?: { inBytes: number };
-		swapAvailable?: { inBytes: number };
-	};
-	system?: {
-		gpuUsage?: number;
-		temp?: number;
-		sysPower?: number;
-	};
 }
 
-interface RawTopologyNode {
-	nodeId: string;
-	nodeProfile?: RawNodeProfile;
+interface RawMemoryUsage {
+	ramTotal?: { inBytes: number };
+	ramAvailable?: { inBytes: number };
+	swapTotal?: { inBytes: number };
+	swapAvailable?: { inBytes: number };
 }
 
-// New connection edge types from Python SocketConnection/RDMAConnection
+interface RawSystemPerformanceProfile {
+	gpuUsage?: number;
+	temp?: number;
+	sysPower?: number;
+	pcpuUsage?: number;
+	ecpuUsage?: number;
+}
+
+interface RawNetworkInterfaceInfo {
+	name?: string;
+	ipAddress?: string;
+	addresses?: Array<{ address?: string } | string>;
+	ipv4?: string;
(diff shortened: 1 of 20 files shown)


src/exo/worker/engines/mlx/auto_parallel.py
```diff
@@ -29,6 +32,40 @@
 from exo.shared.logging import logger
 from exo.shared.types.worker.shards import PipelineShardMetadata
 
+TimeoutCallback = Callable[[], None]
+
+
+def eval_with_timeout(
```
Evanev7: this is so nasty. Given we're swapping to loading layer-by-layer, could we just update the timer with frequent breaks? I guess that doesn't solve the "hangs forever" issue but man this is nasty.

src/exo/worker/engines/mlx/auto_parallel.py
```diff
@@ -272,7 +337,7 @@ def _sharded_to_all(path: str, weight: mx.array):
     if hasattr(model, "shard"):
         try:
             model.shard(group)  # type: ignore
```
Evanev7, line 339: hm - we should hit this code path somewhat often - are our issues ever here? Though i suppose its upstreams fault at that point

Evanev7 (approved):
this is once again "if it works its good". I am thinking we should move all the timeout logic to the supervisor though, since its an isolated process anyway. Surely the timeout logic should be captured as a failure to execute a workers plan or similar?

In fact following up on this - we could make each execution of a stage an async task in the runner supervisor, and cancel based on the process not responding, which lets us clean up resources better as well. Might do that tomorrow.

AlexCheema:
Testing worked with tensor parallel on:

- GPT-OSS
- Qwen30B
- MiniMax-M2

PR #1212, opened 2026-01-20
Added formatting details to  CONTRIBUTING.md

onel:
## Motivation

Added details about `nix fmt` tool and explains the language distribution per folder


onel:
I see. Those are good points. 
Let me see if I can make this more useful but most likely will close if it doesn't add any value

PR #1213, opened 2026-01-20
docs: Created docs/building-macos-app.md

onel:
## Changes

Created a guide on how to build the mac os app

### DIFF
--- a/docs/building-macos-app.md
+++ b/docs/building-macos-app.md
@@ -1,0 +1,256 @@
+# Building the macOS App
+
+This guide explains how to build the exo macOS app from source. The app provides a native macOS interface for running exo in the background with system integration.
+
+## Prerequisites
+
+Before building the app, ensure you have the following tools installed:
+
+### Required Tools
+
+- **Xcode 26.2 or later**: Download from the Mac App Store or [Apple Developer](https://developer.apple.com/xcode/)
+- **Xcode Command Line Tools**: Install with `xcode-select --install`
+- **Homebrew**: Package manager for macOS dependencies
+  ```bash
+  /bin/bash -c "$(curl -fsSL https://raw.githubusercontent.com/Homebrew/install/HEAD/install.sh)"
+  ```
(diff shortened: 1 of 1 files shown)


docs/building-macos-app.md
```diff
@@ -0,0 +1,256 @@
...
+rustup toolchain install nightly
+```
+
+### Verify Xcode Installation
+
+Ensure Xcode is properly configured:
+
+```bash
+# Select Xcode 26.2
+sudo xcode-select -s /Applications/Xcode_26.2.app
```
Evanev7, line 39: This will probably not work for typical Xcode installations, which are located in Xcode.app - can we have a different version check here?

docs/building-macos-app.md
```diff
@@ -0,0 +1,256 @@
...
+
+# Verify Metal toolchain is available
+xcrun -f metal
```
Evanev7, line 42:
```suggestion
xcrun --sdk macosx -f metal
```
had that one bite me before.

docs/building-macos-app.md
```diff
@@ -0,0 +1,256 @@
...
+
+```bash
+# Rebuild only the Python runtime
+uv run pyinstaller packaging/pyinstaller/exo.spec
+
+# Rebuild only the dashboard
+cd dashboard && npm run build && cd ..
+```
+
+## App Structure
```
Evanev7, line 135: Everything about code structure seems irrelevant to actually building the macos app - can you split this into a different file?

Evanev7 (changes requested):

JakeHillion:
This process is likely going to change enough that this will get out of date. Isn't it more appropriate to direct people to build-app.yml where they can see exactly what is happening to build the app?

onel:
If it's going to change soon, yeah definitely.

Otherwise this is more accessible. I can always keep an eye on future releases with changes in the build process and updates this doc.

Evanev7:
closing for now then - once all the nix builds land we should be good to document our builds a bit better.

PR #1221, opened 2026-01-20
swap all instances of model_id: str for model_id: ModelId

Evanev7:
This change uses the stronger typed ModelId, and introduces some convenience methods. It also cleans up some code left over from #1204.

## Changes

`model_id: str -> model_id: ModelId`
`repo_id: str -> model_id: ModelId`

Introduces methods on ModelId, in particular ModelId.normalize() to replace `/` with `--`.

This PR did introduce some circular imports, so has moved some code around to try and limit them.

## Test Plan

Tests still pass, types still check. As this is about metadata, I haven't tested inference.


src/exo/shared/models/model_cards.py
```diff
@@ -20,6 +22,44 @@ class ModelCard(CamelCaseModel):
...
+        config_data = await get_config_data(model_id)
+        num_layers = config_data.layer_count
+        mem_size_bytes = await get_safetensors_size(model_id)
+
+        mc = ModelCard(
+            model_id=ModelId(model_id),
+            storage_size=mem_size_bytes,
+            n_layers=num_layers,
+            hidden_size=config_data.hidden_size or 0,
+            # TODO: all custom models currently do not support tensor. We could add a dynamic test for this?
```
AlexCheema: our current tensor parallel implementation is per-architecture. that means we could check to see if we support the architecture for this model card (it's in `config.json`)

src/exo/shared/models/model_cards.py
```diff
@@ -308,3 +348,116 @@ class ModelCard(CamelCaseModel):
...
+                return layer_count
+
+        raise ValueError(
+            f"No layer count found in config.json: {self.model_dump_json()}"
+        )
+
+
+async def get_config_data(model_id: ModelId) -> ConfigData:
+    """Downloads and parses config.json for a model."""
+    target_dir = (await ensure_models_dir()) / str(model_id).replace("/", "--")
```
AlexCheema: maybe we can make`str(model_id).replace("/", "--")`  a function or property on `ModelId` since it's used a lot

src/exo/shared/models/model_cards.py
```diff
@@ -308,3 +348,116 @@ class ModelCard(CamelCaseModel):
...
+        (card for card in MODEL_CARDS.values() if card.model_id == model_id),
+        None,
+    )
+
+    return ModelCard(
+        model_id=model_id,
+        storage_size=mem_size_bytes,
+        n_layers=num_layers,
+        hidden_size=config_data.hidden_size or 0,
+        # TODO: all custom models currently do not support tensor. We could add a dynamic test for this?
```
AlexCheema: same as above

src/exo/shared/models/model_cards.py
Evanev7: oh - already did that, just didnt use it here

src/exo/shared/models/model_cards.py
Evanev7: ill follow up with this since its a feature we currently don't have.

AlexCheema: ok cool

AlexCheema (approved):

PR #1222, opened 2026-01-20
Replace LaunchDaemon with dynamic Thunderbolt Bridge loop detection

AlexCheema:
## Motivation

The previous approach installed a LaunchDaemon plist that ran periodically to disable Thunderbolt Bridge. This required full admin privileges upfront and ran regardless of whether a problematic loop existed.

This change replaces that with dynamic detection - only prompting the user when an actual TB bridge loop with 3+ machines is detected, and using fine-grained SCPreferences authorization instead of full admin.

## Changes

**Backend (Python):**
- Added `ThunderboltBridgeStatus` model to track bridge enabled/exists state per node
- Added `node_thunderbolt_bridge` and `thunderbolt_bridge_cycles` fields to State
- Added `get_thunderbolt_bridge_cycles()` method to Topology class
- **Robust TB bridge detection:**
  - Finds bridge network services from `-listnetworkserviceorder` (not `-listallhardwareports` which can miss bridges)
  - Checks each bridge's member interfaces via `ifconfig` to verify it contains Thunderbolt interfaces
  - Handles varying service names (e.g., "TB Bridge", "Thunderbolt Bridge", "Bridge (bridge0)")
  - Includes `service_name` in status for correct disable commands
  - Added warning logs for all error cases in detection
- Updated `apply.py` to handle the new event type and recompute cycles on node timeout

**Swift App:**
- New `ThunderboltBridgeService` that monitors for cycles from cluster state
- Shows NSAlert when a cycle with >2 machines is detected
- Uses `SCPreferencesCreateWithAuthorization` with `system.services.systemconfiguration.network` right for targeted permissions
- **Auto-cleanup of legacy LaunchDaemon:** On app startup, checks for and removes old plist/scripts (non-fatal if user cancels)
- **Periodic local network checking:** Re-checks every 10s so the warning disappears when user grants permission
- **Fixed ClusterState model:** Updated to decode new granular state fields (`nodeIdentities`, `nodeMemory`, `nodeSystem`, `nodeThunderboltBridge`) with computed `nodeProfiles` property for backwards compatibility
- **Fixed Topology model:** Updated to match actual JSON structure where `nodes` is an array of strings (not objects) and `connections` is a nested map (not flat array)
- Cleaned up `NetworkSetupHelper` by removing daemon installation code (now only handles uninstall)

**Dashboard:**
- Added yellow warning badge on topology when TB bridge cycle detected
- On hover: highlights affected nodes in yellow on the topology graph
- Shows which machines are in the cycle with friendly names
- Provides copy-paste terminal command with the correct service name:
  ```
  sudo networksetup -setnetworkserviceenabled "<service-name>" off
  ```
- Warning appears in all topology views (full, welcome, and minimized chat sidebar)
- **Debug mode:** Shows "TB:ON" or "TB:OFF" status next to each node in the topology

## Why It Works

- Cycle detection happens on the backend where we have full topology information
- Only cycles with 3+ machines are flagged (2-node connections are fine)
- TB bridge detection is robust:
  - Uses `-listnetworkserviceorder` to find bridges (works on all machines tested)
  - Verifies bridge membership via `ifconfig` to confirm Thunderbolt interfaces
  - Handles different service names across machines
- The Swift app reacts to detected cycles and prompts the user once per cycle
- The dashboard provides visual feedback and actionable instructions
- `SCPreferencesCreateWithAuthorization` provides the minimal permissions needed to modify network service state
- Legacy LaunchDaemon is automatically cleaned up on first launch with this version

### Manual Testing
Here EXO detected a TB bridge cycle:

#### Dashboard:
Screenshot 2026-01-21 at 10 07 30 PM

#### Hovering the warning:
Screenshot 2026-01-21 at 16 30 57

#### macOS app warning popup:
Screenshot 2026-01-21 at 16 29 08

### Which then asks for the user's password:
Screenshot 2026-01-21 at 16 29 28

Which when entered, successfully disables bridge and no longer shows the warning on dashboard.

#### When it fails it shows the error message:
Screenshot 2026-01-21 at 14 45 38

### Automated Testing
- Type checker: 0 errors (`uv run basedpyright`)
- Linter: All checks passed (`uv run ruff check`)
- Tests: 118 passed (`uv run pytest`)
- Dashboard: Builds successfully (`npm run build`)

### DIFF
--- a/app/EXO/EXO/ContentView.swift
+++ b/app/EXO/EXO/ContentView.swift
@@ -25,6 +25,8 @@ struct ContentView: View {
     @State private var bugReportMessage: String?
     @State private var uninstallInProgress = false
     @State private var pendingNamespace: String = ""
+    @State private var pendingHFToken: String = ""
+    @State private var pendingEnableImageModels = false
 
     var body: some View {
         VStack(alignment: .leading, spacing: 12) {
@@ -303,6 +305,49 @@ struct ContentView: View {
                             .font(.caption2)
                             .disabled(pendingNamespace == controller.customNamespace)
                         }
+                    }
+                    VStack(alignment: .leading, spacing: 4) {
+                        Text("HuggingFace Token")
+                            .font(.caption2)
+                            .foregroundColor(.secondary)
+                        HStack {
+                            SecureField("optional", text: $pendingHFToken)
+                                .textFieldStyle(.roundedBorder)
+                                .font(.caption2)
+                                .onAppear {
+                                    pendingHFToken = controller.hfToken
+                                }
+                            Button("Save & Restart") {
+                                controller.hfToken = pendingHFToken
+                                if controller.status == .running || controller.status == .starting {
+                                    controller.restart()
+                                }
+                            }
+                            .font(.caption2)
+                            .disabled(pendingHFToken == controller.hfToken)
+                        }
+                    }
+                    Divider()
+                    HStack {
+                        Toggle(
+                            "Enable Image Models (experimental)", isOn: $pendingEnableImageModels
+                        )
+                        .toggleStyle(.switch)
+                        .font(.caption2)
+                        .onAppear {
+                            pendingEnableImageModels = controller.enableImageModels
+                        }
+
+                        Spacer()
+
+                        Button("Save & Restart") {
+                            controller.enableImageModels = pendingEnableImageModels
+                            if controller.status == .running || controller.status == .starting {
+                                controller.restart()
+                            }
+                        }
+                        .font(.caption2)
+                        .disabled(pendingEnableImageModels == controller.enableImageModels)
                     }
                     HoverButton(title: "Check for Updates", small: true) {
                         updater.checkForUpdates()
--- a/app/EXO/EXO/ExoProcessController.swift
+++ b/app/EXO/EXO/ExoProcessController.swift
@@ -3,6 +3,8 @@ import Combine
 import Foundation
 
 private let customNamespaceKey = "EXOCustomNamespace"
+private let hfTokenKey = "EXOHFToken"
+private let enableImageModelsKey = "EXOEnableImageModels"
 
 @MainActor
 final class ExoProcessController: ObservableObject {
@@ -35,6 +37,22 @@ final class ExoProcessController: ObservableObject {
     {
         didSet {
             UserDefaults.standard.set(customNamespace, forKey: customNamespaceKey)
+        }
+    }
+    @Published var hfToken: String = {
+        return UserDefaults.standard.string(forKey: hfTokenKey) ?? ""
+    }()
+    {
+        didSet {
+            UserDefaults.standard.set(hfToken, forKey: hfTokenKey)
+        }
+    }
+    @Published var enableImageModels: Bool = {
+        return UserDefaults.standard.bool(forKey: enableImageModelsKey)
+    }()
+    {
+        didSet {
+            UserDefaults.standard.set(enableImageModels, forKey: enableImageModelsKey)
         }
     }
 
@@ -191,6 +209,12 @@ final class ExoProcessController: ObservableObject {
         var environment = ProcessInfo.processInfo.environment
         environment["EXO_RUNTIME_DIR"] = runtimeURL.path
         environment["EXO_LIBP2P_NAMESPACE"] = computeNamespace()
+        if !hfToken.isEmpty {
+            environment["HF_TOKEN"] = hfToken
+        }
+        if enableImageModels {
+            environment["EXO_ENABLE_IMAGE_MODELS"] = "true"
+        }
 
         var paths: [String] = []
         if let existing = environment["PATH"], !existing.isEmpty {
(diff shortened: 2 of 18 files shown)


src/exo/shared/topology.py
```diff
@@ -234,3 +235,31 @@ def is_thunderbolt_cycle(self, cycle: Cycle) -> bool:
...
+        Only returns cycles with >2 nodes (3+ machines in a loop), as cycles with
+        2 or fewer nodes don't cause the broadcast storm problem.
+        """
+        tb_cycles = self.get_cycles_tb()
+        result: list[list[NodeId]] = []
+
+        for cycle in tb_cycles:
+            node_ids = list(cycle.node_ids)
+            # Only consider cycles with more than 2 nodes
+            if len(node_ids) <= 2:
```
rltakashige: Does it packet storm if two cables are connected between the macs?

src/exo/shared/apply.py
```diff
@@ -312,6 +325,16 @@ def apply_node_gathered_info(event: NodeGatheredInfo, state: State) -> State:
...
             topology.replace_all_out_rdma_connections(event.node_id, as_rdma_conns)
+        case ThunderboltBridgeInfo():
+            new_tb_bridge: dict[NodeId, ThunderboltBridgeStatus] = {
+                **state.node_thunderbolt_bridge,
+                event.node_id: info.status,
+            }
+            update["node_thunderbolt_bridge"] = new_tb_bridge
+            # Recompute cycles with updated bridge status
+            update["thunderbolt_bridge_cycles"] = (
+                topology.get_thunderbolt_bridge_cycles(new_tb_bridge)
```
rltakashige: Not a huge fan of storing this if we're just recomputing anyway, but this feels a bit nitpicky for a review on a mobile phone

src/exo/utils/info_gatherer/info_gatherer.py
```diff
@@ -34,6 +35,142 @@
 IS_DARWIN = sys.platform == "darwin"
 
 
+async def _get_thunderbolt_devices() -> set[str] | None:
+    """Get Thunderbolt interface device names (e.g., en2, en3) from hardware ports.
+
+    Returns None if the networksetup command fails.
+    """
+    result = await anyio.run_process(
+        ["networksetup", "-listallhardwareports"],
```
rltakashige, line 44: Pretty sure we do this already?

rltakashige (commented):
I think this is fine, although I've left some comments. Will approve once you've looked at the comments, as you're testing this now.

@Evanev7 might want to check the additions to state.

src/exo/shared/topology.py
AlexCheema: nope

src/exo/utils/info_gatherer/info_gatherer.py
AlexCheema: I guess, but it seems to be a cheap operation.
`time networksetup -listallhardwareports  0.01s user 0.00s system 55% cpu 0.018 total`

probably just reading some memory somewhere

src/exo/master/tests/test_placement_utils.py
```diff
@@ -24,6 +23,35 @@
 from exo.shared.types.worker.shards import Sharding
 
 
+def _get_hosts_from_subgraph(cycle_digraph: Topology) -> list[Host]:
```
rltakashige: Why are we defining get host from subgraph within the test? Feels like it makes test_get_hosts_from_subgraph kinda pointless?

src/exo/master/placement_utils.py
```diff
@@ -303,43 +246,24 @@ def _find_ip_prioritised(
...
+        iface.ip_address: iface.interface_type for iface in other_network.interfaces
+    }
 
-    if non_thunderbolt_ip:
-        return non_thunderbolt_ip
+    priority = {"ethernet": 0, "wifi": 1, "unknown": 2, "thunderbolt": 3}
 
-    if ips:
-        return ips[0][0]
+    if not ips:
```
rltakashige: Why do we do this check at the end? Let's return immediately after getting its

src/exo/master/tests/test_placement_utils.py
rltakashige: Okay it's unused code 🤣

AlexCheema: fixed

AlexCheema: removed it

src/exo/master/placement_utils.py
AlexCheema: fixed

AlexCheema:
We have a few edge cases with `apply.py` and I don't like this in general how we're computing the thunderbolt bridge cycles in `apply.py`. I'm worried about the performance impact of calling this every time topology changes, so I won't be doing that.
The better fix seems to be to compute this client-side, since all the info to compute it is already there, it just puts some burden on javascript/Swift to know how to find cycles in a graph.

PR #1224, opened 2026-01-20
remove sharding that has been upstreamed into mlx-lm

Evanev7:
LLama, Ministral, GPT-OSS, MiniMax and DeepSeek all have their tensor sharding upstream. We can remove the sharding code from exo.


Evanev7:
This should not merge until after MLX updates to fix the bias sharding.

PR #1231, opened 2026-01-20
auto_parallel.py: remove unnecessary type ignores

JakeHillion:
`basedpyright` on my system is slightly newer than that in the PyProject. Remove some ignores that it say aren't necessary.

Test plan:
- If the CI says go we go :)


JakeHillion:
Damn, CI still wants them. Can't wait to unify this behaviour!

PR #1233, opened 2026-01-21
Tool Calling

Evanev7:
tool calling yo

## test plan

OPENCODE

### DIFF
--- a/src/exo/shared/models/model_cards.py
+++ b/src/exo/shared/models/model_cards.py
@@ -627,7 +627,7 @@ async def get_config_data(model_id: ModelId) -> ConfigData:
         "main",
         "config.json",
         target_dir,
-        lambda curr_bytes, total_bytes, is_renamed: logger.info(
+        lambda curr_bytes, total_bytes, is_renamed: logger.debug(
             f"Downloading config.json for {model_id}: {curr_bytes}/{total_bytes} ({is_renamed=})"
         ),
     )
@@ -650,7 +650,7 @@ async def get_safetensors_size(model_id: ModelId) -> Memory:
         "main",
         "model.safetensors.index.json",
         target_dir,
-        lambda curr_bytes, total_bytes, is_renamed: logger.info(
+        lambda curr_bytes, total_bytes, is_renamed: logger.debug(
             f"Downloading model.safetensors.index.json for {model_id}: {curr_bytes}/{total_bytes} ({is_renamed=})"
         ),
     )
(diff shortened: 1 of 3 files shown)


src/exo/master/api.py
```diff
@@ -85,6 +87,22 @@ def chunk_to_response(
...
+                    tool_calls=[
+                        ToolCall(
+                            id=str(uuid4()),
+                            index=i,
+                            function=tool,
+                        )
+                        for i, tool in enumerate(chunk.tool_calls)
+                    ],
+                ),
+                finish_reason="tool_calls",
```
rltakashige: Let's defer to chunk.finish_reason

src/exo/master/api.py
```diff
@@ -404,16 +424,21 @@ async def delete_instance(self, instance_id: InstanceId) -> DeleteInstanceRespon
...
     async def _chat_chunk_stream(
         self, command_id: CommandId
-    ) -> AsyncGenerator[TokenChunk, None]:
+    ) -> AsyncGenerator[TokenChunk | ToolCallChunk, None]:
         """Yield `TokenChunk`s for a given command until completion."""
 
         try:
-            self._chat_completion_queues[command_id], recv = channel[TokenChunk]()
+            self._chat_completion_queues[command_id], recv = channel[
+                TokenChunk | ToolCallChunk
```
rltakashige: I think we should use TokenChunk and add a tool_calls field in TokenChunk.

src/exo/master/api.py
```diff
@@ -435,7 +460,7 @@ async def _generate_chat_stream(
         """Generate chat completion stream as JSON strings."""
 
         async for chunk in self._chat_chunk_stream(command_id):
-            if chunk.finish_reason == "error":
+            if isinstance(chunk, TokenChunk) and chunk.finish_reason == "error":
```
rltakashige: I think this is a good reason to not have ToolCallChunk

src/exo/worker/runner/runner.py
```diff
@@ -329,6 +367,120 @@ def parse_thinking_models(
         yield response
 
 
+def parse_tool_calls(
+    responses: Generator[GenerationResponse],
+    tokenizer: TokenizerWrapper,
+) -> Generator[GenerationResponse | ToolCallResponse]:
+    assert tokenizer.tool_call_start is not None
```
rltakashige: I think we should just fall through instead of failing. I'm sure some models will mess this up, in which case this is too extreme.

src/exo/master/api.py
Evanev7: tool chunks dont have a finish reason

src/exo/master/api.py
rltakashige: Discussed in person. I suppose this is a good idea although I still think it complicates things.

src/exo/master/api.py
rltakashige: Discussed in person. Let's use an ErrorChunk

PR #1235, opened 2026-01-21
dashboard: show model total size on downloads page for pending downloads

JakeHillion:
The downloads page showed "0B / 0B" for models that haven't started downloading yet because the download progress data only gets populated after the file list is fetched from HuggingFace.

Added a fetch to the /models API endpoint on page mount and created a helper function that falls back to storage_size_megabytes when the download's totalBytes is 0.

This allows users to see the actual model size (e.g., "0 / 25GB") before a download begins, which is helpful for a future feature that lets you download models explicitly.

Test plan:
- Deployed to a cluster, the previous 0B now show sensible values.

image

### DIFF
--- a/dashboard/src/routes/downloads/+page.svelte
+++ b/dashboard/src/routes/downloads/+page.svelte
@@ -172,6 +172,33 @@
   }
 
   let downloadOverview = $state<NodeEntry[]>([]);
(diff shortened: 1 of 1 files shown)


samiamjidkhan (approved):
LGTM!

PR #1236, opened 2026-01-21
downloads: refactor to run at node level

JakeHillion:
The Worker previously owned the ShardDownloader directly via dependency injection, which prevented --no-worker nodes from downloading and made it impossible for multiple Workers to share a single downloader instance.

Moved download functionality to a new DownloadCoordinator component at the Node level that communicates via the DOWNLOAD_COMMANDS pub/sub topic. Workers now send StartDownload commands instead of calling the downloader directly, and receive progress updates through the event-sourced state.

This decouples downloads from the Worker lifecycle and enables future features like UI-triggered downloads to specific nodes and multi-worker download sharing.

Test plan:
- Mostly tested in the next PR that adds explicit downloads/deletions to the dashboard.
- Started a model that isn't downloaded - it works.


src/exo/download/coordinator.py
```diff
@@ -0,0 +1,289 @@
...
+    DownloadOngoing,
+    DownloadPending,
+    DownloadProgress,
+)
+from exo.shared.types.worker.shards import ShardMetadata
+from exo.utils.channels import Receiver, Sender, channel
+from exo.utils.event_index_counter import EventIndexCounter
+
+
+class DownloadCoordinator:
```
Evanev7, line 40: can we make this a dataclass? consistency + reduced boilerplate is nice imo

src/exo/download/coordinator.py
```diff
@@ -0,0 +1,289 @@
...
+                    NodeDownloadProgress(download_progress=ongoing)
+                )
+                last_progress_time = current_time()
+
+        self.shard_downloader.on_progress(download_progress_callback)
+
+        async def download_wrapper() -> None:
+            try:
+                await self.shard_downloader.ensure_shard(shard)
+            except Exception as e:
```
Evanev7, line 180: wary about how this exception catches cancellation - asyncio tasks are only cancelled once, which leaves this dangling task. the event_sender *should* re-cancel it, but can i ask you to handle cancellation explicitly in this function?

man I miss real error types

src/exo/download/coordinator.py
```diff
@@ -0,0 +1,289 @@
...
+
+        if deleted:
+            logger.info(f"Successfully deleted model {model_id}")
+        else:
+            logger.warning(f"Model {model_id} was not found on disk")
+
+        # Emit pending status to reset UI state, then remove from local tracking
+        if model_id in self.download_status:
+            current_status = self.download_status[model_id]
+            pending = DownloadPending(
```
Evanev7, line 217: worth adding an explicit status here? Pending != Cancelled - ideally I think we'd fully remove the download from state - but I think that's beyond the scope of this pr.

src/exo/utils/event_index_counter.py
Evanev7, line 1: should we not just use an itertools.count() in place of this wrapper?

Also this doc is false. itertools count is not threadsafe, it's an int64_t (for our intents). There aren't multithreaded async runtimes for python so this will be fine for a very long time.

src/exo/worker/main.py
```diff
@@ -157,14 +150,22 @@ async def _event_applier(self):
                 for idx, event in indexed_events:
                     self.state = apply(self.state, IndexedEvent(idx=idx, event=event))
 
+    def _get_local_download_status(self) -> dict[ModelId, DownloadProgress]:
```
Evanev7: if we're pulling local download status from state - it's not local. this is pretty racy with the masters indexing.

because the downloader is idempotent this won't cause first order problems, but the "local" variables are intended to be updated by the time the plan that modifies them completes, and i don't want to violate that assumption.

Evanev7: the previous assumption was that the master creates an instance, and the worker triggers a download. It seems like we're moving that logic toward the dashboard, letting it trigger downloads/deletes and so making the magic "place instance" trigger downloads as well. food for thought.

src/exo/download/coordinator.py
JakeHillion: From 3.8 `asyncio.CancelledError` inherits from `BaseException`, not `Exception`, so this doesn't catch that (https://docs.python.org/3.8/library/asyncio-exceptions.html?highlight=cancellederror#asyncio.CancelledError).

I'm not sure what the proposed issue is though - if this gets cancelled, how does it dangle? because it calls await again and doesn't immediately return?

Anyway, this try/except can be ignored for that, it doesn't catch it.

src/exo/download/coordinator.py
JakeHillion: This replicates the existing behaviour and I'd rather not change it, mostly due to how the dashboard is rendering it - that can come down the line.

https://github.com/exo-explore/exo/blob/ac6efa747b284911c2ab34ea329c0beabec1a028/src/exo/worker/main.py#L509-L537

Whether a model has some previous progress or hasn't started downloading we put up a DownloadProgress event, and this is the same behaviour here. And as that code runs every 5 minutes emitting something like DownloadCancelled here would be temporary at best. Once we have a better system for handling downloads appearing these events should be tightened up, but I don't think it's worth doing now.

src/exo/worker/main.py
JakeHillion: > if we're pulling local download status from state - it's not local. this is pretty racy with the masters indexing.
> 
> because the downloader is idempotent this won't cause first order problems, but the "local" variables are intended to be updated by the time the plan that modifies them completes, and i don't want to violate that assumption.

Could you explain the race here a little more to me? We could always send a `StartDownload` command here and it wouldn't be an issue, this code is merely an optimisation to avoid sending an extra Command. I can document that that's idempotent if it helps.

The race I can see here is if someone on the UI has very recently pressed the delete download button, then started an Instance. If state shows the model as already being downloaded we won't send a StartDownload command, the delete will finish, and the model will fail to start. I'm not very concerned about that case with the current state of the code, and fixes to that can be self contained in the DownloadCoordinator later.

The intention of this function name was to get the download status for the local node. Perhaps I can rename it to `_get_my_nodes_download_status` or some other name to make this clearer?

JakeHillion: > the previous assumption was that the master creates an instance, and the worker triggers a download. It seems like we're moving that logic toward the dashboard, letting it trigger downloads/deletes and so making the magic "place instance" trigger downloads as well. food for thought.

Yes, this has changed. You can now use two separate flows, or any combination of them:
- Start the instance and let downloads happen automatically.
- Manually download the models you care about ahead of time, via the API or dashboard (#1237).

I don't see that these are in conflict. In fact they fit quite nicely with future download aware placement, better cluster sharing of data, nodes with far more static downloads, etc.

What's feeding your thinking?

src/exo/download/coordinator.py
Evanev7: huh - my bad. we definitely had some catch BaseExceptions in the codebase at one point, this is fine then.

src/exo/download/coordinator.py
Evanev7: reasonable

src/exo/worker/main.py
Evanev7: The worker plans - sends a download. downloader starts the download, sends an event. the event goes to the master, gets indexed, gets sent back to the worker, gets added to state. only now does the worker see the download present in its state - if it re-plans before this (the loop is every 0.1s rn iirc) it will see it hasn't started the download, and start the download again.

To some extent the shard downloader should be the one planning and triggering itself, though thats another copy of State hanging around its less likely to suffer from this "laggy state" phenomenon

JakeHillion: I added exponential backoff for this command and dropped the local naming.

PR #1237, opened 2026-01-21
downloads: add download and delete buttons to downloads UI

JakeHillion:
The downloads page showed model download progress but provided no way for users to trigger downloads or remove completed models from disk.

Added API endpoints (POST /download/start, DELETE /download/{node_id}/{model_id}) that send StartDownload and DeleteDownload commands via the download_command_sender. Updated the dashboard downloads page with per-model buttons: a download button for incomplete downloads and a delete button for completed ones.

This allows users to manage downloads directly from the UI without needing to trigger downloads through other means.

Test plan:
- Deployed on a 3 machine cluster. Did several downloads/deletions - all work and the dashboard updates relatively fluently. It takes roughly 5 seconds to render a 131GB model deletion which isn't too bad.


JakeHillion:
Closes #1010

JakeHillion:
Reminder: this is a stacked PR. Please don't merge it before #1236!

src/exo/master/api.py
```diff
@@ -680,3 +691,30 @@ async def _send(self, command: Command):
...
+    ) -> StartDownloadResponse:
+        command = StartDownload(
+            target_node_id=payload.target_node_id,
+            shard_metadata=payload.shard_metadata,
+        )
+        await self._send_download(command)
+        return StartDownloadResponse(command_id=command.command_id)
+
+    async def delete_download(
+        self, node_id: NodeId, model_id: str
```
Evanev7: why does one take a Params and the other two args? Also, this model_id can be a ModelId - pydantic will validate it propery AFAIK.

Evanev7 (dismissed):
how restrictive do you expect these reviews to be? my concern is significant enough that I'd want it acknowledged, but not significant enough that I want to re-review. Seems like an approve.

src/exo/master/api.py
JakeHillion: The model_id was missed in a rebase, thanks.

The two formats are deliberate: delete_download is a DELETE request that we don't pass a body with, it simply hits a path. start_download is a POST and has an encoded body. By my memory of APIs, this is pretty standard.

src/exo/master/api.py
```diff
@@ -1292,3 +1303,30 @@ async def _send(self, command: Command):
         await self.command_sender.send(
             ForwarderCommand(origin=self.node_id, command=command)
         )
+
+    async def _send_download(self, command: DownloadCommand):
+        while self.paused:
+            await self.paused_ev.wait()
```
AlexCheema: wouldn't this need to be on every api endpoint?

JakeHillion: Yeah, the others already have it for `command: Command`, this added it for `command: DownloadCommand`. But actually it's not needed for DownloadCommands - I'll drop it. Thanks!

PR #1248, opened 2026-01-21
dashboard: add placement filter by clicking topology nodes

AlexCheema:
## Motivation

When selecting a model for placement, users often want to see placements that utilize specific nodes in their cluster. Currently there's no way to filter the placement previews to focus on configurations that include particular machines.

## Changes

- **Backend**: Added `node_ids` query parameter to the `/placement-previews` API endpoint. When provided, the endpoint filters the topology to only include the specified nodes before generating placements using the new `Topology.filter_to_nodes()` method.

- **Topology class**: Added `filter_to_nodes(node_ids)` method that creates a new topology containing only the specified nodes and edges between them.

- **App store**: Added `previewNodeFilter` state to track selected nodes, with methods to toggle/clear the filter. Automatically cleans up filter when nodes are removed from the cluster and re-fetches previews when topology changes.

- **TopologyGraph component**: Added click handlers to toggle node filter selection, hover effects to indicate clickable nodes, and visual styling (yellow highlight for selected, dimmed for filtered-out nodes).

- **Main page**: Added filter indicator in top-right corner of topology showing active filter count with a clear button.

## Why It Works

The filtering happens at the backend/placement generation level rather than just filtering the results. This ensures we see all valid placement combinations for the selected nodes, not just a subset that happened to be generated for the full topology.

The visual feedback uses the same rendering approach as the existing highlight system - state is tracked in Svelte and applied during render, so it persists across data updates without flickering.

### Manual Testing

- Click a node in topology → should show yellow highlight and filter indicator
- Click another node → indicator shows "2 nodes", previews update to show only placements using both
- Hover over nodes → subtle yellow highlight indicates they're clickable
- Click X on filter indicator → clears filter, shows all placements again
- Disconnect a node while it's in filter → filter auto-removes that node

### Automated Testing
- Existing tests cover the Topology class; the new `filter_to_nodes` method follows the same patterns

### DIFF
--- a/src/exo/master/api.py
+++ b/src/exo/master/api.py
@@ -339,15 +339,14 @@ class API:
     async def get_placement_previews(
         self,
         model_id: ModelId,
-        node_ids: Annotated[list[str] | None, Query()] = None,
+        node_ids: Annotated[list[NodeId] | None, Query()] = None,
     ) -> PlacementPreviewResponse:
         seen: set[tuple[ModelId, Sharding, InstanceMeta, int]] = set()
         previews: list[PlacementPreview] = []
 
         # Create filtered topology if node_ids specified
         if node_ids and len(node_ids) > 0:
-            filter_set = set(NodeId(n) for n in node_ids)
-            topology = self.state.topology.filter_to_nodes(filter_set)
+            topology = self.state.topology.get_subgraph_from_nodes(node_ids)
         else:
             topology = self.state.topology
 
--- a/src/exo/shared/topology.py
+++ b/src/exo/shared/topology.py
@@ -58,19 +58,6 @@ class Topology:
 
         return topology
 
-    def filter_to_nodes(self, node_ids: set[NodeId]) -> "Topology":
-        """Create a new topology containing only the specified nodes and edges between them."""
-        filtered = Topology()
-        for node_id in self.list_nodes():
-            if node_id in node_ids:
-                filtered.add_node(node_id)
-        for node_id in node_ids:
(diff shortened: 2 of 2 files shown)


src/exo/master/api.py
```diff
@@ -310,11 +310,21 @@ async def get_placement(
         return placements[new_ids[0]]
 
     async def get_placement_previews(
-        self, model_id: ModelId
+        self,
+        model_id: ModelId,
+        node_ids: Annotated[list[str] | None, Query()] = None,
```
Evanev7: no reason this cant be a list of NodeId

src/exo/shared/topology.py
```diff
@@ -53,6 +53,19 @@ def from_snapshot(cls, snapshot: TopologySnapshot) -> "Topology":
...
+        for node_id in self.list_nodes():
+            if node_id in node_ids:
+                filtered.add_node(node_id)
+        for node_id in node_ids:
+            if self.contains_node(node_id):
+                for conn in self.out_edges(node_id):
+                    if conn.sink in node_ids:
+                        filtered.add_connection(conn)
+        return filtered
+
```
Evanev7: we already have `get_subgraph_from_nodes`.

src/exo/master/api.py
AlexCheema: good point - done

src/exo/shared/topology.py
AlexCheema: yup, using that now

Issue #1242, opened 2026-01-21
[BUG] black-forest-labs models not able to be downloaded


frenchtoasters:
## Describe the bug

`black-forest-labs/FLUX.1-dev` and other models from them now require you to share `email and username` with them. This breaks downloads on machines that dont have a hugging face token. 

```
[ 02:09:14.0771PM | ERROR   ] Traceback (most recent call last):
  File "/Users/toast/exo/src/exo/worker/download/download_utils.py", line 279, in download_file_with_retry
    return await _download_file(
           ^^^^^^^^^^^^^^^^^^^^^
        model_id, revision, path, target_dir, on_progress
        ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
    )
    ^
  File "/Users/toast/exo/src/exo/worker/download/download_utils.py", line 325, in _download_file
    assert r.status in [200, 206], (
           ^^^^^^^^^^^^^^^^^^^^^^
AssertionError: Failed to download config.json from https://huggingface.co/black-forest-labs/FLUX.1-schnell/resolve/main/config.json: 401
```

## To Reproduce

1.
2.
3.

## Actual behavior

A clear and concise description of what actually happened.

## Environment

- macOS Version:
- EXO Version:
- Hardware:
  - Device 1: (e.g., MacBook Pro M1 Max, 32GB RAM)
  - Device 2: (e.g., Mac Mini M2, 16GB RAM)
  - Additional devices:
- Interconnection:
  - (e.g., Thunderbolt 4 cable between Device 1 and 2)
  - (e.g., WiFi 6 for Device 3)
  - (e.g., 10GbE Ethernet between all devices)

PR #1250, opened 2026-01-22
Ciaran/hf token

ciaranbor:
## Motivation

black-forest-labs models require hf auth and signup to download. We don't handle this gracefully.
https://github.com/exo-explore/exo/issues/1242

## Changes

- Handle auth errors
- Surface error to UI and suggest resolution
- Support using HF_TOKEN env variable for auto
- Hide image functionality behind `EXO_ENABLE_IMAGE_MODELS=true` for now

## Why It Works

Users are presented with actionable feedback when issue occurs

### Manual Testing

Confirmed loading black-forest-labs model in UI presents the issue in the UI.
Confirmed both `hf auto login` and setting `HF_TOKEN` resolve the issue

### DIFF
--- a/app/EXO/EXO/ContentView.swift
+++ b/app/EXO/EXO/ContentView.swift
@@ -25,6 +25,7 @@ struct ContentView: View {
     @State private var uninstallInProgress = false
     @State private var pendingNamespace: String = ""
     @State private var pendingHFToken: String = ""
+    @State private var pendingEnableImageModels = false
 
     var body: some View {
         VStack(alignment: .leading, spacing: 12) {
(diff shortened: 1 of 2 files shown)


AlexCheema:
We also need some UI for this. Most people will run the macOS app, which doesn't support arbitrary environment variable overrides.

src/exo/worker/main.py
```diff
@@ -443,7 +444,33 @@ async def download_progress_callback(
                 last_progress_time = current_time()
 
         self.shard_downloader.on_progress(download_progress_callback)
-        self._tg.start_soon(self.shard_downloader.ensure_shard, task.shard_metadata)
+
+        async def download_with_error_handling() -> None:
```
Evanev7, line 448: this is the main thing id want to change, but with jake's changes inbound im happy to go with this for now to get the image models back into exo proper.

PR #1256, opened 2026-01-23
Fix Kimi-K2-Thinking tool calls appearing in thinking output

AlexCheema:
## Motivation

Kimi-K2-Thinking was outputting raw tool call tokens (`<|tool_call_begin|>`, `<|tool_call_argument_begin|>`, etc.) in its thinking output, making the output unreadable.

## Changes

- Created `filter_kimi_tokens_and_parse_tool_calls()` that handles Kimi tool calls at the raw token level, BEFORE `parse_gpt_oss` transforms them
- Internal tool calls (UUID format used by Kimi during thinking) are now silently filtered out
- Updated `parse_gpt_oss`, `parse_thinking_models`, and `parse_tool_calls` to pass through `ToolCallResponse`s unchanged
- Non-Kimi models continue to use the existing pipeline

## Why It Works

The issue was that `parse_gpt_oss` transforms the token stream into text deltas before tool call detection happens. Kimi-K2-Thinking uses internal tool calls during thinking (format: `UUID<|tool_call_argument_begin|>args`) that don't match the external tool call format (`functions.name:0 <|...|> args`). By parsing tool calls before `parse_gpt_oss`, we can detect them at the token level and filter out the internal ones.

### Manual Testing
- Tested with Kimi-K2-Thinking model, verified tool call tokens no longer appear in thinking output

### Automated Testing
- Type checking passes (`basedpyright`)
- Linting passes (`ruff check`)

### DIFF
--- a/src/exo/worker/runner/runner.py
+++ b/src/exo/worker/runner/runner.py
@@ -240,6 +240,16 @@ def main(
                             prompt=prompt,
                         )
 
+                        # Kimi-K2 has tool call sections and tool calls that need to be
+                        # handled BEFORE parse_gpt_oss, because parse_gpt_oss transforms
+                        # the token stream into text deltas which prevents tool call detection
+                        is_kimi = "kimi" in shard_metadata.model_card.model_id.lower()
+                        if is_kimi:
+                            patch_kimi_tokenizer(tokenizer)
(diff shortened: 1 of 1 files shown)


AlexCheema:
needs a closer look

PR #1260, opened 2026-01-23
dashboard: decouple prettier-svelte from dashboard source

JakeHillion:
The prettier-svelte formatter depended on the full dashboard build (dashboardFull), causing the devshell to rebuild whenever any dashboard source file changed.

Created a deps-only dream2nix derivation (deps.nix) that uses a stub source containing only package.json, package-lock.json, and minimal files for vite to succeed. Updated prettier-svelte to use this derivation instead of dashboardFull.

The stub source is constant unless lockfiles change, so prettier-svelte and the devshell no longer rebuild when dashboard source files are modified.

Test plan:
- nix flake check passed
- nix fmt successfully formatted svelte files

### DIFF
--- a/dashboard/parts.nix
+++ b/dashboard/parts.nix
@@ -3,6 +3,45 @@
   perSystem =
     { pkgs, lib, ... }:
     let
+      # Stub source with lockfiles and minimal files for build to succeed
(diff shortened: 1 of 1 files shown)


JakeHillion:
There's still a bit more duplication here than I'd like, but it does the job.

PR #1262, opened 2026-01-23
Fix kv prefix cache

rltakashige:
## Motivation

OpenCode sends very large prompts, most of which are repeated on the next call. 

## Changes

Add prefix caching, reducing average time in prefill (in testing) from 40 seconds to 4. This massively improves user experience. 

Also evicts KV caches from this prefix cache in a LRU-style manner. 

## Why It Works

We no longer prefill repeatedly but rather use kv cache stored in memory. A future update may want to use storage to make the prefix cache larger.

### Manual Testing
Tested speedup on OpenCode

### Automated Testing
Added a lot of tests

### DIFF
--- a/src/exo/worker/engines/mlx/utils_mlx.py
+++ b/src/exo/worker/engines/mlx/utils_mlx.py
@@ -402,7 +402,11 @@ def apply_chat_template(
                 continue
 
             message.content = "\n".join(c.text for c in message.content).strip()
-        if message.content is None and message.thinking is None:
+        if (
+            message.content is None
+            and message.thinking is None
+            and message.tool_calls is None
+        ):
             continue
 
         # Null values are not valid when applying templates in tokenizer


src/exo/worker/engines/mlx/cache.py
```diff
@@ -1,39 +1,74 @@
...
+        self._last_used: list[int] = []  # monotonic counter of last access per entry
+        self._access_counter: int = 0
+
+    def clear(self):
+        """Clear all cached prompts and caches."""
+        self.prompts.clear()
+        self.caches.clear()
+        self._last_used.clear()
 
     def add_kv_cache(
```
Evanev7: its a nit but imo none of these functions should accept a tokenizer if theyre just going to immediately tokenize the prompt - cache being a mx.array -> KVCache makes more sense to me.

src/exo/worker/engines/mlx/cache.py
```diff
@@ -1,39 +1,74 @@
...
 from mlx_lm.tokenizer_utils import TokenizerWrapper
 
+from exo.shared.types.mlx import KVCacheType
 from exo.worker.engines.mlx import Model
-from exo.worker.engines.mlx.constants import KEEP_KV_SIZE, KV_BITS, KV_GROUP_SIZE
 from exo.worker.engines.mlx.utils_mlx import make_kv_cache
+from exo.worker.runner.bootstrap import logger
+
+# Fraction of device memory above which LRU eviction kicks in
+_MEMORY_PRESSURE_THRESHOLD = 0.85
```
Evanev7: we should adjust this threshold based on device memory. i wonder if something like a 5GB flat limit might be better. I also think this should go in the main constants / be overrideable as an env var.

src/exo/shared/types/mlx.py
```diff
@@ -0,0 +1,11 @@
+"""Shared types for MLX-related functionality."""
+
+from mlx_lm.models.cache import (
+    KVCache,
+    QuantizedKVCache,
+    RotatingKVCache,
+)
+
+# Type alias for KV cache - matches make_kv_cache return type
```
Evanev7: include in make_kv_cache?

src/exo/worker/engines/mlx/cache.py
rltakashige: _MEMORY_PRESSURE_THRESHOLD was awful naming. It just uses MEMORY_THRESHOLD. Made it an env variable.

rltakashige: We can look into this further in future.

src/exo/worker/engines/mlx/cache.py
rltakashige: KV Prefix Cache is now initialized with a tokenizer.

PR #1269, opened 2026-01-23
Add FLUX.1-Krea-dev model

ciaranbor:
## Why It Works

Same implementation as FLUX.1-dev, just different weights


src/exo/download/download_utils.py
```diff
@@ -481,6 +481,13 @@ async def resolve_allow_patterns(shard: ShardMetadata) -> list[str]:
         return ["*"]
 
 
+def is_image_model(shard: ShardMetadata) -> bool:
+    from exo.shared.models.model_cards import ModelTask
```
AlexCheema: import at top

PR #1271, opened 2026-01-23
Parse GPT OSS tool calling

rltakashige:
## Motivation

image

## Changes

OpenAI Harmony StreamableParser does parsing for us.

## Why It Works

image


src/exo/worker/runner/runner.py
```diff
@@ -257,10 +253,16 @@ def on_model_load_timeout() -> None:
                             patch_kimi_tokenizer(tokenizer)
 
                         # GLM models need patched parser (upstream has bug with None regex match)
-                        if "glm" in shard_metadata.model_card.model_id.lower():
+                        elif "glm" in shard_metadata.model_card.model_id.lower():
```
AlexCheema, line 256: nitpick: we have different ways of detecting the model

PR #1274, opened 2026-01-23
Distribute model downloads via master

### LINKED ISSUE #1257: Why should all nodes download the model directly from Hugging Face?
It's recommended that only the master node download the model from Hugging Face and then distribute it to other nodes over the local network.

jayy-77:
## Summary
- Add a master-only model-store server (HuggingFace-like `tree` + `resolve` w/ Range) to serve cached model files over the LAN.
- Make shard downloads cluster-aware: master downloads from Hugging Face; non-master nodes fetch from the master’s model-store endpoint (retrying while files are still being downloaded).
- When a worker needs a model, it also triggers a download on the elected master so other nodes can pull locally.

## Config
- `--model-store-port` (default: 52416)

## Why
Reduces redundant internet downloads in multi-node clusters while keeping existing download integrity checks.

Closes #1257

## Test plan
- `uv run pytest -q`
- `uv run ruff check ...`
- `uv run basedpyright`

### DIFF
--- a/src/exo/download/coordinator.py
+++ b/src/exo/download/coordinator.py
@@ -12,7 +12,10 @@ from exo.download.download_utils import (
     delete_model,
     map_repo_download_progress_to_download_progress_data,
 )
+from exo.download.model_store_server import ModelStoreServer
+from exo.download.node_address_book import NodeAddressBook
 from exo.download.shard_downloader import ShardDownloader
+from exo.routing.connection_message import ConnectionMessage, ConnectionMessageType
 from exo.shared.models.model_cards import ModelId
 from exo.shared.types.commands import (
     DeleteDownload,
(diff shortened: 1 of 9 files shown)


JakeHillion:
Hi Jay, I took a look at this. I've been working on refactoring downloads to more effectively handle sharing files around, and it's getting close to complete (no promises of when it will be coming or on by default though!). I totally agree this is one of the most annoying things when setting up a new cluster.

I don't think directing downloads through the master is a good solution though. Imagine you have a cluster like this:
- MacBook Air (256GB SSD)
- Mac Studio (1TB SSD) * 2

Your MacBook might be elected the master, and yet it's a totally inappropriate device for running the downloads through - you literally wouldn't be able to run Kimi-K2-Thinking for example, because the MacBook's SSD is too small.

The approach I'm taking here at the moment is to use HTTP backed torrents. They download versioned files from HuggingFace with consistent hashes, then the cluster is able to share them internally. This should get you the best speed and utilisation of all your resources.

I'm going to close this PR for now. Let me know if this won't meet your needs when it lands, and I'm excited for you to try it.

PR #1276, opened 2026-01-24
api cancellation

### LINKED ISSUE #61: No way to cancel
There's no way to stop an inference. Reloading or closing the webpage doesn't stop it either. The nodes just keep going.

Evanev7:
closing the http request to the api now
- sends a cancellation from the api
- writes that canellation in the master
- worker plans off the cancellation
- runner observes that cancellation after every generation step (+1 communication per token)
- cancellation happens synchronously to prevent gpu locks

closes #61

### DIFF
--- a/src/exo/worker/runner/runner.py
+++ b/src/exo/worker/runner/runner.py
@@ -560,9 +560,15 @@ def main(
                     raise ValueError(
                         f"Received {task.__class__.__name__} outside of state machine in {current_status=}"
                     )
-            event_sender.send(
-                TaskStatusUpdated(task_id=task.task_id, task_status=TaskStatus.Complete)
+            was_cancelled = (task.task_id in cancelled_tasks) or (
+                TaskId("CANCEL_CURRENT_TASK") in cancelled_tasks
             )
+            if not was_cancelled:
+                event_sender.send(
+                    TaskStatusUpdated(
+                        task_id=task.task_id, task_status=TaskStatus.Complete
+                    )
+                )
             event_sender.send(
                 RunnerStatusUpdated(runner_id=runner_id, runner_status=current_status)
             )
--- a/src/exo/master/tests/test_claude_tool_use.py
+++ b/src/exo/master/tests/test_claude_tool_use.py
@@ -4,7 +4,11 @@ import json
 from collections.abc import AsyncGenerator
 from typing import Any, cast
 
-from exo.master.adapters.claude import collect_claude_response, generate_claude_stream
+from exo.master.adapters.claude import (
+    ClaudeMessagesResponse,
+    collect_claude_response,
+    generate_claude_stream,
+)
 from exo.shared.types.api import ToolCallItem
 from exo.shared.types.chunks import ErrorChunk, TokenChunk, ToolCallChunk
 from exo.shared.types.common import CommandId, ModelId
@@ -15,6 +19,18 @@ async def _chunks_to_stream(
 ) -> AsyncGenerator[ErrorChunk | ToolCallChunk | TokenChunk, None]:
     for chunk in chunks:
         yield chunk
+
+
+async def _collect_response(
+    command_id: CommandId,
+    model: str,
+    chunk_stream: AsyncGenerator[ErrorChunk | ToolCallChunk | TokenChunk, None],
+) -> ClaudeMessagesResponse:
+    """Helper to consume the async generator and parse the JSON response."""
+    parts: list[str] = []
+    async for part in collect_claude_response(command_id, model, chunk_stream):
+        parts.append(part)
+    return ClaudeMessagesResponse.model_validate_json("".join(parts))
 
 
 MODEL = ModelId("test-model")
@@ -47,7 +63,7 @@ class TestCollectClaudeResponseToolUse:
                 ],
             ),
         ]
-        response = await collect_claude_response(
+        response = await _collect_response(
             COMMAND_ID, "test-model", _chunks_to_stream(chunks)
         )
 
@@ -77,7 +93,7 @@ class TestCollectClaudeResponseToolUse:
                 ],
             ),
         ]
-        response = await collect_claude_response(
+        response = await _collect_response(
             COMMAND_ID, "test-model", _chunks_to_stream(chunks)
         )
 
@@ -102,7 +118,7 @@ class TestCollectClaudeResponseToolUse:
                 ],
             ),
         ]
-        response = await collect_claude_response(
+        response = await _collect_response(
             COMMAND_ID, "test-model", _chunks_to_stream(chunks)
         )
 
@@ -116,7 +132,7 @@ class TestCollectClaudeResponseToolUse:
 
     async def test_no_content_produces_empty_text_block(self):
         chunks: list[ErrorChunk | ToolCallChunk | TokenChunk] = []
-        response = await collect_claude_response(
+        response = await _collect_response(
             COMMAND_ID, "test-model", _chunks_to_stream(chunks)
         )
         assert len(response.content) == 1


MISSED_THINGS.md
```diff
@@ -5,16 +5,16 @@
...
+[X] same as above with "sharded-to-all" became _sharded_to_all in auto_parallel.py.
+[X] Dropped support for Ministral3Model, DeepseekV32Model, Glm4MoeModel, Qwen3NextModel, GptOssMode in auto_parallel.py.
 [] Dropped prefill/decode code in auto_parallel.py and utils_mlx.py.
 [X] KV_CACHE_BITS should be None to disable quantized KV cache.
-[] Dropped _set_nofile_limit in utils_mlx.py.
-[] We have group optional in load_mlx_items in utils_mlx.py.
+[X] Dropped _set_nofile_limit in utils_mlx.py.
+[X] We have group optional in load_mlx_items in utils_mlx.py.
 [] Dropped add_missing_chat_templates for GptOss in load_mlx_items in utils_mlx.py.
 [] Dropped model.make_cache in make_kv_cache in utils_mlx.py.
```
rltakashige: Done

src/exo/master/api.py
```diff
@@ -508,16 +509,14 @@ async def _chat_chunk_stream(
...
-            )
-            """
+            command = TaskCancelled(cancelled_command_id=command_id)
+            with anyio.CancelScope(shield=True):
+                await self.command_sender.send(
+                    ForwarderCommand(origin=self.node_id, command=command)
+                )
             raise
         finally:
-            command = TaskFinished(finished_command_id=command_id)
```
rltakashige, line 551: Why did you make this change here but inconsistently in this file?

ciaranbor:
I think this is a bit more complicated for image models. Comms happen between +/-1 denoising steps between different ranks so we could end up with asymmetric sends/recvs. Also, only the final rank yields data so the other ranks will not cancel as I understand it. Cancellation will also only work when streaming is enable dand only take effect on partial image boundaries. I'll think about how to handle this. I'd prefer not to have to sync between denoising steps as this will impact the level of parallelism we get

src/exo/master/main.py
```diff
@@ -308,6 +321,17 @@ async def _plan(self) -> None:
...
+                        if task.instance_id == instance_id and task.task_status in [
+                            TaskStatus.Pending,
+                            TaskStatus.Running,
+                        ]:
+                            await self.event_sender.send(
+                                TaskStatusUpdated(
+                                    task_status=TaskStatus.Cancelled,
+                                    task_id=task.task_id,
+                                )
+                            )
```
rltakashige: Not sure I understand this logic. Was this meant to go after `if node_id not in connected_node_ids`?

src/exo/worker/engines/mlx/generator/generate.py
```diff
@@ -90,10 +89,6 @@ def warmup_inference(
 
     logger.info("Generated ALL warmup tokens")
 
-    # TODO: Do we want an mx_barrier?
-    #  At least this version is actively incorrect, as it should use mx_barrier(group)
-    mx_barrier()
```
rltakashige: Don't think we want to clean this up here without testing.

src/exo/worker/runner/runner.py
```diff
@@ -99,8 +102,11 @@ def main(
         time.sleep(timeout)
 
     setup_start_time = time.time()
+    cancelled_tasks = set[TaskId]()
 
-    model: Model | DistributedImageModel | None = None
+    # type checker was unhappy with me - splitting these fixed it
+    inference_model: Model | None = None
+    image_model: DistributedImageModel | None = None
```
rltakashige, line 139: Nice, I think it's definitely good to split these up. Not sure if we should call it inference_model, but whatever.

src/exo/worker/engines/mlx/utils_mlx.py
```diff
@@ -536,3 +510,33 @@ def mlx_cleanup(
...
+
+
+def mx_all(bool_: bool, group: Group | None) -> bool:
+    if group is None:
+        return bool_
+    num_true = mx.distributed.all_sum(
+        mx.array(bool_), group=group, stream=mx.default_stream(mx.Device(mx.cpu))
+    )
+    mx.eval(num_true)
+    return num_true.item() == group.size()
```
rltakashige: Where is this used?

src/exo/worker/runner/runner.py
```diff
@@ -273,7 +278,17 @@ def on_model_load_timeout() -> None:
...
+                        last_checked = time.perf_counter()
                         for response in mlx_generator:
+                            if (t := time.perf_counter()) - last_checked > 0.1:
+                                last_checked = t
+                                cancelled_tasks.update(cancel_receiver.collect())
+                                want_to_cancel = (task.task_id in cancelled_tasks) or (
+                                    TaskId("CANCEL_CURRENT_TASK") in cancelled_tasks
+                                )
+                                if mx_any(want_to_cancel, group):
+                                    break
```
rltakashige: Could you check performance against exo_bench? 

For example, I don't see any issue with, for example, putting this every second instead of 0.1s. Or the fact we are using a different stream will be enough, but perhaps we need to explicitly set a new stream?

src/exo/worker/engines/mlx/utils_mlx.py
Evanev7: it's not, just wanted parity

src/exo/worker/engines/mlx/generator/generate.py
Evanev7: ive done reasonable testing - this doesn't seem to have any impact. i'll do a bench as well, though.

src/exo/master/main.py
Evanev7: good catch, this was meant to cancel any tasks in a deleted instance so is in completely the wrong place.

src/exo/master/api.py
Evanev7: because the original was inconsistent iirc

Evanev7:
> why don't we run a loop which sets a flag, rather than doing an all sum within the generator loop, which might be asymmetric. Then, the implementation of the generator can decide how to check this flag.

I don't see what you mean by this, the sum will be in the generator loop no matter what, and the entire purpose of the all sum is to ensure there isnt asymmetry.

src/exo/master/main.py
Evanev7: yeah this was wrong, good catch. Should be in "get_transition_events" after InstanceDeleted

src/exo/worker/engines/mlx/generator/generate.py
Evanev7: I tested!

src/exo/worker/engines/mlx/utils_mlx.py
Evanev7: It's not, tbf.

src/exo/worker/runner/runner.py
Evanev7: good shout

rltakashige:
To just summarise internal discussions - yeah this is a complicated topic and both the suggestion and the solution as is are still asymmetric. But we do have a few ideas on how to resolve it.

ciaranbor:
Reminder to ensure it works for non-streaming mode before merging

AlexCheema:
Just tested this.
It worked when closing the HTTP connection during a text generation task - it cancelled correctly. Once, it didn't cancel - task continued, but haven't been able to reproduce that yet.

More importantly, this causes a crash when deleting an instance while a task is running:

```
hypercorn.utils.LifespanFailureError: Lifespan failure in shutdown. '  + Exception Group Traceback (most recent call last):
  |   File "/Users/gary/coding-2/exo/.venv/lib/python3.13/site-packages/anyio/_backends/_asyncio.py", line 2325, in run
  |     return runner.run(wrapper())
  |            ~~~~~~~~~~^^^^^^^^^^^
  |   File "/opt/homebrew/Cellar/python@3.13/3.13.11/Frameworks/Python.framework/Versions/3.13/lib/python3.13/asyncio/runners.py", line 118, in run
  |     return self._loop.run_until_complete(task)
  |            ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~^^^^^^
  |   File "/opt/homebrew/Cellar/python@3.13/3.13.11/Frameworks/Python.framework/Versions/3.13/lib/python3.13/asyncio/base_events.py", line 725, in run_until_complete
  |     return future.result()
  |            ~~~~~~~~~~~~~^^
  |   File "/Users/gary/coding-2/exo/.venv/lib/python3.13/site-packages/anyio/_backends/_asyncio.py", line 2313, in wrapper
  |     return await func(*args)
  |            ^^^^^^^^^^^^^^^^^
  |   File "/Users/gary/coding-2/exo/src/exo/main.py", line 138, in run
  |     async with self._tg as tg:
  |                ^^^^^^^^
  |   File "/Users/gary/coding-2/exo/.venv/lib/python3.13/site-packages/anyio/_backends/_asyncio.py", line 781, in __aexit__
  |     raise BaseExceptionGroup(
  |         "unhandled errors in a TaskGroup", self._exceptions
  |     ) from None
  +-+---------------- 1 ----------------
    | Exception Group Traceback (most recent call last):
    |   File "/Users/gary/coding-2/exo/src/exo/worker/main.py", line 102, in run
    |     async with self._tg as tg:
    |                ^^^^^^^^
    |   File "/Users/gary/coding-2/exo/.venv/lib/python3.13/site-packages/anyio/_backends/_asyncio.py", line 781, in __aexit__
    |     raise BaseExceptionGroup(
    |         "unhandled errors in a TaskGroup", self._exceptions
    |     ) from None
    +-+---------------- 1 ----------------
      | Traceback (most recent call last):
      |   File "/Users/gary/coding-2/exo/src/exo/worker/main.py", line 234, in plan_step
      |     await self.runners[self._task_to_runner_id(task)].cancel_task(
      |                        ~~~~~~~~~~~~~~~~~~~~~~~^^^^^^
      |   File "/Users/gary/coding-2/exo/src/exo/worker/main.py", line 284, in _task_to_runner_id
      |     instance = self.state.instances[task.instance_id]
      |                ~~~~~~~~~~~~~~~~~~~~^^^^^^^^^^^^^^^^^^
      | KeyError: '94b8626e-3607-4069-a063-860dd988c70a'
      +------------------------------------

  File "/Users/gary/coding-2/exo/.venv/lib/python3.13/site-packages/starlette/routing.py", line 701, in lifespan
    await receive()
  File "/Users/gary/coding-2/exo/.venv/lib/python3.13/site-packages/hypercorn/asyncio/lifespan.py", line 109, in asgi_receive
    return await self.app_queue.get()
           ^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/opt/homebrew/Cellar/python@3.13/3.13.11/Frameworks/Python.framework/Versions/3.13/lib/python3.13/asyncio/queues.py", line 186, in get
    await getter
asyncio.exceptions.CancelledError
'
[ 06:10:19.8116AM | INFO    ] bye from the runner
  + Exception Group Traceback (most recent call last):
  |   File "/Users/gary/coding-2/exo/.venv/bin/exo", line 10, in <module>
  |     sys.exit(main())
  |              ~~~~^^
  |   File "/Users/gary/coding-2/exo/src/exo/main.py", line 267, in main
  |     anyio.run(node.run)
  |     ~~~~~~~~~^^^^^^^^^^
  |   File "/Users/gary/coding-2/exo/.venv/lib/python3.13/site-packages/anyio/_core/_eventloop.py", line 74, in run
  |     return async_backend.run(func, args, {}, backend_options)
  |            ~~~~~~~~~~~~~~~~~^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  |   File "/Users/gary/coding-2/exo/.venv/lib/python3.13/site-packages/anyio/_backends/_asyncio.py", line 2325, in run
  |     return runner.run(wrapper())
  |            ~~~~~~~~~~^^^^^^^^^^^
  |   File "/opt/homebrew/Cellar/python@3.13/3.13.11/Frameworks/Python.framework/Versions/3.13/lib/python3.13/asyncio/runners.py", line 118, in run
  |     return self._loop.run_until_complete(task)
  |            ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~^^^^^^
  |   File "/opt/homebrew/Cellar/python@3.13/3.13.11/Frameworks/Python.framework/Versions/3.13/lib/python3.13/asyncio/base_events.py", line 725, in run_until_complete
  |     return future.result()
  |            ~~~~~~~~~~~~~^^
  |   File "/Users/gary/coding-2/exo/.venv/lib/python3.13/site-packages/anyio/_backends/_asyncio.py", line 2313, in wrapper
  |     return await func(*args)
  |            ^^^^^^^^^^^^^^^^^
  |   File "/Users/gary/coding-2/exo/src/exo/main.py", line 138, in run
  |     async with self._tg as tg:
  |                ^^^^^^^^
  |   File "/Users/gary/coding-2/exo/.venv/lib/python3.13/site-packages/anyio/_backends/_asyncio.py", line 781, in __aexit__
  |     raise BaseExceptionGroup(
  |         "unhandled errors in a TaskGroup", self._exceptions
  |     ) from None
  +-+---------------- 1 ----------------
    | Exception Group Traceback (most recent call last):
    |   File "/Users/gary/coding-2/exo/src/exo/worker/main.py", line 102, in run
    |     async with self._tg as tg:
    |                ^^^^^^^^
    |   File "/Users/gary/coding-2/exo/.venv/lib/python3.13/site-packages/anyio/_backends/_asyncio.py", line 781, in __aexit__
    |     raise BaseExceptionGroup(
    |         "unhandled errors in a TaskGroup", self._exceptions
    |     ) from None
    +-+---------------- 1 ----------------
      | Traceback (most recent call last):
      |   File "/Users/gary/coding-2/exo/src/exo/worker/main.py", line 234, in plan_step
      |     await self.runners[self._task_to_runner_id(task)].cancel_task(
      |                        ~~~~~~~~~~~~~~~~~~~~~~~^^^^^^
      |   File "/Users/gary/coding-2/exo/src/exo/worker/main.py", line 284, in _task_to_runner_id
      |     instance = self.state.instances[task.instance_id]
      |                ~~~~~~~~~~~~~~~~~~~~^^^^^^^^^^^^^^^^^^
      | KeyError: '94b8626e-3607-4069-a063-860dd988c70a'
      +------------------------------------
```

PR #1280, opened 2026-01-24
Ciaran/image cancellation

ciaranbor:


src/exo/worker/engines/image/pipeline/runner.py
```diff
@@ -148,6 +151,54 @@ def _get_effective_guidance_scale(self) -> float | None:
             return self._guidance_override
         return self.config.guidance_scale
 
+    def _check_cancellation(self) -> bool:
+        if self._cancelling:
+            return True
+        if (
+            self.is_first_stage
```
rltakashige, line 158: Why only on the first stage? Maybe I'm misunderstanding.

Issue #606, opened 2025-01-16
Support for Windows Platform


xxwtiancai:
Hi, thank you for this great project!

I'm interested in running this project on Windows. Currently, I noticed there's no explicit documentation or support for Windows environments. Would it be possible to:

I'd like to ask if there are any solutions for running exo on Windows or WSL. 
I've tried all code versions from July until now and reviewed over 300 issues for possible solutions. Unfortunately, 
I still haven't managed to get it running successfully on my two Windows machines - I can't even find other nodes except my own. 
I would be very grateful if you could provide some optimization solutions for running on Windows.

tensorsofthewall:
exo is still not supported for native Windows. I got my WSL setup working with this: [#455 ](https://github.com/exo-explore/exo/issues/455#issuecomment-2480927523)

xxwtiancai:
> exo is still not supported for native Windows. I got my WSL networking issue working with this: [#455 ](https://github.com/exo-explore/exo/issues/455#issuecomment-2480927523)

Thanks for your advice. But it seems do not work on my windows pc. I have tried this issue, added a .wslconfig, exo run correctly, but only detect one node, my two pc still cannot find each other. They are in same network.
Did you make other actions to make exo working on your Windows PC? Such as open/close Windows Firewalls.

tensorsofthewall:
On Windows, my firewall is currently off. But I didn't have to do anything to get my WSL exo node detected and synced with my Mac node. 

If this is the issue, you can add rules in your firewall to allow incoming/outgoing connections on ports used by exo. By default, exo uses UDP so you could allow incoming connections on ports 49152 to 65535. You can check for the specific port numbers in [exo/main.py](https://github.com/exo-explore/exo/blob/c260689a068f0cb4e74e49c9a8e76af2bbc24b26/exo/main.py#L38), mainly in args, if you don't want to expose all ports (I highly recommend you do this instead of option 1).

xxwtiancai:
Thanks a lot.
I sure that I have tried to allow incoming/outgoing connections on ports, but the two of my windows PC still do not work.
I want to know whether your WSL node detected your Mac? Becsuse as I known, many developers have the same problem that windows PC cannot detect any node, but other operate system pc can detect windows. But two of my pc are Windows :( .

tensorsofthewall:
Yes, my WSL node was able to detect the Mac with no issues. I tested with disconnections and everything as well. I had issues with inference, but I'm still sorting that out as we speak.

[...]

PR #1289, opened 2026-01-26
WIP: local changes made for exo windows support 

peterxing:


peterxing:
this is the exo for windows support from here https://github.com/exo-explore/exo/issues/606#issuecomment-3794790600

peterxing:
my clawdbot says:

• Added remote fork → https://github.com/peterxing/exo.git
• Created branch peter/wip
• Committed your local changes (WIP: local changes)
• Pushed successfully: https://github.com/peterxing/exo/tree/peter/wip
DLL/EXE handling:

• I updated .gitignore to ignore root /*.dll and /*.exe so those llama.cpp CUDA 12.4 Windows artifacts won’t get committed.
• I updated docs/windows.md to tell users to download the Windows x64 CUDA 12.4 build from https://github.com/ggml-org/llama.cpp/releases/tag/b7836 and place the binaries (including DLLs) next to llama-server.exe (repo root is fine), then set EXO_LLAMA_SERVER_PATH.

hksfho:
Does it support Windows CPU?

wrapss:
Hello,
Thank you for initiating the work on this. With a few additional modifications, I've managed to get it working on macOS, Windows (AMD GPU via Vulkan), and Linux (NVIDIA GPU) with llama.cpp.
There's still a lot of work to be done to make everything fully operational, such as retrieving GPU memory usage for AMD on Windows and other features.
Can I create a PR to your branch so we can make a comprehensive PR together to support llama.cpp?
cc @Evanev7 
image

peterxing:
Amazing! Yes of course, great to hear

Evanev7:
Would love to see what changes are necessary to support something like Llama cpp as we work out our runner/engine abstraction (see #1347)

AlexCheema:
I think this makes sense to split up into separate PRs for the windows-specific things and Llama cpp support.

AlexCheema:
## Code Review — PR #1289: WIP: local changes made for exo windows support

**CI Status:** No CI checks reported (statusCheckRollup is empty). This PR likely has not run through CI.

---

### Overview

This is a large WIP PR (38 files, +3261/-1207) from an external contributor adding Windows support to exo. It introduces:

1. **GGUF model support** — GGUF file parsing, downloading, and model card resolution
2. **llama.cpp backend** — A new `LlamaRpcInstance` type with `llama-server` process management
3. **CUDA single-node runner** — A `CudaSingleInstance` using PyTorch + transformers on Windows
4. **GPU info gathering** — NVIDIA GPU discovery via pynvml/nvidia-smi, plumbed through state and placement
5. **Dashboard changes** — GPU info display in topology, GGUF badge, download progress bars, Llama RPC runtime selector
6. **Platform compatibility** — Windows paths, conditional `resource` import, MLX guards, Rust listen address env vars

As noted by maintainer @AlexCheema in the comments, this should be split into separate PRs for Windows-specific changes and llama.cpp support.

---

### Critical Issues

**1. Python version downgrade breaks existing users (pyproject.toml)**

```python
-requires-python = ">=3.13"
+requires-python = ">=3.12,<3.13"
```

This **locks out Python 3.13 users** entirely. The project currently requires 3.13+. The `<3.13` upper bound is too restrictive — it should be `>=3.12` without the cap if the goal is to also support 3.12 on Windows. This also changes `.python-version` from 3.13 to 3.12, which would affect all developers.

**2. MLX version downgraded without explanation**

```python
-"mlx==0.30.4; sys_platform == 'darwin'",
+"mlx==0.30.3; sys_platform == 'darwin'",
```

Downgrading MLX from 0.30.4 to 0.30.3 on macOS is unexplained and could regress existing functionality.

**3. `_wait_for_server` can loop forever (llama_cpp_runner.py)**

```python
def _wait_for_server(http_port: int) -> None:
    ...
    deadline = time.time() + timeout_seconds if timeout_seconds > 0 else None
    while deadline is None or time.time() < deadline:
        try:
            with httpx.Client(timeout=2.0) as client:
                resp = client.get(url)
                if resp.status_code == 200:
                    return
        except Exception:
            time.sleep(0.5)
```

When `EXO_LLAMA_SERVER_STARTUP_TIMEOUT_SECONDS` is unset (the default), `timeout_seconds` is 0, `deadline` is `None`, and this becomes an infinite loop. If llama-server fails to start, this runner process hangs forever. There must be a sensible default timeout (e.g., 120 seconds).

**4. `hasEnoughMemory` always returns `true` (dashboard +page.svelte)**

```typescript
function hasEnoughMemory(model: { ... }): boolean {
    return true;
}
```

This removes all memory validation in the model selector, meaning users will be able to select models that cannot possibly fit on their hardware. The previous logic (`modelSizeGB <= availableMemoryGB()`) was protective. The comment says "placement/launch will validate capacity," but by then the user has already waited through a download.

**5. `_parse_gguf_header` removed import it needs**

The diff removes `AliasChoices`, `model_validator` from model_cards.py imports and adds a **duplicate** `from enum import Enum`:

```python
from enum import Enum     # line 1
...
from enum import Enum     # line ~24 in the new code
```

This is harmless but indicates the code wasn't cleaned up.

**6. subprocess.Popen without text mode (llama_cpp_runner.py)**

```python
server_process: subprocess.Popen[str] | None = None
...
return subprocess.Popen(cmd)
```

The type annotation says `Popen[str]` but `Popen()` is called without `text=True`, so it returns `Popen[bytes]`. This is a type error.

---

### Significant Issues

**7. `_card_cache` key type inconsistency (model_cards.py)**

On `main`, `_card_cache` is `dict[ModelId, "ModelCard"]` but the PR changes it to `dict[str, "ModelCard"]`. Lookups then use both `str(model_id)` and `model_id` interchangeably, which could cause cache misses if `ModelId.__str__` doesn't round-trip cleanly.

**8. `ensureInstanceForModel` auto-launches on chat (app.svelte.ts)**

```typescript
if (!(await this.ensureInstanceForModel(modelToUse))) { ... }
```

This is called in `sendMessage`, which means sending a chat message for a GGUF model auto-launches a placement + waits 15 seconds. This is surprising UX — the user expects a message to be sent, not an instance to be launched. The 15-second timeout + 600ms polling loop in `waitForInstance` will also block the UI.

**9. `select_preferred_ip` may return `None` for a required field (placement_utils.py)**

```python
def select_preferred_ip(node_network: NodeNetworkInfo) -> str | None:
    ...
    if not node_network.interfaces:
        return None
    return min(...).ip_address
```

This returns `None` when there are no interfaces, but the caller in `placement.py` only raises if `ip is None`. If `min(...)` returns an interface with `ip_address = ""` (which could happen with malformed data), the code would proceed with an empty IP string for the RPC host.

**10. `get_local_gguf_model_cards` scans the entire models directory (model_cards.py)**

```python
async def get_local_gguf_model_cards() -> list["ModelCard"]:
    ...
    for gguf_path in EXO_MODELS_DIR.rglob("*.gguf"):
```

This is called from `get_models()` which is hit on every `/models` API call. Recursively scanning the models directory and parsing GGUF headers on every request is expensive. The results should be cached.

**11. `_build_gguf_fallback_card` uses `n_layers=1` and `hidden_size=1`**

```python
def _build_gguf_fallback_card(model_id: ModelId, size_bytes: int) -> "ModelCard":
    return ModelCard(
        ...
        n_layers=1,
        hidden_size=1,
    )
```

With `n_layers=1`, the placement logic's layer allocation (`allocate_layers_proportionally`) will assign 1 layer. In `get_shard_assignments_for_pipeline_parallel`, there's a new workaround:

```python
if total_layers < world_size:
    total_layers = world_size
```

But `n_layers=1` combined with multi-node placement will produce incorrect shard boundaries. The fallback card should either refuse multi-node placement or use a more reasonable default.

**12. Rust swarm.rs changes could break existing deployments**

```rust
let listen_addr = if let Ok(multiaddr) = env::var("EXO_LIBP2P_LISTEN_MULTIADDR")
```

Introducing env-var-controlled listen addresses in the Rust networking layer is a useful feature but changes the default networking behavior if any of these env vars happen to be set. The env var names should be documented and the parsing should include better error messages for invalid values.

---

### Minor Issues

- **`.gitignore` adds `/*.dll` and `/*.exe`**: These patterns only match at the repo root, which is intentional, but the comment referencing a specific llama.cpp release tag (b7836) will become stale.
- **`setSelectedChatModel(defaults.modelId)` on page load**: This side-effect of restoring launch defaults could override a user's manual chat model selection.
- **`ErrorChunk` with `finish_reason="error"` in llama_cpp_runner.py**: The `ErrorChunk` model likely doesn't have a `finish_reason` field based on the codebase's chunk types — this might be a runtime error.
- **`model_validator` removed from imports** in model_cards.py but the original `ModelCard` class uses `@model_validator` — if this import was still needed elsewhere in the file this would break.
- **`_read_string` in GGUF parser** has a complex fallback from u64 to u32 string length. This is fragile and could silently misparse corrupted files.
- **No tests for any of the new runners** (cuda_runner.py, llama_cpp_runner.py, entrypoint.py). No tests for GGUF parsing, GPU info gathering, or the new placement logic with GPU memory.
- **`uv.lock` changes** are massive and reflect the Python 3.12 downgrade — these will conflict with main and break the lock file for all existing contributors.

---

### What's Good

- **Clean separation of runner backends** via the new `entrypoint.py` dispatcher pattern
- **GPU info plumbing** through the state system (info_gatherer -> events -> apply -> state -> placement) follows existing architecture patterns well
- **GGUF header parser** is a solid pure-Python implementation covering all GGUF value types
- **Dashboard GPU info display** in the topology graph is well-structured with appropriate fallbacks
- **`select_preferred_ip` priority ordering** (ethernet > wifi > thunderbolt) is sensible
- **Conditional platform imports** (`resource`, MLX) are handled correctly
- **Test updates** properly thread through the new `node_gpus` parameter to existing placement tests

---

### Verdict

This PR is not ready for merge in its current form. As the maintainers noted, it should be split into focused PRs:

1. **Platform compatibility** — Windows path handling, conditional imports, Rust listen address config
2. **GPU info gathering** — pynvml/nvidia-smi discovery, state plumbing, placement awareness
3. **GGUF/llama.cpp backend** — Model card resolution, download support, LlamaRpcInstance, llama_cpp_runner
4. **CUDA single-node runner** — CudaSingleInstance, cuda_runner.py

The Python version downgrade (3.13 -> 3.12) and MLX downgrade are breaking changes that need justification or should be handled differently (e.g., widening the range to `>=3.12` without an upper cap). The infinite loop in `_wait_for_server`, the disabled memory check, and the lack of tests are the most pressing issues to address.

szmania:
> Hello, Thank you for initiating the work on this. With a few additional modifications, I've managed to get it working on macOS, Windows (AMD GPU via Vulkan), and Linux (NVIDIA GPU) with llama.cpp. There's still a lot of work to be done to make everything fully operational, such as retrieving GPU memory usage for AMD on Windows and other features. Can I create a PR to your branch so we can make a comprehensive PR together to support llama.cpp? cc @Evanev7 image

Vulkan on Windows! Awesome! I have an R9 390 myself on Win 11. Did you use llama.cpp for that? Or does exo handle all that for you? Would really love to see this merged in. Let me know you need any testing for this. Would love to get my R9 390 plugged into the greater exo ecosystem I'm running. 

Edit:

ah yep, llama.cpp `GGUF/llama.cpp backend — Model card resolution, download support, LlamaRpcInstance, llama_cpp_runner`

PR #1291, opened 2026-01-26
Remove upstream sharding + misc loading fixes

rltakashige:
## Motivation

Some models, on some configurations, would have several issues that caused the model to be stuck on loading.

## Changes

Several loading issues were with upstream mlx lm shard loading for tensor parallel.
GLM 4.7 Flash now uses GLM 4.7 Lite.
A final portion of the issues were from mlx memory not being properly released before calling mx.eval(model), causing the system to run out of memory.

### Manual Testing
Done a bunch (thanks @AlexCheema), hopefully exhaustive. 

### Automated Testing
A bunch of automated testing is imminent but not landed yet.

### DIFF
--- a/src/exo/download/download_utils.py
+++ b/src/exo/download/download_utils.py
@@ -121,11 +121,20 @@ async def ensure_models_dir() -> Path:
 
 
 async def delete_model(model_id: ModelId) -> bool:
-    model_dir = await ensure_models_dir() / model_id.normalize()
-    if not await aios.path.exists(model_dir):
-        return False
-    await asyncio.to_thread(shutil.rmtree, model_dir, ignore_errors=False)
-    return True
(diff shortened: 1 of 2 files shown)


src/exo/worker/engines/mlx/auto_parallel.py
```diff
@@ -145,6 +147,10 @@ def __call__(self, x: mx.array, *args: object, **kwargs: object) -> mx.array:
             if cache is not None:
                 cache.keys = mx.depends(cache.keys, output)  # type: ignore[reportUnknownMemberType]
 
+        output = mx.distributed.all_gather(output, group=self.group)[
+            -output.shape[0] :
+        ]  # type :ignore
```
Evanev7, line 152: is this intended to come back here? also the type ignore looks vaguely wrong

PR #1294, opened 2026-01-26
Fix download speed/ETA display for re-downloads

AlexCheema:
## Motivation

After the download verification fix, when files are re-downloaded due to upstream changes (size mismatch), the download progress displays correctly (completion %, bytes, file counts), but speed shows 0 B/s and ETA shows "--" for both overall and per-file progress.

## Changes

- Modified `on_progress_wrapper` in `src/exo/download/download_utils.py` to detect re-download scenarios
- Added re-download detection: when `curr_bytes < previous_downloaded`, the file was deleted and download restarted
- On re-download: reset `start_time` to current time and set `downloaded_this_session = curr_bytes`
- Added two tests to `test_download_verification.py` covering re-download and continuing download scenarios

## Why It Works

The bug occurred because:
1. `file_progress` is initialized with the OLD local file size (e.g., 1.5GB)
2. When `_download_file` detects size mismatch, it deletes the file and starts fresh
3. Progress callback receives small `curr_bytes` (e.g., 8KB) but compares against old size
4. `downloaded_this_session = 0 + (8KB - 1.5GB) = -1.5GB` (negative!)
5. Negative session bytes → 0 or negative speed → ETA shows "--"

The fix detects when `curr_bytes < previous_downloaded` (indicating re-download started) and resets tracking to treat it as a fresh download.

### Manual Testing

- Download a model, modify a file to change its size, restart exo, verify speed/ETA display correctly during re-download

### Automated Testing
- Added `TestProgressResetOnRedownload` class with two tests:
  - `test_progress_resets_correctly_on_redownload`: Verifies progress resets correctly when re-download starts
  - `test_progress_accumulates_on_continuing_download`: Verifies continuing downloads still accumulate correctly
- All 11 download tests pass
- Type checking (basedpyright): 0 errors
- Linting (ruff): All checks passed

### DIFF
--- a/src/exo/download/tests/test_download_verification.py
+++ b/src/exo/download/tests/test_download_verification.py
@@ -1,6 +1,8 @@
 """Tests for download verification and cache behavior."""
 
+import time
 from collections.abc import AsyncIterator
+from datetime import timedelta
 from pathlib import Path
 from unittest.mock import AsyncMock, MagicMock, patch
 
@@ -14,7 +16,8 @@ from exo.download.download_utils import (
     fetch_file_list_with_cache,
 )
 from exo.shared.types.common import ModelId
-from exo.shared.types.worker.downloads import FileListEntry
+from exo.shared.types.memory import Memory
+from exo.shared.types.worker.downloads import FileListEntry, RepoFileDownloadProgress
 
 
 @pytest.fixture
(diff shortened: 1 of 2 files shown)


AlexCheema:
After the fix a re-download shows correctly:

Screenshot 2026-01-26 at 1 46 45 PM

src/exo/download/tests/test_download_verification.py
```diff
@@ -319,3 +322,130 @@ async def test_delete_nonexistent_model(
...
+            start_time=time.time() - 10,  # Started 10 seconds ago
+        )
+
+        # Simulate the logic from on_progress_wrapper after re-download starts
+        # This is the exact logic from the fixed on_progress_wrapper
+        curr_bytes = 100_000  # 100 KB - new download just started
+        previous_progress = file_progress.get("model.safetensors")
+
+        # Detect re-download: curr_bytes < previous downloaded
+        is_redownload = (
```
rltakashige, line 363: Feel like this test is just rewriting the same logic instead of testing the intended function

PR #1312, opened 2026-01-28
remove mdns discovered peers from appearing in state

Evanev7:
## motivation
eagerly discovered peers through gossipsub were added to state. this left things looking broken from one-sided connections

## changes
the worker no longer writes topology edges from these gossipsub messages
we now strictly rely on http-discovered topology, which tends to be more reflective of the actual state of the systems connectivity

### DIFF
--- a/src/exo/main.py
+++ b/src/exo/main.py
@@ -90,7 +90,6 @@ class Node:
             worker = Worker(
                 node_id,
                 session_id,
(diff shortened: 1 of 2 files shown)


AlexCheema:
Based on conversation with @Evanev7: Just need to check the channel on the rust end that it doesn't end up blocking

PR #1317, opened 2026-01-28
feat(docker): Adds a dockerfile setup

frenchtoasters:
## Motivation

I wanted to be able to run exo in a container on linux with cuda support. Now that mlx 0.30.4 is out with better cuda support.  

## Changes

* Adds a dockerfile for setting up exo in linux
* Swaps to using mlx[cuda] on linux
* Updates mlx version to 0.30.4
* Updates jusfile to add docker build commands
* Adds a `pyproject.docker.toml` this is kinda janky but I needed it to work around the `exo_pyo3_bindings` wheel install. 

## Why It Works

I was able to load and use `mlx-community/Llama-3.2-3B-Instruct` on my DGX Spark. Getting about 36.4 tok/s with a small test prompt. 

### Manual Testing

Hardware: Linux machine with cuda 12.8+
* `docker compose build && docker compose up`


AlexCheema:
Edit: Closed by mistake.

AlexCheema:
We need to see how this compares to e2e-tests branch as there is a Dockerfile there too. I think this might handle some of the CUDA specific things.

AlexCheema:
## Code Review — PR #1317: feat(docker): Adds a dockerfile setup

**CI status:** No CI checks have run on this PR (statusCheckRollup is empty).

---

### Overview

This PR adds Docker support for running exo on Linux with NVIDIA CUDA GPUs. It introduces a multi-stage Dockerfile (Rust build, dashboard build, runtime), a docker-compose.yml, a `.dockerignore`, justfile targets, and a separate `pyproject.docker.toml` to work around workspace validation issues during the Docker build.

It also changes the **main** `pyproject.toml` from `mlx[cpu]` to `mlx[cuda13]` on Linux, and pins mlx to 0.30.4.

---

### Critical Issues

**1. `pyproject.toml` change breaks non-CUDA Linux users**

```diff
-    "mlx[cpu]==0.30.4; sys_platform == 'linux'",
+    "mlx[cuda13]==0.30.4; sys_platform == 'linux'",
```

The main `pyproject.toml` now requires `mlx[cuda13]` for *all* Linux installations, not just Docker. This means anyone running `uv sync` on a Linux machine without an NVIDIA GPU (or without CUDA 13 libraries) will get broken dependencies. The previous `mlx[cpu]` default was correct for the general case. This change should be reverted; the Docker-specific `pyproject.docker.toml` already handles the CUDA variant for the container build.

**2. `pyproject.toml` version pins diverge from `main`**

The PR's `pyproject.toml` diff is based on an older version of main. Current main has:
- `mlx; sys_platform == 'darwin'` (unpinned on macOS, using custom git fork)
- `mlx[cpu]==0.30.6; sys_platform == 'linux'`
- `mlx-lm==0.30.6`
- Additional deps: `msgspec>=0.19.0`, `zstandard>=0.23.0`

The PR pins mlx to `0.30.4` and `mlx-lm` (unpinned), removes the `exo_pyo3_bindings` workspace dependency line, and is missing the `msgspec` and `zstandard` dependencies. This PR needs to be rebased onto current main.

**3. `pyproject.docker.toml` is a full copy that will drift**

The 122-line `pyproject.docker.toml` is a near-complete copy of `pyproject.toml` with the workspace section removed. This will silently go stale every time someone updates dependencies, scripts, tool config, etc. in the real `pyproject.toml`. Consider instead:
- Using a build argument or environment variable to control workspace behavior
- Using a small script in the Dockerfile to strip the workspace section at build time (e.g., `python -c "import tomlkit; ..."`)
- Or at minimum, adding a prominent comment in both files linking them together

---

### Significant Issues

**4. Hardcoded `LD_LIBRARY_PATH` is fragile**

```dockerfile
ENV LD_LIBRARY_PATH="/app/.venv/lib/python3.13/site-packages/nvidia/cu13/lib:..."
```

This path assumes specific package layout from nvidia pip packages. If nvidia package versions change their internal directory structure, this will break silently. Consider using a dynamic approach:

```dockerfile
RUN python3.13 -c "import nvidia; print(nvidia.__path__)" ...
```

Or at least document why these specific paths are needed.

**5. `uv sync` then `uv pip install . --no-deps` ordering concern**

```dockerfile
RUN uv venv --python python3.13 \
    && uv sync \
    && uv pip install /tmp/wheels/*.whl \
    && uv pip install . --no-deps \
    && apt-get update && apt-get install -y python3.13-dev && rm -rf /var/lib/apt/lists/* \
    && uv pip install "mlx[cuda13]==0.30.4" \
    && rm -rf /tmp/wheels
```

- `uv sync` will install `mlx[cuda13]` from pyproject.docker.toml, but then the last line reinstalls `mlx[cuda13]==0.30.4` again. This is redundant (or the sync might install a different version, then get overwritten).
- `python3.13-dev` is installed via `apt-get` in the middle of the pip install chain. This should be moved earlier in the Dockerfile (or into the initial `apt-get` block) for clarity and better layer caching.
- Mixing `apt-get` inside a `RUN` that's doing pip installs is messy and breaks if cached layers change.

**6. `uv.lock` is not updated**

The PR changes dependencies in `pyproject.toml` but does not include an updated `uv.lock`. The lockfile on main currently resolves `mlx[cpu]==0.30.6` for Linux. Changing to `mlx[cuda13]==0.30.4` without updating the lockfile will cause `uv sync` to fail or produce inconsistent results.

**7. Runtime image uses `cudnn-devel` — likely oversized**

```dockerfile
FROM nvidia/cuda:12.8.0-cudnn-devel-ubuntu24.04 AS runtime
```

The runtime stage uses the full `devel` image (several GB). If MLX only needs JIT kernel compilation (which needs CUDA headers/nvcc), consider using `nvidia/cuda:12.8.0-devel-ubuntu24.04` without cudnn-devel, or test if `runtime` + specific headers is sufficient. The builder already uses the plain `devel` image.

---

### Minor Issues

**8. `docker-compose.yml` uses `gpus: all` (compose v2 syntax only)**

The `gpus: all` shorthand works with Docker Compose v2 but not v1. Since this is a new feature, it's probably fine, but it's worth noting in documentation or the PR description.

**9. `.dockerignore` excludes `.git/` but the Dockerfile doesn't use git**

This is fine — just confirming it's intentional. The Rust build copies `Cargo.toml`/`Cargo.lock` and `rust/` directly, so `.git/` exclusion is correct and reduces build context size.

**10. `TIKTOKEN_ENCODINGS_BASE` environment variable**

The Dockerfile sets `TIKTOKEN_ENCODINGS_BASE=/app/tiktoken_cache` after downloading the vocab file. The standard environment variable for tiktoken's cache is `TIKTOKEN_CACHE_DIR`, not `TIKTOKEN_ENCODINGS_BASE`. Verify this is the correct env var name for the version of tiktoken being used.

**11. `mlx-lm` source differs from main**

The `pyproject.docker.toml` uses `mlx-lm = { git = "https://github.com/ml-explore/mlx-lm", branch = "main" }` while main uses `mlx-lm==0.30.6` (pinned). This could introduce unexpected behavior differences between Docker and non-Docker deployments.

**12. Missing `bench` workspace member**

The real `pyproject.toml` has `tool.uv.workspace.members` including `"bench"`. The docker version has no workspace at all, which is intentional, but the `pyproject.docker.toml` is also missing the `exo-master` and `exo-worker` entry points that exist in the real one (though those may have been removed from main since this PR was opened).

---

### What's Good

- **Multi-stage build** is well-structured: Rust builder, dashboard builder, runtime. This is the right approach for minimizing image size and build time.
- **Pre-downloading tiktoken vocab** is a thoughtful touch for air-gapped or restricted-network deployments.
- **Volume mounts** in docker-compose.yml for exo data, cache, config, and HuggingFace cache are well-chosen and prevent re-downloading models on container restart.
- **`.dockerignore`** is thorough and well-organized.
- **justfile targets** (`docker-build`, `docker-run`, `docker-shell`) provide convenient shortcuts.
- The overall approach of building `exo_pyo3_bindings` as a wheel in a separate stage is sound.

---

### Verdict

The PR addresses a real need (CUDA Docker support) and the Dockerfile structure is solid. However, there are **blocking issues** that need to be resolved before merge:

1. **Revert the `mlx[cuda13]` change in the main `pyproject.toml`** — this breaks non-CUDA Linux installs. Keep `mlx[cpu]` as the default; Docker handles its own deps.
2. **Rebase onto current main** — dependency versions and the dependency list have diverged significantly (mlx 0.30.4 vs 0.30.6, missing msgspec/zstandard, mlx-lm pinning, custom mlx fork on darwin).
3. **Address the `pyproject.docker.toml` drift risk** — either automate its generation or add a clear maintenance strategy.
4. **Update `uv.lock`** to match the pyproject.toml changes.

After those are addressed, the remaining issues are straightforward fixes.

PR #1318, opened 2026-01-28
Add minimax and fix qwen sharding strategies

rltakashige:
## Motivation

MiniMax tensor sharding does not provide equivalent outputs to running it as a single node because RMSNorm weights cannot be split without affecting the output.

Qwen3Next sharding was broken, and something with Qwen3MoE was likely changed upstream, as several variables no longer exist.

This also ballooned into fixing prefix caching for non-standard models as Qwen3Next was behaving weirdly.

### Manual Testing
Worked for a 8 hour long eval at the same performance and a more similar completion/reasoning token distribution.

### DIFF
--- a/bench/exo_bench.py
+++ b/bench/exo_bench.py
@@ -431,7 +431,12 @@ def main() -> int:
     ap.add_argument(
         "--skip-pipeline-jaccl",
         action="store_true",
-        help="Pipeline jaccl is often pointless, skip by default",
+        help="Skip pipeline+jaccl placements, as it's often pointless.",
+    )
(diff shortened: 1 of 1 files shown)


juoboinshurima543-sys:
That’s clear thanks

JakeHillion:
This breaks Kimi-K2.5 with Tensor+RDMA on 2 machines. I think it breaks Kimi K2.5 without those conditions, but it's not small enough to fit on one machine for me to find out :)

Evanev7 (approved):
👍 :o)

Issue #918, opened 2025-12-18
Run custom MLX models


arno4000:
Is it possible to run custom MLX models from HuggingFace like for example https://huggingface.co/mlx-community/gpt-oss-20b-MXFP4-Q8? If not, I think this would be a feature many would be interested in. Particularly now since Apple released RDMA support.

Evanev7:
Probably the single feature we want to ship ASAP - Christmas notwithstanding. It's totally possible right now through model_cards.py, while some models require more custom support for tensor sharding, gpt-oss-20b is (I believe) not one of them.

nightguarder:
Hi, I’ve been sort of implementing this on my own fork of exo. How can I help with this?

Evanev7:
Testing is perhaps the main thing - add a model card to src/exo/shared/model_cards/model_cards.py and see what breaks. As a heads up @rltakashige is also working on this but YMMV

We had some version incompatibility between gpt oss and kimi k2's tokenizers - seems like the release candidate of transformers isnt quite there yet.

nightguarder:
Ok I will try my best to test out the smaller models. I don’t have a powerful hardware to test out the bigger ones you mentioned. 
#937

rltakashige:
GPT OSS and GLM sharding support is around the corner, as well as a few more types of Qwen models.

There is a transformers version incompatibility with Ministral3 models, which @Evanev7 pointed out.

edixongomez98:
hola

rltakashige:
Been looking through old issues... You can run custom models in EXO v1.0.68 and the now released v1.0.69!

PR #937, opened 2025-12-20
Integrating custom models for Exo

nightguarder:
# Custom MLX Models Support - Issue #918

## Motivation
Fixes Issue #918: Enable users to run custom MLX based models from `mlx-community` on Hugging Face without manual code updates.

# What Changed

### 1. Frontend UI for Custom Models
**Commit: “[Add custom models to dashboard](https://github.com/exo-explore/exo/pull/937/commits/979d122d81772622dda33adccc449b4dc4366989)"**
- Added "Custom Models" section to downloads page with HuggingFace model ID input
- Implemented "Download Model" button triggering model Download into the exo HOME folder
- SAFE registration of the custom model via separate `/custom_models` API

### 2. Tests  
**Commit: “[Integrate new test](https://github.com/exo-explore/exo/pull/937/commits/e3048d89d2eb7b7f89b872a0fe3dd7cbc6e56b80)”**
`@pytest` are available in `src/exo/worker/tests/test_custom_model.py` folder
with future possibility to run them on CI/CD Pipeline on Github file (`.github/workflows/custom_models.yml`)

### 3. Persistent Storage & Model Registration
**Commit: "Persist storage for custom models"**
- Fixed [resolve_model_meta()](file:///Users/cyrils/Developer/GPTExo/exo-mac/src/exo/master/api.py#89-107) to check both short_id keys and full model_id values
- Enabled custom model registration to `~/.exo/custom_models.json` during download
- Models reload automatically on EXO restart from persistent storage
- Prettified name for custom models in [pretty_name](file:///Users/cyrils/Developer/GPTExo/exo-mac/src/exo/shared/models/model_cards.py#564-607) key

### 4. Safe Downloading logic
**Commit: “[SAFE model registration](https://github.com/exo-explore/exo/pull/937/commits/c891a74364bcb69f026016f5c8bb803bb3bb8df8)”**
- Implemented "lazy loading" logic. Runners are now skipped for `download_only` instances until a task is received, while model downloads continue in the background.
- Added automatic model registration. Custom models are now registered in `~/.exo/custom_models.json` immediately when the download starts, ensuring they are recognized by the system right away.

## Why It Works

This implementation enables dynamic custom model loading without requiring manual modifications to [model_cards](file:///Users/cyrils/Developer/GPTExo/exo-mac/src/exo/shared/models/model_cards.py#690-694).

### Users can:
- Download any `mlx-community` based model via the dashboard from HuggingFace 
- Have models persist across restarts
- Test out the their model once it loads

[...]

Issue #801, opened 2025-03-23
Is it possible for importing model in to exo node?


hefish:
i can download model in tinychat webui . 
i find in exo/exo/models.py  there is a model list which i can use. 
and then i think , is it possible to add a new model into models.py ?

chunxiao365:
I also want to know

TanishkBansode:
if it supports tinygrad or mlx, yes

Evanev7:
Will get on this soon

rltakashige:
Marking this feature as completed! 

You are able to add custom models from HuggingFace through our model selector (and even write custom model cards for them to enable features such as vision capabilities and thinking toggles.

PR #1323, opened 2026-01-29
Custom models support LATEST FORK

nightguarder:
## Motivation

Fixes Issue https://github.com/exo-explore/exo/issues/918: Enable users to run custom MLX based models from mlx-community on Hugging Face without manual code updates.
**Sidenote:**
This is a new _latest_ fork based on `main` branch of `exo-explore/exo`. The previous PR (#937) got way too complicated than it needed to be. 
#### 1. New UI for Downloads
- Added "Custom Models" section to downloads page with HuggingFace model ID input
- Implemented "Download Model" button triggering model Download into the exo HOME folder
- SAFE registration of the custom model via separate /custom_models API
#### 2. Persistent Storage & Model Registration
- Model is registered via separate API endpoint
- All the custom models are saved to EXO home PATH DIR —> `~/.exo/custom_models.json`
- Models reload automatically on EXO restart from persistent storage
- Prettified name for custom models in pretty_name key
- Custom models are now registered in ~/.exo/custom_models.json immediately when the download starts, ensuring they are recognized by the system right away.
#### 3. Runnable fix
- Refactored the model download in `plan.py` to trigger based on assigned instance shards rather than active runners
- allowing downloads to start even before a runner is initialized for a node.
#### 4. Tests
- @pytest is available in src/exo/worker/tests/test_custom_model.py dir 

## Closes

Closes a lot of Issues: #918 #801 #937 
### Manual Testing
- Already tested by various users in the old PR 
### Automated Testing
Integration test: `src/exo/worker/tests/test_custom_model.py`
### Changed files: 
**Frontend**
- `dashboard/src/lib/components/CustomModel.svelte`
- `dashboard/src/lib/stores/app.svelte.ts`
- `dashboard/src/routes/downloads/+page.svelte`
- `dashboard/src/routes/+page.svelte`

**Backend**
- `src/exo/download/impl_shard_downloader.py`
- `src/exo/master/api.py`
- `src/exo/shared/models/model_cards.py`
- `src/exo/shared/types/api.py`

### DIFF
--- a/dashboard/src/lib/stores/app.svelte.ts
+++ b/dashboard/src/lib/stores/app.svelte.ts
@@ -1173,7 +1173,9 @@ class AppStore {
         this.models = (data.data || []).map((m: any) => ({
           modelId: m.id,
           prettyName: m.name,
-          storageSize: { inBytes: (m.storage_size_megabytes || 0) * 1024 * 1024 },
+          storageSize: {
+            inBytes: (m.storage_size_megabytes || 0) * 1024 * 1024,
+          },
           nLayers: m.n_layers || 0,
           hiddenSize: m.hidden_size || 0,
           supportsTensor: m.supports_tensor || false,
--- a/src/exo/download/impl_shard_downloader.py
+++ b/src/exo/download/impl_shard_downloader.py
@@ -149,7 +149,7 @@ class ResumableShardDownloader(ShardDownloader):
                 if card.model_id == shard.model_card.model_id:
                     known = True
                     break
-        
+
         if not known:
             repo_id = shard.model_card.model_id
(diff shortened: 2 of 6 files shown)


nightguarder:
Just a sidenote. I really don’t like the Download logic and in what phase it currently is. 

1. Why are we showing all the (non)-downloaded models? This is not a good solution.
2. The annoying `[Download Debug] Current downloads:` is running in a effect loop.

Honestly this should not be in a production-ready branch. I can quickly solve this and suggest changes.

AlexCheema:
Agreed the download code is not great and needs a rework.

> I can quickly solve this and suggest changes

Go for it, it would be a welcome contribution.

nightguarder:
Merged with #1368 and #1369

PR #1324, opened 2026-01-29
Leo/assert nix uv mlx version match

rltakashige:
## Changes

Assertion fails when mlx version is changed.

Fixed failing mlx version number.

Screenshot 2026-01-29 at 13 54 18


djhdusjbd-eng:
Acknowledged

PR #1328, opened 2026-01-29
Ciaran/image non streaming

ciaranbor:
## Motivation

The dashboard UI attempted to parse all image generation responses as SSE streams, even when streaming was disabled. This broke non-streaming image generation.

## Changes

- Parse JSON responses directly when not streaming, use SSE parser only when stream=true AND partialImages > 0
- explicitly disable partial images when not streaming

## Why It Works

Both API and dashboard now use the same condition (stream && partialImages > 0) to determine response format, ensuring correct parsing.

### Manual Testing

Non-streamed image generation results appear in the UI. Streamed image generation still works

### DIFF
--- a/src/exo/worker/engines/image/generate.py
+++ b/src/exo/worker/engines/image/generate.py
@@ -98,8 +98,8 @@ def generate_image(
 
     partial_images = (
         task.partial_images
-        if task.partial_images is not None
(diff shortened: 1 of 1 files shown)


AlexCheema:
On hold waiting for #1167 since that changes a bunch of API stuff.

PR #1342, opened 2026-01-30
fix skipping logic in worker plan

Evanev7:
the worker plan function had some skipping logic missing, leading to double-submitting tasks.


twubndgsijlc-netizen:
That checks out

Op vr 30 jan 2026 om 14:19 schreef Evan Quiney ***@***.***>:

PR #1344, opened 2026-01-30
fix: handle unclosed tool calls and GLM arg parsing edge cases

AlexCheema:
## Motivation

Tool-call requests can hang indefinitely when `max_tokens` truncates generation mid-tool-call.

## Reproduction

1. Send a chat completion with `tools` and a low `max_tokens` (e.g. 65) to Qwen3-0.6B
2. Model generates `<think>...</think>` then starts `<tool_call>` but `max_tokens` cuts it off before `</tool_call>`
3. **Before this fix:** `parse_tool_calls` buffers tokens after `<tool_call>`, generator exhausts, buffered tokens (including `finish_reason`) are silently dropped → stream hangs forever
4. **After this fix:** buffered tokens are flushed as regular text with `finish_reason` propagated → response returns normally with `finish_reason: "length"`

Confirmed with fresh local testing: 4 unclosed tool call flushes triggered in a single session. Also confirmed via production logs from Jan 29 (2 occurrences).

## Changes

1. **`parse_tool_calls` unclosed tool call flush** — when the generator exhausts inside an open `<tool_call>` block, flush buffered tokens as regular text and propagate `finish_reason`
2. **GLM regex fix** — match literal `\n` (not escaped `\\n`) between arg tags; handle missing `</arg_value>` via lookahead
3. **7 new unit tests** for `parse_tool_calls` covering unclosed, closed, passthrough, and failed-parse scenarios

## Why It Works

- `parse_tool_calls` now has a post-loop check: if `in_tool_call` is still true, it yields the buffered text with the tracked `finish_reason` instead of silently dropping it
- The GLM regex now matches real-world output where newlines appear between tags and `</arg_value>` may be absent

### Manual Testing
- Qwen3-0.6B-4bit with `tools` + various `max_tokens` values (61-75)
- Confirmed responses return with `finish_reason: "length"` instead of hanging
- Log output shows `"generator exhausted inside unclosed tool call, flushing buffered text"`

### Automated Testing
- 7 new tests in `test_parse_tool_calls.py`
- Full test suite passes (`uv run pytest`)

### DIFF
--- a/src/exo/worker/runner/runner.py
+++ b/src/exo/worker/runner/runner.py
@@ -16,8 +16,9 @@ from openai_harmony import (  # pyright: ignore[reportMissingTypeStubs]
 )
 from pydantic import ValidationError
 
-from exo.shared.constants import EXO_MAX_CHUNK_SIZE
+from exo.shared.constants import EXO_MAX_CHUNK_SIZE, EXO_TRACING_ENABLED
 from exo.shared.models.model_cards import ModelId, ModelTask
+from exo.shared.tracing import clear_trace_buffer, get_trace_buffer
 from exo.shared.types.api import ImageGenerationStats
 from exo.shared.types.chunks import ErrorChunk, ImageChunk, TokenChunk, ToolCallChunk
 from exo.shared.types.common import CommandId
@@ -27,6 +28,8 @@ from exo.shared.types.events import (
     RunnerStatusUpdated,
     TaskAcknowledged,
     TaskStatusUpdated,
+    TraceEventData,
+    TracesCollected,
 )
 from exo.shared.types.tasks import (
     ConnectToGroup,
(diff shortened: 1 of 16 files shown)


src/exo/worker/runner/runner.py
```diff
@@ -788,7 +830,7 @@ def patch_glm_tokenizer(tokenizer: TokenizerWrapper):
 
     _func_name_regex = re.compile(r"^(.*?)<arg_key>", re.DOTALL)
     _func_arg_regex = re.compile(
-        r"<arg_key>(.*?)</arg_key>(?:\\n|\s)*<arg_value>(.*?)</arg_value>",
+        r"<arg_key>(.*?)</arg_key>(?:\n|\s)*<arg_value>(.*?)(?:</arg_value>|(?=<arg_key>)|$)",
```
rltakashige, line 842: Understood about the termination, but still think `|(?=<arg_key>)` is unnecessary as per my comment on the other pr.

Also gpt oss still needs handling

PR #1349, opened 2026-01-31
fix GIL race condition in PyO3 networking bindings

srikalyan:
## Summary

- Fix GIL race condition causing panics on Linux startup
- Use pure Rust types in Tokio tasks instead of Python-wrapped types (`PyConnectionUpdate` with `PyPeerId`)
- Convert to Python types only at the receive boundary (in `#[pymethods]`)

## Problem

The networking bindings were holding Python objects (`PyPeerId`) inside `mpsc::Sender` channels that were moved into Tokio tasks. When these senders were dropped from Tokio worker threads, the `AsyncioWaker` tried to acquire the Python GIL, causing:

- **ubuntu**: `AlreadyAttached` panic on startup

## Solution

1. Added pure Rust `ConnectionUpdate` and `ConnectionUpdateType` structs (no `Py*` wrappers)
2. Changed the channel to use `ConnectionUpdate` instead of `PyConnectionUpdate`
3. Convert to Python types only in `connection_update_recv()` and `connection_update_recv_many()` methods, which run with the GIL held

## Test plan

- [x] Verified exo starts without crash on Linux (previously crashed immediately)
- [x] All 132 existing tests pass
- [x] Ran exo for 30+ seconds without GIL panics

### DIFF
--- a/rust/exo_pyo3_bindings/src/networking.rs
+++ b/rust/exo_pyo3_bindings/src/networking.rs
@@ -13,7 +13,7 @@ use crate::pylibp2p::ident::{PyKeypair, PyPeerId};
 use libp2p::futures::StreamExt as _;
 use libp2p::gossipsub::{IdentTopic, Message, MessageId, PublishError};
 use libp2p::swarm::SwarmEvent;
-use libp2p::{gossipsub, mdns};
+use libp2p::{gossipsub, mdns, PeerId};
 use networking::discovery;
 use networking::swarm::create_swarm;
 use pyo3::prelude::{PyModule, PyModuleMethods as _};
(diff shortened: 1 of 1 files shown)


Evanev7:
The thing that doesn't track is that your pr says that you ran EXO without gil panics - the panics occur on shutdown (as you identified) which I cannot see any evidence of this addressing after my own testing.

srikalyan:
Thanks for the review @Evanev7!

To clarify what this PR addressed:

## Before the fix (Linux)

EXO crashed **immediately on startup** - the application couldn't even start. The crash occurred during `NodeConfig.gather()` in the info_gatherer:

```
pyo3_runtime.PanicException: called `Result::unwrap()` on an `Err` value: AlreadyAttached
```

**Stack trace:**
```
File "src/exo/utils/info_gatherer/info_gatherer.py", line 340, in run
    nc = await NodeConfig.gather()

thread '<unnamed>' panicked at pyo3-0.25.0/src/internal/state.rs:62:13

Stack frames:
  - <pyo3::internal::state::AttachGuard>::attach
  - <pyo3::coroutine::waker::AsyncioWaker as alloc::task::Wake>::wake_by_ref
  - core::ptr::drop_in_place::<tokio::sync::mpsc::bounded::Sender<PyConnectionUpdate>>
  - <exo_pyo3_bindings::networking::PyNetworkingHandle>::py_new::{closure#1}
```

**Root cause:** The `networking_task` spawned in `py_new()` held `mpsc::Sender<PyConnectionUpdate>` which contained `PyPeerId` (a Python object wrapper). When Tokio worker threads dropped or interacted with this sender, `AsyncioWaker` tried to acquire the GIL from a non-Python thread, causing the `AlreadyAttached` panic.

## After the fix

EXO **starts and runs normally on Linux**. I ran it for 10+ minutes, loaded models (Qwen3-30B), and processed chat completion requests without the startup crash.

## What this doesn't fix

You're right that **shutdown panics** ("Python interpreter is not initialized" during `Py_Finalize`) are a separate issue. That occurs when Python is already finalizing and Tokio threads try to wake coroutines. This PR doesn't address that.

| Scenario | Before | After |
|----------|--------|-------|
| Linux startup | ❌ Immediate crash (`AlreadyAttached`) | ✅ Works |
| Shutdown | ❌ Panic during `Py_Finalize` | ❌ Still panics (not addressed) |

The startup crash was blocking Linux users entirely. I understand you're planning a networking rewrite - happy to have this closed if you prefer to handle it there!

Evanev7:
man i hate talking to peoples llms but i love how this stupid pyo3 bug catches them all outta wack

srikalyan:
Dude! Not at all cool. You are maintainer of a public github repo that is helping communities to serve their own models/llms on commodity hardware. It was not working on my Ubuntu servers and I can keep the patch locally but thought it would avoid headaches for others. Don't see the point of you being rude, my only point of follow up is to see if my setup is wrong.

Evanev7:
fair, i shouldn't have looked at work at 11pm. the bug you ran into was fixed by #1353 at any rate

PR #1354, opened 2026-02-02
migrate model cards to .toml files

Evanev7:


Evanev7:
@JakeHillion should be ready to go now

Evanev7:
tested with maccery o7

PR #1361, opened 2026-02-03
Ciaran/parallel cfg

ciaranbor:
## Motivation

Enable parallel classifier-free guidance (CFG) for Qwen image models. CFG requires two forward passes (positive/negative prompts) - this allows them to run on separate nodes simultaneously, reducing latency.

## Changes

  - Added uses_cfg flag to ModelCard to identify CFG-based models
  - Extended PipelineShardMetadata with CFG topology fields (cfg_rank, cfg_world_size, peer device info)
  - Updated placement to create two CFG groups with reversed ordering (places CFG peers as ring neighbors)
  - Refactored DiffusionRunner to process CFG branches separately with exchange at last pipeline stage
  - Added get_cfg_branch_data() to PromptData for single-branch embeddings
  - Fixed seed handling in API for distributed consistency
  - Fixed image yield to only emit from CFG rank 0 at last stage
  - Increased num_sync_steps_factor from 0.125 to 0.25 for Qwen

## Why It Works

  - 2 nodes + CFG: Both run all layers, process different CFG branches in parallel
  - 4+ even nodes + CFG: Hybrid - 2 CFG groups × N/2 pipeline stages
  - Odd nodes or non-CFG: Falls back to pure pipeline parallelism
 
 Ring topology places CFG peers as neighbors to enable direct exchange.

### Manual Testing

Verified performance gain for Qwen-Image for 2 node and 4 node cluster. Non-CFG models still work

### Automated Testing
 
Added tests in test_placement_utils.py covering 2-node CFG parallel, 4-node hybrid, odd-node fallback, and non-CFG pipeline modes.

### DIFF
--- a/src/exo/master/placement_utils.py
+++ b/src/exo/master/placement_utils.py
@@ -75,17 +75,60 @@ def allocate_layers_proportionally(
     return result
 
 
+def _validate_cycle(cycle: Cycle) -> None:
+    if not cycle.node_ids:
+        raise ValueError("Cannot create shard assignments for empty node cycle")
+
+
+def _compute_total_memory(
+    node_ids: list[NodeId],
+    node_memory: Mapping[NodeId, MemoryUsage],
+) -> Memory:
+    total_memory = sum(
+        (node_memory[node_id].ram_available for node_id in node_ids),
+        start=Memory(),
+    )
+    if total_memory.in_bytes == 0:
+        raise ValueError("Cannot create shard assignments: total available memory is 0")
+    return total_memory
+
+
+def _allocate_and_validate_layers(
+    node_ids: list[NodeId],
+    node_memory: Mapping[NodeId, MemoryUsage],
+    total_memory: Memory,
+    model_card: ModelCard,
+) -> list[int]:
+    layer_allocations = allocate_layers_proportionally(
+        total_layers=model_card.n_layers,
+        memory_fractions=[
+            node_memory[node_id].ram_available.in_bytes / total_memory.in_bytes
+            for node_id in node_ids
+        ],
+    )
+
+    total_storage_bytes = model_card.storage_size.in_bytes
+    total_layers = model_card.n_layers
+    for i, node_id in enumerate(node_ids):
+        node_layers = layer_allocations[i]
+        required_memory = (total_storage_bytes * node_layers) // total_layers
+        available_memory = node_memory[node_id].ram_available.in_bytes
+        if required_memory > available_memory:
+            raise ValueError(
+                f"Node {i} ({node_id}) has insufficient memory: "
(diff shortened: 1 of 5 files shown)


src/exo/master/placement_utils.py
```diff
@@ -116,24 +131,62 @@ def get_shard_assignments_for_pipeline_parallel(
...
+        if is_first_stage:
+            prev_pipeline_device = None
+        else:
+            prev_pipeline_device = cfg_pipeline_to_device[(cfg_rank, pipeline_rank - 1)]
+
+        if is_last_stage and use_cfg_parallel:
+            other_cfg_rank = 1 - cfg_rank
+            cfg_peer_device = cfg_pipeline_to_device[(other_cfg_rank, pipeline_rank)]
+        else:
+            cfg_peer_device = None
```
AlexCheema: You weren't joking about placement getting more complex.

Can we separate this out into a separate function? It's getting too big and nasty

src/exo/shared/types/worker/shards.py
```diff
@@ -51,17 +52,70 @@ def __hash__(self) -> int:
         )
 
 
+@final
 class PipelineShardMetadata(BaseShardMetadata):
```
AlexCheema, line 56: Wouldn't it be cleaner to make a new type for CFG instead of piling more features on Pipeline?
It might simplify the rest of the code too if we can just treat this separately (potentially?)

src/exo/master/placement_utils.py
```diff
@@ -94,20 +94,35 @@ def get_shard_assignments_for_pipeline_parallel(
     runner_to_shard: dict[RunnerId, ShardMetadata] = {}
     node_to_runner: dict[NodeId, RunnerId] = {}
 
+    # Determine CFG parallelism topology
+    # CFG parallel only for even node counts with CFG models (2+ nodes)
+    use_cfg_parallel = model_card.uses_cfg and world_size >= 2 and world_size % 2 == 0
+    cfg_world_size = 2 if use_cfg_parallel else 1
+    pipeline_world_size = world_size // cfg_world_size
```
AlexCheema: So CFG models don't support normal pipeline parallel anymore?
See prev comment, maybe we should make a separate parellism type instead of piling it into PipelineParalell

AlexCheema (changes requested):
I would think about how we can make this cleaner - one suggestion is to not use PipelineParallelism and make a new type for CFG.

src/exo/shared/types/worker/shards.py
ciaranbor: Yeah, this makes sense

ciaranbor:
An improvement I think, at least

PR #1363, opened 2026-02-03
Reduce reliance on internet

rltakashige:
## Motivation

Offline users currently have to wait for every retry to fail before being able to launch a model.
For users that restart clusters often or share API keys between devices, we also spam HuggingFace with downloads every 5 minutes.
These issues are caused by _emit_existing_download_progress being inefficient.

## Changes

- Only query HuggingFace once while EXO is running (assumption being that a change should only be reflected on a new EXO session)
- Only query HuggingFace when there is an internet connection (polling connectivity every 10 seconds)
- Request download progress if we switch from no connectivity -> connected to reduce the wait.
- Reduce download progress sleep as it's no longer expensive (queries cache most of the time).
- Reduce retries as 30 is way too many.

### Manual Testing
Manually tested the behaviour.

### Automated Testing
None, should I add any? We do have some tests for this folder, but they are probably not too helpful.

### DIFF
--- a/src/exo/download/download_utils.py
+++ b/src/exo/download/download_utils.py
@@ -284,7 +284,7 @@ def create_http_session(
     else:
         total_timeout = 1800
         connect_timeout = 60
-        sock_read_timeout = 1800
+        sock_read_timeout = 60
         sock_connect_timeout = 60
 
     ssl_context = ssl.create_default_context(


src/exo/download/coordinator.py
```diff
@@ -60,10 +61,31 @@ def __post_init__(self) -> None:
...
+        )
+
+    async def _check_internet_connection(self) -> None:
+        first_connection = True
+        while True:
+            await asyncio.sleep(10)
+            self._test_internet_connection()
+            if first_connection and self.shard_downloader.internet_connection:
+                first_connection = False
+                self._tg.start_soon(self._emit_existing_download_progress)
```
JakeHillion, line 94: I think we should return from this once we get a success, and then use errors elsewhere in downloads to imply we don't have Internet.

rltakashige: added a callback (this code is so messy...)

PR #1365, opened 2026-02-03
Fix tool call results persisting across stateless requests

### LINKED ISSUE #1362: [BUG] Tool Call Results Persist Across Independent Stateless Requests (v1.0.67 / gpt-oss-120b-MXFP4-Q8)
## Describe the bug

There appears to be a state leak where EXO "remembers" the result of a function call from a previous request. 

In a stateless `/v1/chat/completions` setup, each request should be independent. However, after successfully providing a tool result to the model, subsequent identical requests (without the tool history) bypass the tool-calling mechanism and directly return the final answer. This behavior persists until the instance is deleted and re-launched.

## To Reproduce

Model : `mlx-community/gpt-oss-120b-MXFP4-Q8`
Sharding : Pipeline
Instance Type : MLX Ring

I have attached two scripts, **`p1.sh`** and **`p2.sh`**, to facilitate testing.

### 1. Initial Request (using `p1.sh`)
Execute `p1.sh` to request `2+2` using the `calculator` tool.

**Expected & Actual Result:**
The model correctly identifies the need for a tool and returns a `tool_calls` response:
```json
{
  "id": "b34fb07b-5176-4ce6-9703-1e36200748ee",
  "choices": [
    {
      "index": 0,
      "message": {
        "role": "assistant",
        "content": "<think>We need to use calculator tool. Use function.</think>",
        "tool_calls": [
          {
            "id": "36415a46-9a79-440d-b2e0-4ec57ed917d8",
            "type": "function",
            "function": {
              "name": "calculator",
              "arguments": "{\n  \"expression\": \"2+2\"\n}"
            }
          }
        ]
      },
      "finish_reason": "tool_calls"
    }
  ]
}
```

### 2. Provide Tool Result (using `p2.sh`)
Execute `p2.sh`, which includes the full history (User message, Assistant tool call, and Tool output).

**Actual Result:**
The model correctly returns the final answer:
```json
{
  "choices": [
    {
      "index": 0,
      "message": {
        "role": "assistant",
        "content": "The result of 2 + 2 is **4**."
      },
      "finish_reason": "stop"
    }
  ]
}
```

### 3. Re-run Initial Request (using `p1.sh` again)
Run `p1.sh` for the second time (no tool history provided).

## Expected behavior

The model should trigger `tool_calls` again, identical to Step 1.

## Actual behavior

The model bypasses the tool call and returns the final answer directly, as if it "remembered" the state from Step 2:
```json
{
  "choices": [
    {
      "index": 0,
      "message": {
        "role": "assistant",
        "content": "The result of 2 + 2 is **4**.",
        "tool_calls": []
      },
      "finish_reason": "stop"
    }
  ]
}
```

## Environment

- macOS Version:26.2
- EXO Version:1.0.67
- Hardware:M3 Ultra Mac Studio (512GB RAM)
- Interconnection:Single Host

## Additional context

* **Persistence**: Repeatedly executing `p1.sh` continues to return the direct answer without triggering the tool.
* **Workaround**: Deleting the instance and re-launching it "clears" this memory. After a fresh launch, `p1.sh` will correctly trigger a tool call again (but only until a tool result is provided once).

---

## Appendix: Test Scripts

`p1.sh`: This script sends a **stateless initial request**. It asks "What is 2+2?" and provides the `calculator` tool definition. Under normal conditions, the model should always return a `tool_calls` object to request the calculation.

### p1.sh
```bash
curl -N -X POST http://localhost:52415/v1/chat/completions \
  -d '{
  "model": "mlx-community/gpt-oss-120b-MXFP4-Q8",
  "messages": [
    {
      "role": "user",
      "content": "What is 2+2? Use the calculator tool."
    }
  ],
  "stream": false,
  "tools": [
    {
      "type": "function",
      "function": {
        "name": "calculator",
        "description": "Perform basic arithmetic",
        "parameters": {
          "type": "object",
          "properties": {
            "expression": {
              "type": "string",
              "description": "Math expression"
            }
          },
          "required": [
            "expression"
          ]
        }
      }
    }
  ]
}' | jq
```

`p2.sh`: This script sends a **request with conversation history**. It includes the user's question, the assistant's previous tool call, and the tool's response (`2+2 = 4`). This is used to complete the function-calling cycle.

### p2.sh
```bash
curl -N -X POST http://localhost:52415/v1/chat/completions \
  -d '{
  "model": "mlx-community/gpt-oss-120b-MXFP4-Q8",
  "messages": [
    {
      "role": "user",
      "content": "What is 2+2? Use the calculator tool."
    },
    {
      "role": "assistant",
      "content": "<think>User asks: \"What is 2+2? Use the calculator tool.\" So we need to call calculator function with expression \"2+2\".</think>",
      "tool_calls": [
        {
          "id": "be3f751a-ff6f-4b60-8e84-d60cffabe589",
          "index": 0,
          "type": "function",
          "function": {
            "name": "calculator",
            "arguments": "{\n  \"expression\": \"2+2\"\n}"
          }
        }
      ]
    },
    {
      "role": "tool",
      "content": "2+2 = 4",
      "tool_name": "calculator",
      "tool_call_id": "be3f751a-ff6f-4b60-8e84-d60cffabe589"
    }
  ],
  "stream": false,
  "tools": [
    {
      "type": "function",
      "function": {
        "name": "calculator",
        "description": "Perform basic arithmetic",
        "parameters": {
          "type": "object",
          "properties": {
            "expression": {
              "type": "string",
              "description": "Math expression"
            }
          },
          "required": [
            "expression"
          ]
        }
      }
    }
  ]
}' | jq
```

[p1.sh](https://github.com/user-attachments/files/25047238/p1.sh)

[p2.sh](https://github.com/user-attachments/files/25047252/p2.sh)

1. **Baseline**: Run `p1.sh`. You will receive a `tool_calls` response (Correct).
2. **State Injection**: Run `p2.sh`. You will receive the final answer "4" (Correct).
3. **The Leak**: Run `p1.sh` again.
* *Expected*: A `tool_calls` response (identical to step 1).
* *Actual*: The model directly returns the final answer "4", bypassing the tool call (Bug).

## Screenshot

Image

ssam18:
Fixes #1362

The KV prefix cache was causing tool results from previous requests to leak into new independent requests. When a conversation with tool results was cached, subsequent requests with the same initial prompt would incorrectly reuse that cached state, causing the model to skip tool calling and return the final answer directly.

This fix disables KV prefix caching when tool-related messages (tool_calls or tool results) are present in the conversation, ensuring each stateless request is truly independent.

### DIFF
--- a/src/exo/worker/engines/mlx/generator/generate.py
+++ b/src/exo/worker/engines/mlx/generator/generate.py
@@ -155,6 +155,26 @@ def eos_ids_from_tokenizer(tokenizer: TokenizerWrapper) -> list[int]:
     return eos
 
 
+def _has_tool_messages(task: TextGenerationTaskParams) -> bool:
+    """Check if the conversation contains tool-related messages.
+
+    Tool-related messages include:
+    - Messages with tool_calls (assistant requesting tool execution)
+    - Messages with role="tool" (tool results)
+
+    These need special cache handling to avoid prefix matching issues.
+    """
+    if task.chat_template_messages is None:
+        return False
+
+    for msg in task.chat_template_messages:
+        if msg.get("tool_calls") is not None:
+            return True
+        if msg.get("role") == "tool":
+            return True
+
+    return False
+
 def mlx_generate(
     model: Model,
     tokenizer: TokenizerWrapper,
@@ -170,6 +190,11 @@ def mlx_generate(
     # Do not use the prefix cache if we are trying to do benchmarks.
     is_bench = task.bench
     if is_bench:
+        kv_prefix_cache = None
+
+    # Disable prefix cache if there are tool-related messages
+    if _has_tool_messages(task):
+        logger.info("Disabling KV prefix cache due to tool messages in conversation")
         kv_prefix_cache = None
 
     # Use prefix cache if available, otherwise create fresh cache


Evanev7:
Thanks for the contribution! KV Prefix caching is most useful for coding tasks which is also where tool calling is most heavily used, so this seems like it will be a pretty hefty performance regression. I do not believe this is the correct way to go about solving this issue, but I'll do some testing ASAP.

ssam18:
> Thanks for the contribution! KV Prefix caching is most useful for coding tasks which is also where tool calling is most heavily used, so this seems like it will be a pretty hefty performance regression. I do not believe this is the correct way to go about solving this issue, but I'll do some testing ASAP.

You are right! This trades performance for correctness. I chose the conservative approach of disabling the cache when tool messages are present because the prefix matching was causing incorrect behavior where the model would skip tool calls entirely. I am definitely open to a better solution that maintains both correctness and performance. Perhaps using a cache key that includes message structure metadata rather than just prompt text?

Let me know your thought.

rltakashige:
Thank you for the contribution. However, prefix caching is certainly not the root cause of this issue (which has been closed).

PR #1191, opened 2026-01-18
Add custom model support with model_id as sole identifier

AlexCheema:
## Summary
- Add support for adding custom HuggingFace models to exo with robust config parsing
- Use full HuggingFace model_id as the sole identifier (removes short_id/hugging_face_id confusion)
- New registry data structure with base_models.json and variants.json

## New API Endpoints
- `POST /models/custom` - Add custom model by HuggingFace repo ID
- `DELETE /models/custom?model_id=...` - Remove custom model
- `PATCH /models/custom?model_id=...` - Update custom model settings
- `GET /models/search` - Search HuggingFace Hub for MLX models

## Key Features
- Automatic detection of model architecture, layers, and tensor parallel support
- Robust config parsing with fallbacks for various HuggingFace config formats
- User models persisted to `~/.cache/exo/user_models/`
- Migration logic for existing user model files

## Breaking Changes
- API responses now use `id` field with full model_id (e.g., "mlx-community/Meta-Llama-3.1-8B-Instruct-4bit")
- Removed `hugging_face_id` field from API responses (now redundant)
- Custom model endpoints use query params instead of path params

## Test plan
- [x] `uv run basedpyright` - Type check passes
- [x] `uv run ruff check` - Lint passes
- [x] `cd dashboard && npm run build` - Dashboard builds
- [x] `uv run pytest` - Tests pass

🤖 Generated with [Claude Code](https://claude.ai/code)

PR #1368, opened 2026-02-04
feat: add custom HuggingFace model support

AlexCheema:
## Motivation

Users should be able to run any HuggingFace model, not just the ones we ship TOML cards for. Continues the aim of #1191 with a minimal implementation on top of the current TOML model card system.

Custom cards are saved to `~/.exo/custom_model_cards/` rather than the bundled `resources/inference_model_cards/` because `RESOURCES_DIR` is read-only in PyInstaller bundles (`sys._MEIPASS`). This also fixes `fetch_from_hf` which was saving cards to the wrong path (`resources/` root instead of `resources/inference_model_cards/`).

## Changes

- Add `EXO_CUSTOM_MODEL_CARDS_DIR` constant (`~/.exo/custom_model_cards/`)
- Update `model_cards.py`: add custom dir to search path, fix `save_to_custom_dir`, add `delete_custom_card`/`is_custom_card`
- Add `POST /models/add` and `DELETE /models/custom/{model_id}` API endpoints
- Add `is_custom` field to `ModelListModel` API response
- Dashboard: add custom model input form in dropdown, delete button for custom models, show actual API errors, auto-select newly added model

## Why It Works

Two separate directories for model cards: the bundled read-only `resources/inference_model_cards/` for built-in cards, and user-writable `~/.exo/custom_model_cards/` for custom cards. Both are scanned when listing models. This works in all environments including PyInstaller bundles where `RESOURCES_DIR` points to `sys._MEIPASS`.

### Manual Testing
- Add a custom model via the dropdown (e.g. `mlx-community/Llama-3.2-1B-Instruct-4bit`)
- Verify it appears in the model list with the delete (x) button
- Delete it and verify it disappears
- Try adding an invalid model ID and verify the actual error is shown

### Automated Testing
- `uv run pytest src/` — passes
- `cd dashboard && npm run build` — builds

### DIFF
--- a/dashboard/src/routes/+page.svelte
+++ b/dashboard/src/routes/+page.svelte
@@ -551,13 +551,21 @@
       });
 
       if (!response.ok) {
-        const err = await response.json();
-        customModelError = err.detail || "Failed to add model";
+        try {
+          const err = await response.json();
+          customModelError =
+            err.detail || `Failed to add model (${response.status})`;
+        } catch {
+          customModelError = `Failed to add model (${response.status}: ${response.statusText})`;
+        }
         return;
       }
 
+      const added = await response.json();
       customModelInput = "";
       await fetchModels();
+      selectPreviewModel(added.id);
+      isModelDropdownOpen = false;
     } catch {
       customModelError = "Network error";
     } finally {


Evanev7:
Main point here is that this only works on the device you're looking at, when ideally we'd sync this across all devices, putting custom model cards in state in some capacity. i think we should merge this now with this caveat though

AlexCheema:
> Main point here is that this only works on the device you're looking at, when ideally we'd sync this across all devices, putting custom model cards in state in some capacity. i think we should merge this now with this caveat though

That's a good catch!
I wonder if it completely breaks things if the model card is only on one node and you try to launch an Instance on multiple nodes. Going to try this now.

AlexCheema:
Just tested this and indeed there's a caveat that the model card itself only gets added to one node, however launching model distributed does work which is cool! e.g. Here I'm running two custom models: `mlx-community/gemma-3-1b-it-4bit` and `unsloth/Ministral-3-3B-Instruct-2512` which I only added on `james`.

Screenshot 2026-02-04 at 4 41 50 AM

PR #1369, opened 2026-02-04
feat: add model picker modal with grouped models and HF Hub search

AlexCheema:
## Motivation

Reimplements the model picker modal from #1191 on top of the custom model support branch. Replaces the inline model dropdown with a full-featured modal that groups models by base model, supports filtering, favorites, and HuggingFace Hub search.

## Changes

**Backend:**
- Add `family`, `quantization`, `base_model`, `capabilities` metadata fields to `ModelCard` and all 40 TOML model cards
- Pass new fields through `ModelListModel` and `get_models()` API response
- Add `GET /models/search` endpoint using `huggingface_hub.list_models()`

**Dashboard (7 new files):**
- `ModelPickerModal.svelte` — Main modal with search, family filtering, HuggingFace Hub tab
- `ModelPickerGroup.svelte` — Expandable model group row with quantization variants
- `FamilySidebar.svelte` — Vertical sidebar with family icons (All, Favorites, Hub, model families)
- `FamilyLogos.svelte` — SVG icons for each model family
- `ModelFilterPopover.svelte` — Capability and size range filters
- `HuggingFaceResultItem.svelte` — HF search result item with download/like counts
- `favorites.svelte.ts` — localStorage-backed favorites store

**Integration:**
- Replace inline dropdown in `+page.svelte` with button that opens `ModelPickerModal`
- Custom models shown in Hub tab with delete support

**Polish:**
- Real brand logos (Meta, Qwen, DeepSeek, OpenAI, GLM, MiniMax, Kimi, HuggingFace) from Simple Icons / LobeHub
- Clean SVG stroke icons for capabilities (thinking, code, vision, image gen)
- Consistent `border-exo-yellow/10` borders, descriptive tooltips throughout
- Cluster memory (used/total) shown in modal header
- Selected model highlight with checkmark for both single and multi-variant groups
- Cursor pointer on all interactive elements, fix filter popover click-outside bug
- Custom models now appear in All tab alongside built-in models

## Bug Fix: Gemma 3 EOS tokens

Also included in this branch: fix for Gemma 3 models generating infinite `<end_of_turn>` tokens. The tokenizer's `eos_token_ids` was missing token ID 106 (`<end_of_turn>`), so generation never stopped. The fix appends this token to the EOS list after loading the tokenizer. Also handles `eos_token_ids` being a `set` (not just a `list`).

## Why It Works

Model metadata (family, capabilities, etc.) is stored directly in TOML cards rather than derived from heuristics, ensuring accuracy. The modal groups models by `base_model` field so quantization variants appear together. Custom models are separated into the Hub tab since they lack grouping metadata.

### Manual Testing
- Open dashboard, click model selector to open modal
- Browse models by family sidebar, search, and filters
- Expand model groups to see quantization variants
- Star favorites and verify persistence across page reloads
- Navigate to Hub tab, search and add models
- Verify error messages shown for invalid model IDs
- Run a Gemma 3 model and verify generation stops at `<end_of_turn>`

### Automated Testing
- `nix fmt` — clean
- `uv run pytest src/` — 173 passed
- `cd dashboard && npm run build` — builds successfully


src/exo/worker/engines/mlx/utils_mlx.py
```diff
@@ -384,6 +384,15 @@ def _patched_encode(text: str, **_kwargs: object) -> list[int]:
         eos_token_ids=eos_token_ids,
     )
 
+    if "gemma-3" in model_id_lower:
```
rltakashige, line 387: Can we use the type of the rather than the model id?

rltakashige (approved):
Haven't run it but looks good!

src/exo/worker/engines/mlx/utils_mlx.py
AlexCheema: Good call — tracked in https://github.com/exo-explore/exo/issues/1371. Will fix in a follow-up PR.

PR #1372, opened 2026-02-04
Fix graceful process shutdown in macOS app

### LINKED ISSUE #1370: [BUG] Memory not released after stopping exo instance
## Describe the bug

When running exo and then stopping the instance (either through normal shutdown or process termination), GPU/system memory allocated for models is not properly released. This persists even after the exo process has completely stopped, requiring a system restart to reclaim the memory.

## To Reproduce

1. Start exo on a Mac machine
2. Load a model and run inference
3. Stop the exo instance
4. Check memory usage
5. Observe that memory remains allocated and is not released back to the system

## Expected behavior

When exo is stopped, all allocated memory (especially GPU memory used by MLX) should be properly released and returned to the system.

## Actual behavior

Memory remains allocated after exo stops. The only way to reclaim the memory is to restart the Mac.

## Environment

- macOS Tahoe (Version 26.2)
- EXO Version: 1.0.67
- Hardware:
  - Device 1: Mac Studio (2025), Apple M4 Max chip, 128GB RAM
  - Device 2: Mac Studio (2025), Apple M4 Max chip, 128GB RAM
- Interconnection:
  - Thunderbolt 5 cable between Device 1 and 2

AlexCheema:
## Motivation

Fixes #1370

When the macOS app stops exo, GPU/system memory isn't released. This happens because:

1. The macOS app calls `process.terminate()` (SIGTERM) but the Python process only registers a graceful shutdown handler for SIGINT, not SIGTERM. SIGTERM's default Python behavior raises `SystemExit` which bypasses the cleanup cascade (runner subprocess MLX cleanup via `mx.clear_cache()`, channel closing, etc.).
2. The app doesn't wait for the process to actually finish cleanup — it immediately nils out the process reference.

## Changes

**`src/exo/main.py`**: Register SIGTERM handler alongside SIGINT so the graceful shutdown cascade (`Node.shutdown()` → cancel task group → worker/runner cleanup → `mx.clear_cache()` + `gc.collect()`) runs regardless of which signal is received.

**`app/EXO/EXO/ExoProcessController.swift`**: Replace immediate `process.terminate()` with escalating shutdown per @Evanev7's suggestion:
1. Send SIGINT via `process.interrupt()` — triggers the registered Python handler for graceful cleanup
2. Wait up to 5 seconds for the process to exit
3. If still running, escalate to SIGTERM via `process.terminate()`
4. Wait up to 3 seconds
5. If still running, force kill via SIGKILL

The escalation runs in a detached `Task` so the UI updates immediately (status → stopped) without blocking.

## Why It Works

The root cause is that SIGTERM wasn't triggering the graceful shutdown path. By registering a SIGTERM handler in Python and sending SIGINT first from the macOS app, the process gets a chance to run the full cleanup cascade: cancelling the task group, shutting down runners (which call `del model; mx.clear_cache(); gc.collect()`), closing channels, and flushing logs. The escalation to SIGTERM and SIGKILL ensures the process always terminates even if graceful shutdown hangs.

### Manual Testing

- Start exo via macOS app, load a model, run inference
- Stop via the toggle switch, verify memory is released without requiring a system restart
- Test rapid stop/start (restart) to ensure no race conditions

### Automated Testing
- `nix fmt` — no changes

### DIFF
--- a/.github/workflows/pipeline.yml
+++ b/.github/workflows/pipeline.yml
@@ -8,33 +8,6 @@ on:
       - main
 
 jobs:
-  typecheck:
-    runs-on: ubuntu-latest
-    steps:
-      - name: Checkout repository
-        uses: actions/checkout@v4
-        with:
-          lfs: false
-
-      - uses: cachix/install-nix-action@v31
-        with:
-          nix_path: nixpkgs=channel:nixos-unstable
-
-      - uses: cachix/cachix-action@v14
-        name: Configure Cachix
-        with:
-          name: exo
-          authToken: "${{ secrets.CACHIX_AUTH_TOKEN }}"
-
-      - name: Load nix develop environment
-        run: nix run github:nicknovitski/nix-develop/v1
-
-      - name: Sync dependencies
-        run: uv sync --all-packages
-
-      - name: Run type checker
-        run: uv run basedpyright --project pyproject.toml
-
   nix:
     name: Build and check (${{ matrix.system }})
     runs-on: ${{ matrix.runner }}
--- a/MISSED_THINGS.md
+++ b/MISSED_THINGS.md
@@ -5,21 +5,21 @@
 [X] Fetching download status of all models on start
 [X] Deduplication of tasks in plan_step.
 [X] resolve_allow_patterns should just be wildcard now.
-[] no mx_barrier in genreate.py mlx_generate at the end.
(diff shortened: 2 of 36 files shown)


AlexCheema:
Hey @Evanev7 can you check this? Does it make sense to also catch SIGTERM like this?

PR #1373, opened 2026-02-04
EXO: add CLI flags for root install/uninstall

JakeHillion:
The macOS app required user interaction via AppleScript prompts to install or uninstall network configuration components, making automated deployments difficult.

Added --install and --uninstall command line flags that execute the network setup scripts directly when running as root, bypassing GUI prompts. Created a new main.swift entry point that parses CLI arguments and delegates to NetworkSetupHelper's new direct execution methods.

This enables headless installation via `sudo EXO --install` for automated deployment scenarios while preserving the existing GUI behavior when launched normally.

Test plan:

- Deployed to a machine that didn't have the content installed. Got blocked on the popup and EXO never launched.
- Relaunched EXO, confirmed it still never starts because of the popup.
- Ran `sudo /Applications/EXO.app/Contents/MacOS/EXO --install`
- Launched EXO - the API started as expected.
- Ran `sudo /Applications/EXO.app/Contents/MacOS/EXO --uninstall`
- Launched EXO - got the popup.

### DIFF
--- a/app/EXO/EXO/main.swift
+++ b/app/EXO/EXO/main.swift
@@ -1,0 +1,85 @@
+//
+//  main.swift
+//  EXO
+//
+//  Created by Jake Hillion on 2026-02-03.
+//
+
+import Foundation
+
+/// Command line options for the EXO app
+enum CLICommand {
+    case install
+    case uninstall
+    case help
+    case none
+}
+
+/// Parse command line arguments to determine the CLI command
+func parseArguments() -> CLICommand {
+    let args = CommandLine.arguments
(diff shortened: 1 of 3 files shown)


app/EXO/EXO/main.swift
```diff
@@ -0,0 +1,85 @@
+//
+//  main.swift
+//  EXO
+//
+//  Created by Jake Hillion on 2025-02-03.
```
samiamjidkhan:
```suggestion
//  Created by Jake Hillion on 2026-02-03.
```

PR #1376, opened 2026-02-04
add resources dir to nix

Evanev7:
add resources directory to the nix exo package, and fix the env for the dashboard dir

### DIFF
--- a/python/parts.nix
+++ b/python/parts.nix
@@ -7,9 +7,6 @@
       workspace = inputs.uv2nix.lib.workspace.loadWorkspace {
         workspaceRoot = inputs.self;
       };
-      resourcesSrc = lib.cleanSourceWith {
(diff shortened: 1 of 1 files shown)


resources/parts.nix
```diff
@@ -0,0 +1,17 @@
...
+    let
+      # Filter source to only include resources directory
+      resourcesSrc = lib.cleanSourceWith {
+        src = inputs.self + "/resources";
+      };
+    in
+    {
+      packages.resources = pkgs.runCommand "exo-resources" { } ''
+        cp -r ${resourcesSrc} $out
+      '';
```
JakeHillion: Do we need this? I wonder whether the `resourcesSrc` will work as a package by itself.

Evanev7:
yeah we can just drop the resources/parts.nix. a whole 9 line change

PR #1378, opened 2026-02-04
Rust rewrite

Evanev7:
## Motivation

We've hit the limits of our early networking, time to start working on it.

## Changes

Essentially rewrote the libp2p to be simpler and closer to default.
Also moves the main loop into rust, not the bindings layer. 

### Manual Testing
Ran a four node cluster stably over a few minutes - networking seems essentially unchanged (yay!)


AlexCheema:
## Code Review: Rust Rewrite (#1378)

Thanks for this PR, @Evanev7. This is a significant simplification of the networking stack -- 2838 lines deleted with only 649 added. I've reviewed the changes in detail below.

---

### What I Like

- **Massive reduction in complexity.** The networking layer is dramatically easier to follow now. Removing the `NetworkingHandle` with its oneshot channels, the `discovery.rs`/`keep_alive.rs`/`swarm.rs` modules, and the `pylibp2p/` wrappers is a big clarity win.
- **Removing the nightly Rust requirement.** Pinning to nightly via `rust-toolchain.toml` was a recurring source of build fragility. Compiling on stable is a major improvement.
- **QUIC transport is a good choice.** It natively handles encryption and multiplexing, making the explicit Noise + Yamux + TCP stack unnecessary.
- **The new `PyPeer` API is clean.** `run()/recv()/send()/subscribe()/unsubscribe()` is a much simpler interface than the old channel-based approach.
- **Version-based default namespace** via `importlib.metadata.version("exo")` is a sensible default for peer isolation.

---

### Critical / High Severity

#### 1. Namespace isolation is significantly weaker

The old approach used pnet pre-shared keys for **cryptographic** isolation -- peers in different namespaces physically could not communicate. The new `identify` protocol approach blacklists peers *after* connecting and exchanging identify info. This means:

- Peers from different namespaces still discover each other via mDNS and establish full connections before being filtered.
- There is a race window where gossipsub messages could leak between namespaces before identify completes.
- An attacker on the LAN could potentially inject messages before the namespace check runs.

This is a meaningful security regression. Is the weaker isolation model acceptable for the use cases you're targeting? If not, consider at least adding a gossipsub message-level namespace check as a defense-in-depth measure.

#### 2. `known_peers` HashMap grows unboundedly

In `rust/networking/src/lib.rs`, mDNS-discovered peers are added to `known_peers`, but when `failed_namespace()` is called, the peer is blacklisted in gossipsub without being removed from `known_peers`. This means:

- Every mDNS discovery cycle re-dials blacklisted peers (since they're still in `known_peers`).
- The map accumulates entries for every peer ever seen, with no eviction.

Consider removing peers from `known_peers` when they fail namespace validation, or at least skipping dial attempts for blacklisted peers.

#### 3. Error handling uses bare `()` error type

`Peer::run()` returns `Result<(), ()>`, and `handle_event`/`handle_message` propagate errors via `.map_err(|_| ())?`. When the `to_client` channel send fails, the entire peer loop exits silently. There is no way for the Python caller to distinguish between "channel closed intentionally" and "unexpected failure." Consider introducing a small error enum (even just `enum PeerError { ChannelClosed, SwarmError(String) }`) so callers can react appropriately.

#### 4. `ListenError` is empty and not useful

`pub struct ListenError;` has no `Display` or `Error` impl. It cannot be meaningfully reported or matched on by Python callers. Either add a message field and trait impls, or use an existing error type.

#### 5. `assert!` in recv handler will crash in production

In `router.py`, the networking recv loop does:
```python
assert (msg := swarm_event.downcast_message()) is not None
```

If a `PySwarmEvent` is somehow neither discovered, expired, nor message, this crashes the entire router with an `AssertionError`. In production code, this should log a warning and `continue` instead.

---

### Medium Severity

#### 6. Re-dials *all* known peers on every mDNS discovery

In `handle_event` for `Mdns::Discovered`, after adding the newly discovered peer, the code iterates **all** `known_peers` and dials every one of them. The comment mentions `DialOpts` handles rate limiting, but this is still O(n) work on every discovery event and will scale poorly. Consider dialing only the newly discovered peer.

#### 7. Spurious `Expired` events for namespace-filtered peers

When a peer fails the namespace check, it gets blacklisted but no `Discovered` event was ever sent to the client. If that peer later expires from mDNS, the code sends `FromSwarm::Expired(peer_id)` to the Python side for a peer it never knew about. The Python side should either:
- Ignore `Expired` events for unknown peers, or
- The Rust side should suppress `Expired` for peers that never passed namespace validation.

#### 8. `ConnectionMessage` lost IP/port information

The old `ConnectionMessage` included `remote_ipv4` and `remote_tcp_port`. The new version only carries `node_id` and `expired`. If any downstream code or diagnostics relied on IP/port info, this is a silent regression. Worth confirming no consumers depend on those fields (dashboard, logging, etc.).

#### 9. `no_api` field on `Args` may break programmatic construction

The `Args` model now has `no_api: bool` without a default value. While argparse provides a default of `False` via `store_true`, constructing `Args(...)` in tests or programmatically now requires explicitly passing `no_api=`. Consider adding `no_api: bool = False` to the model definition.

#### 10. Typo

`rust/networking/src/lib.rs` contains:
```
"gossipsub behaviour failed ot build"
```
Should be `"to"`.

---

### Minor / Nits

- **`Keypair.generate()` rename**: Previously `generate_ed25519()`, now just `generate()`. The underlying behavior is the same, but the old name was more explicit. Just noting the rename.
- **Log level changes** (`debug` -> `trace`) in `master/main.py` and `worker/main.py` are good for reducing noise. Worth noting in the PR description.
- **Test coverage**: There is only one Rust test (`subscribe_and_unsubscribe_do_not_error`), which is a smoke test. There are no tests for namespace filtering, peer discovery/expiry, or message routing. The `// TODO: more tests` comment acknowledges this, but given this is a rewrite of critical infrastructure, more coverage would increase confidence.
- **`pub swarm` field on `Peer`**: The `swarm` field is `pub`, exposing the internal libp2p swarm directly. Consider `pub(crate)` to preserve encapsulation.

---

### Summary

This is a well-motivated simplification that makes the networking stack much more approachable. The main concerns are around the weaker namespace isolation model (pnet crypto isolation -> post-connect identify filtering), unbounded growth in `known_peers`, and the silent error handling in the peer loop. Addressing the critical items -- especially the `assert!` crash, the `known_peers` growth, and the error types -- would make this ready to ship with confidence.

Nice work on the cleanup overall.

PR #1384, opened 2026-02-04
UI: show failure reason when an instance fails

### LINKED ISSUE #1350: Show Failure Reason in Dashboard
Often users are running into instance failures but don't know why it failed.

We should show the failure reason.

jayy-77:
Fixes #1350

## Summary
- Dashboard: include `RunnerFailed` error messages in instance status details.
- macOS app: thread runner failure messages into `InstanceViewModel` and show them in the instance row when failed.

## Test plan
- `cd dashboard && npm run check`


AlexCheema:
## Code Review: UI: show failure reason when an instance fails

Thanks for working on this, @jayy-77! Surfacing error messages that were previously hidden is a genuinely useful improvement. I've reviewed all four changed files. Here are my findings.

---

### Overall Assessment

The feature works as intended: it threads `RunnerFailed` error messages through both the Swift macOS app and the Svelte dashboard so users can see *why* an instance failed. The implementation is correct and defensive. I have one medium-priority concern about the refactor verbosity and a few minor items below.

---

### Medium Priority

**1. `deriveInstanceStatus` refactor adds significant verbosity without clear benefit**

The original function used ~10 concise one-liner returns. The refactored version expands these into ~60 lines of explicit objects, where every non-`FAILED` branch now manually specifies `errorMessage: null`.

A cleaner approach would be to compute `errorMessage` at the top, then add it to the return at the single point where it matters (the `FAILED` branch), and default all other branches without requiring the explicit field. For example:

```ts
function deriveInstanceStatus(instanceWrapped: unknown): {
  statusText: string;
  statusClass: string;
  errorMessage: string | null;
} {
  // ... existing runner status extraction logic ...
  // ... failureMessages collection logic (this part is good) ...

  const errorMessage =
    failureMessages.length > 0
      ? Array.from(new Set(failureMessages)).join(" \u2022 ")
      : null;

  const base = (statusText: string, statusClass: string) => ({
    statusText,
    statusClass,
    errorMessage: null as string | null,
  });

  if (statuses.length === 0) return base("PREPARING", "inactive");
  if (has("Failed")) return { statusText: "FAILED", statusClass: "failed", errorMessage };
  if (has("Shutdown")) return base("SHUTDOWN", "inactive");
  if (has("Loading")) return base("LOADING", "starting");
  // ... etc, preserving the original compact style
  return base("RUNNING", "active");
}
```

This keeps the function readable, adds the new `errorMessage` field, and only the `FAILED` branch differs. The compact one-liner style was one of the strengths of the original code.

**2. Dual camelCase/snake_case field check**

```ts
const msg =
  (typeof p.errorMessage === "string" ? p.errorMessage : null) ??
  (typeof p.error_message === "string" ? p.error_message : null);
```

I verified the serialization path: `RunnerFailed` extends `TaggedModel` which extends `CamelCaseModel`, and the API serializes using Pydantic aliases. So the API response will contain `errorMessage` (camelCase) inside the tagged wrapper `{"RunnerFailed": {"errorMessage": "..."}}`.

The `error_message` (snake_case) fallback should never match in practice. This isn't a bug, but it does suggest uncertainty about the API contract. I'd recommend either:
- Removing the snake_case fallback and only checking `p.errorMessage`, or
- Adding a brief comment explaining *why* both are checked (e.g., "fallback for potential future API changes")

Without explanation, the dual check looks like the author wasn't sure which format the API sends, which could confuse future readers.

---

### Minor

**3. SVG `<title>` fix is unrelated to the failure reason feature**

The `ModelPickerGroup.svelte` changes (moving `title` from SVG attribute to child element for proper tooltip rendering) are a legitimate fix, but unrelated to showing failure reasons. Ideally this would be a separate commit or at least called out in the PR description. Not a blocker, just good housekeeping.

**4. Swift: only shows first runner failure reason**

```swift
.compactMap { runners[$0]?.errorMessage?.trimmingCharacters(in: .whitespacesAndNewlines) }
.first(where: { !$0.isEmpty })
```

If multiple runners fail with different reasons, only the first non-empty message is shown. The dashboard handles this better by collecting all unique messages and joining them with `" \u2022 "`. Consider aligning the Swift behavior for consistency -- or at minimum, add a comment noting the intentional limitation (e.g., `// Shows only the first failure; multi-runner errors are rare`).

**5. Redundant `.trimmingCharacters` in `InstanceRowView.swift`**

The `failureReason` is already trimmed at extraction time in `InstanceViewModel.swift`:
```swift
runners[$0]?.errorMessage?.trimmingCharacters(in: .whitespacesAndNewlines)
```

Then it's trimmed again at display time in `InstanceRowView.swift`:
```swift
if ... !reason.trimmingCharacters(in: .whitespacesAndNewlines).isEmpty
```

The second trim is redundant. Either trust the already-trimmed value or only trim in one place.

---

### Positive Notes

- Clean, focused feature that directly addresses #1350
- Good defensive coding with trimming and empty-string checks
- Error deduplication with `Set` in the dashboard prevents repetitive messages
- `lineLimit(3)` in Swift is a sensible cap to prevent UI overflow
- The `getTagged()` pattern is used correctly to extract the `RunnerFailed` payload

---

### Summary

The core feature is solid and the logic is correct. The main actionable feedback is reducing the verbosity of the `deriveInstanceStatus` refactor (item 1) and clarifying the camelCase/snake_case situation (item 2). The rest are minor polish items. Nice work overall!

PR #1385, opened 2026-02-05
move runner ack after status update

Evanev7:
Runner only ACKs after its status gets updated.
Before a double plan could run after an ACK + 0.1s but before the StatusUpdated.

It's been hard to catch this race, so no manual testing other than the most basic "yes exo is ok after i spawn and delete like 20 instances"


Evanev7:
closing in favor of #1381

PR #1398, opened 2026-02-05
fix: retry downloads on transient errors instead of breaking

Hmbown:
## Motivation

`download_file_with_retry()` has a `break` in the generic exception handler that exits the retry loop after the first transient failure. This means network timeouts, connection resets, and server errors all cause an immediate download failure — the two remaining retry attempts never run.

## Changes

**download_utils.py**: Replaced `break` with logging and exponential backoff in the generic exception handler, matching the existing rate-limit handler behavior.

Before:
```python
except Exception as e:
    on_connection_lost()
    if attempt == n_attempts - 1:
        raise e
    break  # exits loop immediately
```

After:
```python
except Exception as e:
    on_connection_lost()
    if attempt == n_attempts - 1:
        raise e
    logger.error(f"Download error on attempt {attempt + 1}/{n_attempts} ...")
    logger.error(traceback.format_exc())
    await asyncio.sleep(2.0**attempt)
```

## Why It Works

The `break` statement was bypassing the retry mechanism entirely. Replacing it with the same log-and-backoff pattern used by the `HuggingFaceRateLimitError` handler means all 3 attempts are actually used before giving up. The exponential backoff (1s, 2s) gives transient issues time to resolve between attempts.

### Manual Testing
- Downloads that hit transient network errors now retry instead of failing immediately

### Automated Testing
- `uv run pytest src/exo/download/tests/ -v` — 11 tests pass

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -31,8 +31,6 @@ dependencies = [
 ]
 
 [project.scripts]
-exo-master = "exo.master.main:main"
-exo-worker = "exo.worker.main:main"
 exo = "exo.main:main"
 
 # dependencies only required for development
--- a/python/parts.nix
+++ b/python/parts.nix
@@ -59,6 +59,16 @@
         }
       );
 
+      mkPythonScript = name: path: pkgs.writeShellApplication {
+        inherit name;
+        runtimeInputs = [ exoVenv ];
+        runtimeEnv = {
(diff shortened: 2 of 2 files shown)


rltakashige (approved):
Thanks for this!

My initial intention was that we would not want to retry on unknown errors. However, given the state of the downloads code anyway, it is probably sensible to do this.

PR #1407, opened 2026-02-06
slow down catchup

Evanev7:
our event log request blasted the whole event log over libp2p, now it just does the next 1000 messages - hopefully allowing nodes to catch up a bit more consistently for long lived clusters


rltakashige:
Going to merge for now, as this won't make anything worse.

PR #1408, opened 2026-02-06
Handle config.json not found (image models)

ciaranbor:
## Motivation

When downloading image models, a missing config.json file triggers a FileNotFoundError inside download_file_with_retry. This error was being caught by the generic except Exception handler and retried 3 times before failing. Then, the whole thing would be retried from the start

## Changes

  - src/exo/download/download_utils.py: Added FileNotFoundError to the list of immediately-raised exceptions in download_file_with_retry, alongside HuggingFaceAuthenticationError. This prevents useless retries when a file genuinely doesn't exist on the remote.
  - src/exo/master/api.py: Wrapped ModelCard.load(model_id) in a try/except that converts failures into an HTTPException(400) with a descriptive error message, giving API consumers a clear error response.

## Why It Works

  - FileNotFoundError is a deterministic error — the file won't appear on retry, so re-raising immediately avoids 3 wasted download attempts with exponential backoff.
  - Catching ModelCard.load() failures and returning a 400 HTTP response prevents unhandled exceptions from surfacing as opaque 500 errors in the API.

### Manual Testing

Verified an image model not in model cards does not cause an infinite error loop

### DIFF
--- a/dashboard/src/lib/components/ChatMessages.svelte
+++ b/dashboard/src/lib/components/ChatMessages.svelte
@@ -13,6 +13,7 @@
   import type { MessageAttachment } from "$lib/stores/app.svelte";
   import MarkdownContent from "./MarkdownContent.svelte";
   import TokenHeatmap from "./TokenHeatmap.svelte";
+  import ImageLightbox from "./ImageLightbox.svelte";
 
   interface Props {
     class?: string;
@@ -100,6 +101,9 @@
   // Copied state for feedback
   let copiedMessageId = $state<string | null>(null);
   let expandedThinkingMessageIds = $state<Set<string>>(new Set());
+
+  // Lightbox state
+  let expandedImageSrc = $state<string | null>(null);
 
   // Uncertainty heatmap toggle
   let heatmapMessageIds = $state<Set<string>>(new Set());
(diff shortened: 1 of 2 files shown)


src/exo/master/api.py
```diff
@@ -386,7 +386,12 @@ async def get_placement_previews(
         if len(list(self.state.topology.list_nodes())) == 0:
             return PlacementPreviewResponse(previews=[])
 
-        model_card = await ModelCard.load(model_id)
+        try:
+            model_card = await ModelCard.load(model_id)
+        except Exception as exc:
+            raise HTTPException(
+                status_code=400, detail=f"Failed to load model card: {exc}"
+            ) from exc
```
AlexCheema, line 394: What's the purpose of this? Wouldn't it already be raised as a 500 HTTPException by default?

ciaranbor: Yeah, it's just semantic. If this happens it means the user requested a non-existant custom model so 500 is inaccurate. Doesn't really matter

AlexCheema: makes sense - that works.

AlexCheema (approved):

PR #1421, opened 2026-02-07
Always load image model cards into cache

AlexCheema:
## Motivation

Follows up on #1408. Image models (FLUX, Qwen-Image, etc.) don't have a `config.json` on HuggingFace. Previously, image model TOML cards were only loaded into `_card_cache` when `EXO_ENABLE_IMAGE_MODELS=true`. When the flag was off but an image model was requested (e.g., via `get_placement_previews`), `ModelCard.load()` fell through to `fetch_from_hf()` which tried to download `config.json` — causing `FileNotFoundError` spam. #1408 added defensive error handling; this PR fixes the root cause.

## Changes

**`model_cards.py`**: Always include `image_model_cards/` in `CARD_SEARCH_PATH` so image model TOML cards are always loaded into `_card_cache`. `ModelCard.load()` then finds them directly and never falls through to `fetch_from_hf()`. The `EXO_ENABLE_IMAGE_MODELS` flag now controls whether image models appear in `get_model_cards()` (the listing) rather than whether they're loaded at all.

## Why It Works

`fetch_from_hf()` is designed for text models only (it hardcodes `tasks=[ModelTask.TextGeneration]` and requires `config.json`). Image models should never reach that path. By always having them in the cache, the lookup succeeds immediately and `fetch_from_hf()` is never called.

### Automated Testing

### DIFF
--- a/.github/workflows/build-app.yml
+++ b/.github/workflows/build-app.yml
@@ -396,7 +396,7 @@ jobs:
           path: output/EXO-${{ env.RELEASE_VERSION }}.dmg
 
       - name: Upload to S3
-        if: env.SPARKLE_S3_BUCKET != '' && github.ref_type == 'tag'
+        if: env.SPARKLE_S3_BUCKET != ''
         env:
           AWS_ACCESS_KEY_ID: ${{ secrets.AWS_ACCESS_KEY_ID }}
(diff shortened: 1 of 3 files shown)


ciaranbor (approved):

djhdusjbd-eng:
Understood thank you

aydenhailey78-hub:
No issues on my end

PR #1422, opened 2026-02-07
Add error handling to info gatherer monitor loops

AlexCheema:
## Motivation

If any of the `InfoGatherer` monitor loops throw an unexpected exception, the entire monitoring task crashes and never recovers. This can silently stop memory, network, or Thunderbolt data collection for the lifetime of the process.

## Changes

Wrap the body of each `while True` monitor loop in a try/except that logs the exception as a warning and continues to the next iteration. The sleep at the end of each loop runs regardless, providing natural backoff before retry.

Affected methods: `_monitor_misc`, `_monitor_system_profiler_thunderbolt_data`, `_monitor_memory_usage`, `_watch_system_info`, `_monitor_thunderbolt_bridge_status`.

`_monitor_macmon` already had its own error handling so was left as-is.

## Why It Works

A transient error (e.g., a subprocess failing, a permission issue) in one iteration no longer kills the loop. The warning log provides visibility while the monitor continues collecting data on subsequent iterations.

### Automated Testing

### DIFF
--- a/.mlx_typings/mlx/core/__init__.pyi
+++ b/.mlx_typings/mlx/core/__init__.pyi
@@ -2366,7 +2366,7 @@ class custom_function:
 def default_device() -> Device:
     """Get the default device."""
 
-def default_stream(device: Device) -> Stream:
+def default_stream(device: Device | DeviceType) -> Stream:
     """Get the device's default stream."""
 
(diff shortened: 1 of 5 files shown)


Evanev7 (approved):
ya fair enough.
if we ever see a scenario where we have looping errors this needs changing though

PR #1423, opened 2026-02-08
Fix unbounded event log memory growth

mlpy0:
## Summary

Fixes unbounded memory growth in the event log that causes ~17GB heap usage after 24 hours of operation on a 3-node cluster.

**Problem:**
- `_event_log: list[Event]` grows unbounded, reaching 200k+ events in ~1 day
- `out_for_delivery` dict never cleans up if acks are lost
- Observed via `sudo footprint -p <pid>`: 17GB in MALLOC_SMALL (Python heap)

**Solution:**
1. **CompactingEventLog**: New event log that automatically deduplicates high-frequency events with "replace by key" semantics:
   - `NodeGatheredInfo`: keeps only latest per `node_id`
   - `NodeDownloadProgress`: keeps only latest per `(node_id, model_id)`
   - `RunnerStatusUpdated`: keeps only latest per `runner_id`
   
   Critical events (tasks, instances, topology) are preserved in full. Index stability is maintained for `RequestEventLog` compatibility.

2. **out_for_delivery TTL**: Events pending delivery to master now have a 5-minute TTL. Previously, if an ack was lost, entries would stay in the dict forever.

**Memory impact (3-node cluster, 24 hours):**
- Before: ~270,000 events → ~17GB heap  
- After: ~5,100 events → ~50MB heap

## Changes

- **NEW**: `src/exo/utils/compacting_event_log.py` - CompactingEventLog class
- `src/exo/master/main.py` - Use CompactingEventLog, update RequestEventLog handler
- `src/exo/master/api.py` - Use CompactingEventLog, update /events endpoint
- `src/exo/worker/main.py` - Add TTL to out_for_delivery dict

## Test plan

- [x] Type checking passes (`uv run basedpyright`)
- [x] Linting passes (`uv run ruff check`)
- [x] Existing tests pass (`uv run pytest`)
- [ ] Manual verification: Run 3-node cluster for 24+ hours, monitor memory via `footprint`

## Notes

This addresses the feedback from the exo Discord channel: "The event log needs to exist from the first event, or the state cannot be reconstructed. However, we are looking at ways to massively reduce the size of it as it does get large over a long session. e.g. we only need the latest memory measured event."

The compaction approach keeps the latest value for events where only the current state matters, while preserving full history for events where sequence matters (tasks, instances, topology).

### DIFF
--- a/src/exo/master/api.py
+++ b/src/exo/master/api.py
@@ -144,6 +144,7 @@ from exo.shared.types.worker.instances import Instance, InstanceId, InstanceMeta
 from exo.shared.types.worker.shards import Sharding
 from exo.utils.banner import print_startup_banner
 from exo.utils.channels import Receiver, Sender, channel
+from exo.utils.compacting_event_log import CompactingEventLog
 from exo.utils.event_buffer import OrderedBuffer
 
 
@@ -175,7 +176,7 @@ class API:
         election_receiver: Receiver[ElectionMessage],
     ) -> None:
         self.state = State()
-        self._event_log: list[Event] = []
+        self._event_log = CompactingEventLog()
         self.command_sender = command_sender
         self.download_command_sender = download_command_sender
         self.global_event_receiver = global_event_receiver
@@ -224,6 +225,7 @@ class API:
     def reset(self, new_session_id: SessionId, result_clock: int):
         logger.info("Resetting API State")
         self.state = State()
+        self._event_log = CompactingEventLog()
         self.session_id = new_session_id
         self.event_buffer = OrderedBuffer[Event]()
         self._text_generation_queues = {}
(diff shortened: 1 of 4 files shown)


JakeHillion:
Thank you for the contribution! I'm going to close this in favour of #1432 which also solves the unbounded event log problem, but takes a different approach of storing our rarely read event log on disk. By doing that we avoid the problem of hosts having a serious issue with missing events in the history and keep it out of RAM.

One thing to note though, as I moved this on disk, I'm not seeing the sizes you mentioned of many gigabytes. Of course the on disk format is much more efficient than Python's heap, and my usage pattern might be different to yours, but I'd be shocked if this hit a gigabyte with my usage. If you're still seeing unbounded memory issues after this lands, please open an issue and we'll take a deeper look.

PR #1427, opened 2026-02-09
Yield from reachability checks

rltakashige:
## Motivation

check_reachable waits for all connection profile checks to be completed. Since there are retries on failures, this can take around 20s to resolve, preventing any instances from showing up. This feels very slow for UX, and it slows down distributed testing.

## Changes

Made check_reachable an async generator.

### Manual Testing
Works for me at least.

### DIFF
--- a/.mlx_typings/mlx/core/__init__.pyi
+++ b/.mlx_typings/mlx/core/__init__.pyi
@@ -2366,7 +2366,7 @@ class custom_function:
 def default_device() -> Device:
     """Get the default device."""
 
-def default_stream(device: Device) -> Stream:
+def default_stream(device: Device | DeviceType) -> Stream:
     """Get the device's default stream."""
 
 def degrees(a: array, /, *, stream: Stream | Device | None = ...) -> array:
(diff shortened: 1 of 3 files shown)


src/exo/utils/info_gatherer/net_profile.py
```diff
@@ -93,22 +95,33 @@ async def check_reachable(
...
+        expected_node_id: NodeId,
+        client: httpx.AsyncClient,
+        send: Sender[tuple[str, NodeId]],
+    ) -> None:
+        out: defaultdict[NodeId, set[str]] = defaultdict(set)
+        await check_reachability(target_ip, expected_node_id, out, client)
+        if expected_node_id in out:
+            await send.send((target_ip, expected_node_id))
+
+    async def _run_probes() -> None:
```
Evanev7: a channel! be still my beating heart! even if its immediately yielding into a generator...

anyway, can you inline the _run_probes into the main task group? it seems a little silly to write this function just to call it

rltakashige: I need a channel here because I don't think (?) there's an as_completed in anyio!

Done (on the inlining)

JakeHillion (approved):
Very nice! This data will have much better liveness now.

PR #1430, opened 2026-02-10
Fix setrlimit crash when hard file descriptor limit < 65535

mlpy0:
## Summary

`exo` crashes on startup when the system's hard file descriptor limit is below 65535, which occurs in macOS LaunchDaemon environments, Docker containers, and other restricted setups.

**Root cause:** `resource.setrlimit(resource.RLIMIT_NOFILE, (max(soft, 65535), hard))` raises `ValueError` when `hard < 65535` because the soft limit cannot exceed the hard limit.

**Fix:**
- **`main.py`**: Clamp the target soft limit to `min(max(soft, 65535), hard)` so it never exceeds the hard limit
- **`utils_mlx.py`**: Query current limits instead of hardcoding `(2048, 4096)`, which both crashed on restricted systems and incorrectly lowered the hard limit when it was set higher

## Test plan

- [x] `basedpyright` passes with 0 errors
- [x] `ruff check` passes
- [x] Verify startup works on system with hard limit < 65535 (tested in macOS LaunchDaemon with hard limit 10240)
- [x] Verify startup still works on default macOS (hard limit typically unlimited)


src/exo/worker/engines/mlx/utils_mlx.py
```diff
@@ -64,7 +64,8 @@
 
 Group = mx.distributed.Group
 # Needed for 8 bit model
-resource.setrlimit(resource.RLIMIT_NOFILE, (2048, 4096))
```
JakeHillion, line 67: Instead of extending this code in the globals, could you move this into `main` in `runner.py`? Thanks!

mlpy0:
Moved `setrlimit` from module-level in `utils_mlx.py` to the top of `main()` in `runner.py` as suggested.

PR #1432, opened 2026-02-10
event_log: move event log from unbounded in-memory list to disk

JakeHillion:
The master and API event logs (list[Event]) grew unbounded in RAM for
the lifetime of the process. Events are rarely read back (only for
RequestEventLog when a new node catches up, or the dashboard /events
endpoint).

Introduced a DiskEventLog class that writes length-prefixed msgpack
records to an append-only file, using a bounded LRU cache of byte
offsets for indexed access. On close, the active file is compressed
with ZSTD and rotated into a numbered archive slot, keeping the last 5
archives (events.1.bin.zst through events.5.bin.zst). On construction,
any stale active file from a crash is rotated before opening a fresh
log. The /events API endpoint now streams the JSON array one event at a
time rather than materializing the full list in memory. Deserialization
routes msgpack through json.dumps into Pydantic's validate_json() to
get correct JSON-mode coercion (e.g. string to enum) under strict mode.

This bounds memory usage to the LRU cache (128 entries) regardless of
event volume, while still supporting efficient sequential reads from
disk when needed.

Test plan:
- CI
- New unit tests for DiskEventLog: append/read, range queries, rotation
  on close, stale file recovery, idempotent close, successive sessions,
  archive retention limit (5 max)
- Tested on a cluster with 9000 events. /events continues working.
- On disk size is 3.9MiB with ~8000 events, and the compression is very
  effective.
- Disconnected and rejoined a machine, it rejoined fine.

### DIFF
--- a/src/exo/master/api.py
+++ b/src/exo/master/api.py
@@ -48,9 +48,6 @@ from exo.shared.constants import (
     EXO_MAX_CHUNK_SIZE,
     EXO_TRACING_CACHE_DIR,
 )
-
-_API_EVENT_LOG_DIR = EXO_EVENT_LOG_DIR / "api"
-
 from exo.shared.election import ElectionMessage
 from exo.shared.logging import InterceptLogger
 from exo.shared.models.model_cards import (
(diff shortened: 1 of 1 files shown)


src/exo/master/api.py
```diff
@@ -223,6 +225,8 @@ async def _log_requests(  # pyright: ignore[reportUnusedFunction]
 
     def reset(self, new_session_id: SessionId, result_clock: int):
         logger.info("Resetting API State")
+        self._event_log.close()
+        self._event_log = DiskEventLog(EXO_EVENT_LOG_DIR / "api")
```
rltakashige: Create a magic string for api instead of having it in two places?

aydenhailey78-hub:
Received and noted thank you

PR #1434, opened 2026-02-10
gossipsub: stop silent message dropping and warn

JakeHillion:
The 15-second publish_queue_duration caused messages in peer queues to be silently dropped. When events are dropped, workers detect gaps in the event index sequence and request missing events via the NACK path (RequestEventLog), but this recovery is inefficient.

Removed the timeout configuration - gossipsub now uses its default behavior without time-based eviction. If queue buildup is a concern, queue size should be limited explicitly rather than dropping by timeout.

Split error handling to log AllQueuesFullError as a warning (indicates peers are unresponsive) while keeping NoPeersSubscribedToTopicError silent (expected during startup and network partitions).

Test plan:
- CI

### DIFF
--- a/rust/networking/src/swarm.rs
+++ b/rust/networking/src/swarm.rs
@@ -106,7 +106,6 @@ mod behaviour {
     use crate::{alias, discovery};
     use libp2p::swarm::NetworkBehaviour;
     use libp2p::{gossipsub, identity};
-    use std::time::Duration;
 
     /// Behavior of the Swarm which composes all desired behaviors:
     /// Right now its just [`discovery::Behaviour`] and [`gossipsub::Behaviour`].
@@ -134,7 +133,6 @@ mod behaviour {
         gossipsub::Behaviour::new(
             MessageAuthenticity::Signed(keypair.clone()),
             ConfigBuilder::default()
-                .publish_queue_duration(Duration::from_secs(15))
                 .max_transmit_size(1024 * 1024)
                 .validation_mode(ValidationMode::Strict)
                 .build()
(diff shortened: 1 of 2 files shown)


src/exo/routing/router.py
```diff
@@ -207,10 +207,10 @@ async def _networking_publish(self):
                 try:
                     logger.trace(f"Sending message on {topic} with payload {data}")
                     await self._net.gossipsub_publish(topic, data)
-                # As a hack, this also catches AllQueuesFull
```
rltakashige, line 210: Don't think you want to remove this comment.

rust/networking/src/swarm.rs
```diff
@@ -134,7 +133,6 @@ mod behaviour {
         gossipsub::Behaviour::new(
             MessageAuthenticity::Signed(keypair.clone()),
             ConfigBuilder::default()
-                .publish_queue_duration(Duration::from_secs(15))
```
rltakashige, line 137: What's the default?

JakeHillion: Infinite - it doesn't drop events in the middle of the queues.

src/exo/routing/router.py
JakeHillion: Yeah, I think I do. It's not very helpful, and the warning should cover it - it's not a hack anymore!

rltakashige: Do we not want to recover from this error?

aydenhailey78-hub:
No issues on my end

djhdusjbd-eng:
Noted thanks

PR #1435, opened 2026-02-10
Dont save to app directory

rltakashige:
## Motivation

App keeps losing Local Network permissions.

## Changes

Don't save stuff to the app directory anymore. Instead, save to .exo.

### Manual Testing
Before:
image

After:
Screenshot 2026-02-10 at 17 43 06

### DIFF
--- a/app/EXO/EXO/ExoProcessController.swift
+++ b/app/EXO/EXO/ExoProcessController.swift
@@ -267,7 +267,6 @@ final class ExoProcessController: ObservableObject {
         let custom = customNamespace.trimmingCharacters(in: .whitespaces)
         return custom.isEmpty ? base : custom
     }
-
 }
 
 struct RuntimeError: LocalizedError {


app/EXO/EXO/ExoProcessController.swift
```diff
@@ -259,6 +263,65 @@ final class ExoProcessController: ObservableObject {
...
+        {
+            return settings
+        }
+
+        // Migration: read from UserDefaults if present
+        let defaults = UserDefaults.standard
+        let hasLegacy =
+            defaults.string(forKey: customNamespaceKey) != nil
+            || defaults.string(forKey: hfTokenKey) != nil
+            || defaults.object(forKey: enableImageModelsKey) != nil
```
AlexCheema: do we want this? what's legacy?

rltakashige: got rid of that 😆

PR #1441, opened 2026-02-10
feat: add intermediate model-fit state for cluster-capacity-only models

Hmbown:
## Motivation
In the model picker, non-runnable models were not clearly separated between two different cases:

- model exceeds currently available RAM but can fit in total cluster capacity
- model exceeds total cluster capacity

That made it harder to distinguish "not runnable right now" from "too large for this cluster."

## Changes
- Added a tri-state fit status in dashboard model-picker flow:
  - `fits_now`
  - `fits_cluster_capacity`
  - `too_large`
- Updated dashboard logic to compute both available cluster RAM and total cluster RAM.
- Passed fit status through picker components.
- Updated model size color mapping:
  - `fits_now` -> white/gray
  - `fits_cluster_capacity` -> orange
  - `too_large` -> red
- Updated group ordering for non-runnable models:
  - orange groups (`fits_cluster_capacity`) are listed above red groups (`too_large`).

## Why It Works
Launch safety is unchanged: selection is still gated by existing placement feasibility (`canModelFit`), so models that cannot run now remain disabled.

The new fit status is used for visual distinction and ordering only:
- runnable now
- fits cluster capacity but not free RAM now
- too large for cluster capacity

## Before
Non-runnable models were not clearly distinguished by temporary capacity vs hard capacity limit.

## After
Model picker clearly separates states by both color and order:
- Runnable now (white/gray)
- Fits cluster capacity but not free RAM now (orange, disabled)
- Exceeds cluster capacity (red, disabled)

### Manual testing
- Open model picker with live cluster memory telemetry.
- Verify white/gray models are selectable.
- Verify orange models are disabled and appear above red models.
- Verify red models are disabled and appear below orange models.

### Automated checks
- `npm --prefix dashboard run build` passes.
- `uv run basedpyright` passes.
- `uv run ruff check` passes.
- `npm --prefix dashboard run check` reports existing pre-change Svelte diagnostics (same known SVG title/a11y items).
- `uv run pytest` in this local environment exits during collection due existing `tests/start_distributed_test.py` `SystemExit` usage.
- No Python code was changed.


AlexCheema (changes requested):
good catch - we do want to distinguish these.

not sure about the colors though. does yellow make sense for not enough capacity? we tend to use yellow for other things. perhaps orange? or grey?

Hmbown:
great point - updated to orange and fixed typematch errors on +page.svelte

Hmbown:
updated: rebased on latest origin/main, switched middle state to orange, and verified selection guard still blocks non-runnable models. local checks run: basedpyright + ruff + dashboard build pass; dashboard check/pytest still show pre-existing repo issues noted in PR body. ready for re-review.

AlexCheema:
can you take some screenshots of what it looks like in various cases?

Hmbown:
Screenshot 2026-02-10 at 16 45 45
Screenshot 2026-02-10 at 16 46 30
Screenshot 2026-02-10 at 17 05 25

screenshots of oom orange models attached - in all + qwen/glm portions

PR #1443, opened 2026-02-10
nix: override apple-sdk to 26.2 and enable MLX_BUILD_CPU

JakeHillion:
The pinned nixpkgs provides apple-sdk 26.0, but building MLX requires SDK 26.2. The upstream package reads versions.json via a relative path at eval time, so it can't be overridden through callPackage args.

Added a thin overlay that copies the upstream apple-sdk source and patches only metadata/versions.json to point at SDK 26.2. Also enabled MLX_BUILD_CPU in the MLX nix build.

This avoids vendoring the entire apple-sdk package (~2200 lines) while still getting the SDK version we need.

Test plan:
- CI
- Built and ran on two machines connected with Thunderbolt 5 - Kimi K2.5 starts in Tensor+RDMA and seems sensible.

### DIFF
--- a/flake.nix
+++ b/flake.nix
@@ -83,6 +83,9 @@
           _module.args.pkgs = import inputs.nixpkgs {
             inherit system;
             config.allowUnfreePredicate = pkg: (pkg.pname or "") == "metal-toolchain";
+            overlays = [
+              (import ./nix/apple-sdk-overlay.nix)
+            ];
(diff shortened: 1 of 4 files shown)


Evanev7:
No complaints, the hard part was getting it working, I'm happy we can keep it working with 50 lines instead of 2000.

PR #1446, opened 2026-02-10
util: remove VecExt trait, inline at call site

JakeHillion:
VecExt added a .map() convenience method on Vec<T> that simply called .into_iter().map(f).collect(). This thin wrapper provided no optimisation benefit and obscured a standard iterator pattern behind a nightly feature gate and an extra dependency.

Replaced the single call site in exo_pyo3_bindings with the equivalent iterator chain and removed the ext module, the extend dependency, and the trait_alias feature gate from the util crate.

Test plan:
- CI


Evanev7:
im never going to get my rust rewrite merged if these are the changes we're making lmao

PR #1448, opened 2026-02-11
Make info gatherer monitors resilient with retry loops and timeouts

AlexCheema:
## Motivation

Info gatherer monitors could silently stop posting events, causing stale node state after rejoins. The macmon monitor was especially fragile — it had no retry loop, so a crash or parse error would kill it permanently. Worse, the unhandled exception would propagate to the TaskGroup and take down *all* sibling monitors. Additionally, none of the monitors had timeouts on their subprocess calls, so a hung `system_profiler` or `networksetup` could stall a monitor indefinitely.

## Changes

- Wrap `_monitor_macmon` in a `while True` retry loop with `except Exception`, matching the pattern used by all other monitors
- Add `fail_after` timeouts to all monitor loop bodies:
  - 10s for lightweight commands (`_monitor_misc`, `_watch_system_info`, `_gather_iface_map` init)
  - 30s for heavier commands (`_monitor_system_profiler_thunderbolt_data`, `_monitor_thunderbolt_bridge_status`)
- Remove unused `CalledProcessError` and `cast` imports

## Why It Works

All monitors now follow the same resilient pattern: `while True` → `try` with `fail_after` → `except Exception` (logs warning) → `sleep`. If a subprocess hangs, the timeout fires and `TimeoutError` is caught by the existing `except Exception` handler. If macmon crashes, it restarts after the interval instead of dying permanently. No single monitor failure can cascade to kill the others.

### Manual Testing

- Run exo, kill macmon process (`kill $(pgrep macmon)`), verify it restarts and metrics resume
- Verify all monitors continue posting events after simulated hangs

### Automated Testing
- All 188 existing tests pass
- basedpyright: 0 errors
- ruff: all checks passed

### DIFF
--- a/src/exo/utils/info_gatherer/info_gatherer.py
+++ b/src/exo/utils/info_gatherer/info_gatherer.py
@@ -8,16 +8,17 @@ from subprocess import CalledProcessError
 from typing import Self, cast
 
 import anyio
-from anyio import create_task_group, fail_after, open_process
+from anyio import create_task_group, fail_after, open_process, to_thread
 from anyio.abc import TaskGroup
 from anyio.streams.buffered import BufferedByteReceiveStream
 from anyio.streams.text import TextReceiveStream
 from loguru import logger
 from pydantic import ValidationError
 
-from exo.shared.constants import EXO_CONFIG_FILE
+from exo.shared.constants import EXO_CONFIG_FILE, EXO_MODELS_DIR
 from exo.shared.types.memory import Memory
 from exo.shared.types.profiling import (
+    DiskUsage,
     MemoryUsage,
     NetworkInterfaceInfo,
     ThunderboltBridgeStatus,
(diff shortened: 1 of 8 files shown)


src/exo/utils/info_gatherer/info_gatherer.py
```diff
@@ -461,26 +469,23 @@ async def _monitor_macmon(self, macmon_path: str):
...
+                    ]
+                ) as p:
+                    if not p.stdout:
+                        logger.critical("MacMon closed stdout")
+                        return
+                    async for text in TextReceiveStream(
+                        BufferedByteReceiveStream(p.stdout)
+                    ):
+                        await self.info_sender.send(MacmonMetrics.from_raw_json(text))
+            except Exception as e:
```
rltakashige, line 555: Can you explicitly continue catching CalledProcessError as the old code used to?

AlexCheema: done.

PR #1450, opened 2026-02-11
dashboard: show available disk space on downloads page

JakeHillion:
The downloads page previously only showed the approximate space used by downloaded models (summed from completed download sizes), but did not show how much disk space was actually available. This made it difficult to know if a download would succeed before pressing the button.

Added disk space tracking to the InfoGatherer that polls the models directory partition every 30 seconds. The DiskUsage type captures total and available space, which flows through the event system to State and is exposed via the /state API. The dashboard now displays "X on disk / Y available" for each node in the downloads view.

Test plan:
- CI

image

It doesn't look fantastic, but the data is right and hopefully we make this page look better in the near future anyway.


AlexCheema (approved):
looks good.
also im not sure if necessary but im coding more defensively with these things now: can we add timeout for await NodeDiskUsage.gather() like in #1448

PR #1451, opened 2026-02-11
refactor: replace timestamp-based node liveness with event-index staleness

AlexCheema:
## Motivation

The node liveness system uses `last_seen: Mapping[NodeId, datetime]` in the event-sourced State and checks `now - last_seen > 30s` to time out nodes. This has two fundamental problems:

1. **Timestamps don't belong in event-sourced state.** `datetime.now()` is a side effect — replaying the same events produces different state depending on *when* they're replayed. The master was even overwriting the worker's timestamp with its own clock at indexing time (`event.when = str(datetime.now(...))`), adding another source of non-determinism.

2. **The timeout is finicky.** The 30-second wall-clock threshold fires whenever *any* info-gathering task is slow — not when the node is actually unreachable. A node can be perfectly connected via libp2p, actively participating in topology, but get removed because `system_profiler` took 31 seconds to respond. This causes false-positive node removals in production.

## Changes

**State** (`state.py`):
- Replace `last_seen: Mapping[NodeId, datetime]` with `last_event_index_by_node: Mapping[NodeId, int]` — stores the global event index at which each node last produced an event. Fully deterministic: derived from the event stream itself.

**Events** (`events.py`):
- Remove `when: str` field from `NodeGatheredInfo` — no longer needed.
- Rename `NodeTimedOut` → `NodeDisconnected` — reflects the actual semantics (the node stopped producing events, not "we checked the clock").

**Apply** (`apply.py`):
- Top-level `apply()` now records `last_event_index_by_node[node_id] = event.idx` whenever a `NodeGatheredInfo` is applied. This keeps index tracking as a concern of the indexed event layer, not the individual event handler.
- `apply_node_gathered_info()` no longer touches liveness state.
- `apply_node_timed_out()` → `apply_node_disconnected()` with `last_event_index_by_node` cleanup.

**Master** (`master/main.py`):
- Remove all `datetime`/`timedelta`/`timezone` imports and the `event.when` overwrite in `_event_processor`.
- Replace the timestamp check in `_plan()` with cycle-based staleness detection: each plan cycle (10s), the master snapshots `last_event_index_by_node` and compares to the previous snapshot. If a node's index hasn't advanced for 3 consecutive cycles, it's disconnected.
- Staleness tracking (`_last_checked_indices`, `_stale_cycles`) is local master state — ephemeral, not event-sourced, reset on re-election.

**Worker** (`worker/main.py`):
- Remove `when` field from `NodeGatheredInfo` construction. Remove unused `datetime` import.

## Why It Works

The old system asked: *"Has this node's timestamp been updated within 30 wall-clock seconds?"* — conflating "slow info gathering" with "node is dead."

The new system asks: *"Has this node produced ANY event since the last time I checked?"* — which is the actual question we care about. A node running macmon at 1s intervals, memory polling, network monitoring, etc. produces many events per second. If *none* of those events arrive for 3 consecutive 10-second plan cycles, the node is genuinely unreachable — not just slow at one particular task.

The event index is deterministic and monotonically increasing. It's derived from the event stream itself, making it a natural fit for event-sourced state. No wall clocks, no timezone issues, no race between worker timestamps and master timestamps.

### Automated Testing
- `test_master` updated: removed `when` field from `NodeGatheredInfo` fixture — passes.
- `basedpyright`: 0 errors, 0 warnings.
- `ruff check`: all passed.
- `nix fmt`: 0 files changed.

### DIFF
--- a/.mlx_typings/mlx_lm/models/step3p5.pyi
+++ b/.mlx_typings/mlx_lm/models/step3p5.pyi
@@ -1,0 +1,151 @@
+from dataclasses import dataclass
+from typing import Any, Dict, List, Optional
+
+import mlx.core as mx
+import mlx.nn as nn
+
+from .base import BaseModelArgs
+from .switch_layers import SwitchGLU
+
+@dataclass
+class ModelArgs(BaseModelArgs):
+    model_type: str
+    hidden_size: int
+    num_hidden_layers: int
+    vocab_size: int
+    num_attention_heads: int
+    num_attention_groups: int
+    head_dim: int
+    intermediate_size: int
+    rms_norm_eps: float
+    rope_theta: float
+    rope_scaling: Optional[Dict[str, Any]]
+    max_position_embeddings: int
+    sliding_window: int
+    layer_types: Optional[List[str]]
+    yarn_only_types: Optional[List[str]]
+    partial_rotary_factors: Optional[List[float]]
+    attention_other_setting: Optional[Dict[str, Any]]
+    use_head_wise_attn_gate: bool
+    moe_num_experts: int
+    moe_top_k: int
+    moe_intermediate_size: int
+    share_expert_dim: int
+    moe_layers_enum: Optional[str]
+    moe_router_scaling_factor: float
+    norm_expert_weight: bool
+    swiglu_limits: Optional[List[float]]
+    swiglu_limits_shared: Optional[List[float]]
+    tie_word_embeddings: bool
+
+class Step3p5MLP(nn.Module):
+    hidden_size: int
+    intermediate_size: int
+    gate_proj: nn.Linear
+    up_proj: nn.Linear
+    down_proj: nn.Linear
+    limit: Optional[float]
+
+    def __init__(
+        self, args: ModelArgs, intermediate_size: int, swiglu_limit: float = 0
+    ) -> None: ...
+    def __call__(self, x: mx.array) -> mx.array: ...
+
+class Step3p5MoEGate(nn.Module):
+    top_k: int
+    n_routed_experts: int
+    routed_scaling_factor: float
+    norm_topk_prob: bool
+    gate: nn.Linear
+    router_bias: mx.array
+
+    def __init__(self, args: ModelArgs) -> None: ...
+    def __call__(self, x: mx.array) -> tuple[mx.array, mx.array]: ...
+
+class Step3p5MoE(nn.Module):
+    gate: Step3p5MoEGate
+    switch_mlp: SwitchGLU
+    share_expert: Step3p5MLP
+    sharding_group: Optional[mx.distributed.Group]
+
+    def __init__(self, args: ModelArgs, layer_idx: int) -> None: ...
+    def __call__(self, x: mx.array) -> mx.array: ...
+
+class Step3p5Attention(nn.Module):
+    is_sliding: bool
+    num_heads: int
+    num_kv_heads: int
+    head_dim: int
+    scale: float
+    q_proj: nn.Linear
+    k_proj: nn.Linear
+    v_proj: nn.Linear
+    o_proj: nn.Linear
+    q_norm: nn.Module
+    k_norm: nn.Module
+    use_head_wise_attn_gate: bool
+    g_proj: nn.Linear
+    rope: nn.Module
+
+    def __init__(self, args: ModelArgs, layer_idx: int) -> None: ...
+    def __call__(
+        self,
+        x: mx.array,
+        mask: Optional[mx.array] = None,
+        cache: Optional[Any] = None,
+    ) -> mx.array: ...
+
+class Step3p5DecoderLayer(nn.Module):
+    self_attn: Step3p5Attention
+    is_sliding: bool
+    is_moe_layer: bool
+    mlp: Step3p5MLP | Step3p5MoE
+    input_layernorm: nn.Module
+    post_attention_layernorm: nn.Module
+
+    def __init__(self, args: ModelArgs, layer_idx: int) -> None: ...
+    def __call__(
+        self,
+        x: mx.array,
+        mask: Optional[mx.array] = None,
+        cache: Optional[Any] = None,
+    ) -> mx.array: ...
+
+class Step3p5Model(nn.Module):
+    args: ModelArgs
+    vocab_size: int
+    num_layers: int
+    embed_tokens: nn.Embedding
+    layers: list[Step3p5DecoderLayer]
+    norm: nn.Module
+    _swa_idx: Optional[int]
+    _full_idx: Optional[int]
+
+    def __init__(self, args: ModelArgs) -> None: ...
+    def __call__(
+        self,
+        x: mx.array,
+        cache: Optional[List[Any]] = None,
+    ) -> mx.array: ...
+
+class Model(nn.Module):
+    args: ModelArgs
+    model_type: str
+    model: Step3p5Model
+    lm_head: nn.Linear
+
+    def __init__(self, args: ModelArgs) -> None: ...
+    def __call__(
+        self,
+        inputs: mx.array,
+        cache: Optional[List[Any]] = None,
+    ) -> mx.array: ...
+    def sanitize(self, weights: dict[str, Any]) -> dict[str, Any]: ...
+    def shard(self, group: Optional[mx.distributed.Group] = None) -> None: ...
+    @property
+    def layers(self) -> list[Step3p5DecoderLayer]: ...
+    def make_cache(self) -> list[Any]: ...
+    @property
+    def cast_predicate(self) -> Any: ...
+    @property
+    def quant_predicate(self) -> Any: ...
--- a/bench/exo_bench.py
+++ b/bench/exo_bench.py
@@ -19,6 +19,11 @@ from urllib.parse import urlencode
 from loguru import logger
 from transformers import AutoTokenizer
 
+# Backoff constants for cluster settling retry
+_SETTLE_INITIAL_BACKOFF_S = 1.0
+_SETTLE_MAX_BACKOFF_S = 60.0
+_SETTLE_BACKOFF_MULTIPLIER = 2.0
+
 # Monkey-patch for transformers 5.x compatibility
 # Kimi's tokenization_kimi.py imports bytes_to_unicode from the old location
 # which was moved in transformers 5.0.0rc2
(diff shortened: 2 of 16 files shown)


Evanev7:
just throwing out that claude's analysis is straight up wrong here so please understand if this change is what you want before making it. "i haven't seen an event in 30s" is the outcome either way, this obfuscates that imo. Also the master timestamp was an intentional inclusion for debugging that is being removed here.

AlexCheema:
The motivation for this change is that consumer device timestamps should not be trusted. Event-index staleness gives us a reliable, monotonic measure of liveness that doesn't depend on clock synchronization across nodes.

AlexCheema:
To clarify — we're fine keeping the master timestamp for our own debugging purposes, but it shouldn't be used for liveness detection logic. The core liveness check should rely on event-index staleness, not timestamps from consumer devices that can drift or be out of sync.

AlexCheema:
Code Review: PR #1451 — refactor: replace timestamp-based node liveness with event-index staleness
===================================================================================================

Author: AlexCheema
Status: OPEN (auto-merge disabled)
Changes: +41 / -37 across 6 files

Overview
--------
Replaces wall-clock timestamp-based node liveness (last_seen: datetime, 30s
timeout) with event-index staleness detection. Instead of checking "has this
node's timestamp been updated within 30 seconds?", it checks "has this node
produced any new events in the last 3 plan cycles (~30s)?".

Files changed:
  - src/exo/shared/types/state.py — last_seen → last_event_index_by_node
  - src/exo/shared/types/events.py — NodeTimedOut → NodeDisconnected, remove when field
  - src/exo/shared/apply.py — update apply logic for new state shape
  - src/exo/master/main.py — new staleness detection in _plan(), remove datetime usage
  - src/exo/worker/main.py — remove when field from NodeGatheredInfo construction
  - src/exo/master/tests/test_master.py — remove when field from fixture

Motivation
----------
SOUND. Two real problems with the old approach:
1. datetime.now() in event-sourced state breaks deterministic replay
2. 30s wall-clock timeout fires on slow info-gathering (e.g. system_profiler
   taking 31s), not actual node unreachability — causing false-positive removals

The new approach is deterministic (event indices are derived from the event
stream) and more robust (checks event production, not clock deltas).

Correctness — State Changes
----------------------------
PASS.

state.py: Replacing `last_seen: Mapping[NodeId, datetime]` with
`last_event_index_by_node: Mapping[NodeId, int]` is clean. Integer indices are
deterministic and monotonically increasing.

events.py: Removing the `when: str` field from NodeGatheredInfo eliminates
the non-deterministic timestamp from the event stream. The rename
NodeTimedOut → NodeDisconnected better reflects semantics.

Correctness — Apply Logic
--------------------------
PASS.

In apply() (top-level), the index tracking is done at the IndexedEvent layer:
    if isinstance(event.event, NodeGatheredInfo):
        update["last_event_index_by_node"] = {
            **new_state.last_event_index_by_node,
            event.event.node_id: event.idx,
        }
This is correct — the global event index is available on the IndexedEvent
wrapper, and the update is applied atomically with last_event_applied_idx.

apply_node_disconnected() properly cleans up last_event_index_by_node,
topology, downloads, node_identities, node_memory, node_disk, instances,
and runners for the disconnected node. This matches the old
apply_node_timed_out() behavior.

Correctness — Staleness Detection
-----------------------------------
PASS with notes.

The new logic in _plan():
  1. Snapshot current_indices from state.last_event_index_by_node
  2. Compare each node's index to _last_checked_indices (default -1)
  3. If unchanged → increment _stale_cycles counter
  4. If changed → reset counter
  5. If stale for >= 3 cycles → emit NodeDisconnected

This is correct:
  - New nodes: index > -1, so not stale on first check ✓
  - list() copies prevent mutation-during-iteration issues ✓
  - del _stale_cycles[node_id] after disconnect prevents double-fire ✓
  - _last_checked_indices updated AFTER the comparison loop ✓

Race Condition Analysis
-----------------------
NONE. _plan() runs in a single async task. _last_checked_indices and
_stale_cycles are only accessed within _plan(). No concurrent mutation.

Master Re-election
------------------
SAFE. On re-election, a new Master instance is created with fresh empty dicts:
    _last_checked_indices: dict[NodeId, int] = {}
    _stale_cycles: dict[NodeId, int] = {}
First cycle: all nodes compare against -1 (default), so none are marked stale.
Staleness tracking resumes normally from cycle 2. Effective grace period of
~10s after re-election before staleness counting begins.

Edge Case — Only NodeGatheredInfo Updates the Index
----------------------------------------------------
NOTE. The apply() function only updates last_event_index_by_node for
NodeGatheredInfo events. Other node-originated events (RunnerStatusUpdated,
NodeDownloadProgress, etc.) do NOT update the index.

This means a node that is actively running inference (producing
RunnerStatusUpdated events) but whose InfoGatherer has stalled would still be
disconnected.

In practice this is LOW RISK because:
  - macOS nodes: macmon runs at ~1s intervals, thunderbolt at ~5s
  - Linux nodes: psutil memory at ~1s, disk at ~30s
  - A healthy InfoGatherer produces events every 1-2 seconds

However, if InfoGatherer crashes while the runner continues working, the node
would be falsely disconnected after ~30s. Consider whether other event types
with a node_id should also update the index — though NodeGatheredInfo is
arguably the right "heartbeat" signal.

Timing Comparison
-----------------
Old: Exact 30-second wall-clock check (but subject to clock skew)
New: 3 × 10-second plan cycles ≈ 30 seconds (but subject to plan loop drift)

The effective timeout is 20-40 seconds depending on when the node stops
producing events relative to the plan cycle. This is comparable to the old
behavior and arguably more predictable since it doesn't depend on clock sync.

Dashboard Compatibility
-----------------------
PASS. No references to `last_seen` or `lastSeen` in the dashboard source code.
The dashboard does not consume this state field directly.

Backwards Compatibility
-----------------------
BREAKING for any external consumers that:
  - Parse the `when` field from NodeGatheredInfo events
  - Listen for NodeTimedOut events (now NodeDisconnected)
  - Read state.last_seen (now state.last_event_index_by_node)

These are internal types, so the risk is low unless there are external
integrations consuming the event stream or state endpoint directly.

Test Coverage
-------------
ADEQUATE. The test_master.py fixture is updated to remove the `when` field.
No new tests for the staleness detection logic itself (_plan loop behavior).
The staleness logic is simple enough that the lack of a unit test is acceptable,
though a test verifying the 3-cycle threshold would add confidence.

Nits
----
None.

Verdict
-------
LGTM. Well-motivated refactor that eliminates non-deterministic timestamps from
event-sourced state. The staleness detection logic is correct with no race
conditions. The only notable design choice is that only NodeGatheredInfo events
update the index (not all node-originated events), which is a reasonable
"heartbeat" approach but could theoretically cause false disconnects if
InfoGatherer fails while runners are healthy. Low risk in practice.

AlexCheema:
## Code Review -- PR #1451: refactor: replace timestamp-based node liveness with event-index staleness

**CI status:** All 8 checks PASSED (typecheck, Build and check on aarch64-darwin, aarch64-linux, x86_64-linux)

**Merge status:** CONFLICTING -- this PR has merge conflicts with `main` and needs a rebase.

### Overview

Replaces wall-clock `datetime`-based node liveness (`last_seen: Mapping[NodeId, datetime]`, 30s timeout) with event-index staleness detection (`last_event_index_by_node: Mapping[NodeId, int]`, 3 consecutive plan cycles with no new events). The motivation is sound: `datetime.now()` is a side effect that breaks deterministic event replay, and the 30s wall-clock timeout conflates "slow info gathering" with "node actually unreachable."

Changes span 6 files: `state.py`, `events.py`, `apply.py`, `master/main.py`, `worker/main.py`, `test_master.py`.

### Critical Issues

**1. Merge conflict with #1493 -- `node_identities` cleanup regression**

PR #1493 ("don't time out node identities", merged today) intentionally removed `node_identities` cleanup from `apply_node_timed_out` so that nodes rejoining the cluster retain their identity data. However, this PR's `apply_node_disconnected` still includes `node_identities` cleanup:

```python
node_identities = {
    key: value
    for key, value in state.node_identities.items()
    if key != event.node_id
}
```

When rebasing, the `node_identities` cleanup **must** be removed to preserve the behavior from #1493. GitHub reports `mergeStateStatus: DIRTY`, confirming the conflict.

**2. `_master_time_stamp` assignment silently removed**

The PR removes this line from `_event_processor`:
```python
event._master_time_stamp = datetime.now(tz=timezone.utc)  # pyright: ignore[reportPrivateUsage]
```

The `_master_time_stamp` field on `BaseEvent` still exists in `events.py` (untouched by this PR), but the only place it was ever *set* is now deleted. As Evanev7 noted in the PR comments, this was an intentional debugging aid. Alex agreed it should be kept for debugging but not used for liveness.

**Recommended fix:** Restore the `_master_time_stamp` assignment (keeping the `datetime` import in `master/main.py` for this one line). It's a private debug field with zero impact on liveness logic. Alternatively, if the team decides it's no longer needed, remove the field definition from `BaseEvent` too for consistency.

### Significant Issues

**3. Only `NodeGatheredInfo` events update `last_event_index_by_node`**

In `apply()`:
```python
if isinstance(event.event, NodeGatheredInfo):
    update["last_event_index_by_node"] = {
        **new_state.last_event_index_by_node,
        event.event.node_id: event.idx,
    }
```

Other node-originated events (`RunnerStatusUpdated`, `NodeDownloadProgress`, `TaskStatusUpdated`) do NOT update the index. If a node's InfoGatherer crashes while its runner continues working, the node would be falsely disconnected after ~30s.

In practice this is **low risk** because the InfoGatherer produces events from multiple independent sources:
- macOS: macmon at 1s, system_profiler at 5s, thunderbolt bridge at 10s, disk at 30s
- Linux: psutil memory at 1s, network interfaces at 10s, disk at 30s

ALL of these would have to fail simultaneously to miss the heartbeat window. But this is a design decision worth making explicitly -- a comment noting that `NodeGatheredInfo` serves as the heartbeat signal would help future readers.

### Minor Issues

**4. Detection latency is variable (20-40s, not "~30s")**

The effective detection window depends on when the node stops producing events relative to the plan cycle boundary. The PR description and comments say "~30s" -- more precisely it's 20-40 seconds, comparable to the old behavior.

**5. No unit test for the staleness detection logic**

The `_plan()` method's staleness tracking (`_last_checked_indices`, `_stale_cycles`, the 3-cycle threshold) has no dedicated test. The logic is simple, but a test verifying the 3-cycle threshold would add confidence for the core liveness mechanism.

### Correctness Analysis

The staleness detection logic in `_plan()` is correct:
- New nodes: `_last_checked_indices.get(node_id, -1)` returns -1, which won't match any real index, so new nodes are never falsely marked stale on first check.
- `list(current_indices)` prevents mutation-during-iteration.
- `del _stale_cycles[node_id]` after disconnect prevents double-fire.
- `_last_checked_indices = current_indices` is set AFTER the comparison loop, ensuring correct before/after comparison.
- No race conditions: `_plan()` runs in a single async task; staleness dicts are only accessed there.
- Master re-election: fresh `Master` instance gets empty dicts, giving a natural ~10s grace period before staleness counting begins.

The `apply_node_disconnected` cleanup of topology, downloads, memory, disk, system, network, thunderbolt, and rdma_ctl is complete and correct (modulo the `node_identities` issue in #1 above).

### What's Good

- **Eliminates non-determinism from event-sourced state.** The `datetime.now()` calls in `apply_node_gathered_info` and the `when` field overwrite in `_event_processor` were genuine event-sourcing violations. Event indices are fully deterministic.
- **`NodeDisconnected` is a better name** than `NodeTimedOut` -- reflects actual semantics.
- **Clean removal of the `when: str` field** from `NodeGatheredInfo`. The old field was awkward (manually cast datetime, overwritten by master, had a TODO comment).
- **Staleness tracking correctly scoped as ephemeral master state**, not event-sourced. Resets on re-election, which is the right behavior.
- **Dashboard unaffected.** No references to `last_seen`/`lastSeen` in the frontend code.
- **Net code reduction** (+41/-37) despite adding new functionality.

### Verdict

Sound design that eliminates a real problem. The implementation is correct, but the PR needs work before merge:

1. **Rebase required** -- merge conflicts with `main`
2. **Drop `node_identities` cleanup** from `apply_node_disconnected` to match #1493
3. **Restore `_master_time_stamp` assignment** or remove the field from `BaseEvent` entirely

After addressing those three items, this is ready.

PR #1452, opened 2026-02-11
fix: prevent DownloadModel TaskCreated event flood

AlexCheema:
## Motivation

When a model download fails repeatedly (e.g. `ContentLengthError` on a large model like `zai-org/GLM-5`), the download coordinator accumulates duplicate progress callbacks — one per retry cycle. Each callback independently throttles at 1 event/sec, so after N retries, every download progress tick generates N events instead of 1. After an hour of failures (~60 retry cycles), this produces ~60 `NodeDownloadProgress` events/sec, overwhelming the master, delaying heartbeats, and causing the node to time itself out.

### The callback accumulation cycle
1. `_start_download_task()` calls `shard_downloader.on_progress(callback)` which **appends** to a list
2. Download fails → `DownloadFailed` status set, but old callback stays in the list
3. 60s later: `_emit_existing_download_progress()` scans disk → resets status to `DownloadPending`
4. Worker sends new `StartDownload` → coordinator accepts (guard didn't check `DownloadFailed`)
5. `_start_download_task()` appends **another** callback
6. Each callback has its own throttle → N callbacks = N events per progress tick

### Commit 1: `src/exo/worker/main.py`
Move the `DownloadModel` backoff check **before** `TaskCreated` emission in `plan_step()`. Previously `TaskCreated` was emitted unconditionally every 0.1s even when backoff blocked the download command.

### Commit 2: `src/exo/download/coordinator.py`
1. **Register progress callback once** in `__post_init__` instead of per-download in `_start_download_task()`. Uses a per-model throttle dict instead of per-callback closure variables.
2. **Add `DownloadFailed` to the `_start_download()` guard** so redundant `_start_download_task()` calls don't happen. Retries still work because `_emit_existing_download_progress` resets `DownloadFailed` → `DownloadPending` by scanning disk every 60s.

## Why It Works

The root cause was callbacks accumulating in `ResumableShardDownloader.on_progress_callbacks` (a list that only appends, never clears). By registering one callback per coordinator lifetime and guarding against re-entry on `DownloadFailed`, we ensure exactly one progress event per model per progress tick regardless of how many retry cycles have occurred.

### Manual Testing
- Verified the download retry flow: failed download → 60s scan resets status → new `StartDownload` accepted → download retries with single callback

### Automated Testing
- `uv run pytest` — 188 passed

### DIFF
--- a/src/exo/download/coordinator.py
+++ b/src/exo/download/coordinator.py
@@ -284,6 +284,12 @@ class DownloadCoordinator:
                     _,
                     progress,
                 ) in self.shard_downloader.get_shard_download_status():
+                    model_id = progress.shard.model_card.model_id
+
+                    # Active downloads emit progress via the callback — don't overwrite
+                    if model_id in self.active_downloads:
+                        continue
+
                     if progress.status == "complete":
                         status: DownloadProgress = DownloadCompleted(
                             node_id=self.node_id,


rltakashige (approved):
I like the fact we're throttling the download models sent to the event log

PR #1461, opened 2026-02-12
Pass usage and generation stats through all adapters correctly

rltakashige:
## Motivation

Exo is not returning usage stats correctly at the moment.

## Changes

- Correctly pass usage stats instead of generation stats.
- Pass usage stats within tool calls.

### Manual Testing
Needs manual testing.

### Automated Testing
Passes CI.


Evanev7 (approved):
This could merge as is, but how about instead of threading through usage like this, we instead have a dedicated usage/stats response?

rltakashige:
Discussed in person, but the ordering logic for a Usage Stats Response could get messy, so will merge this for now.

PR #1466, opened 2026-02-12
fix: forcefully terminate runner on Shutdown timeout

AlexCheema:
## Motivation

When an instance is deleted while its runner is in the `RunnerLoading` state (actively loading model weights via `mx.eval()`), the system doesn't handle it cleanly. The Worker sends a `Shutdown` task, but the runner is blocked in `mx.eval()` and can't process it. The 3-second timeout fires, the runner is removed from `self.runners`, but **the process keeps running** — holding GPU/memory resources as an orphan until loading eventually completes.

## Changes

When the Shutdown task times out, call `supervisor.shutdown()` to forcefully terminate the runner process. This:
- Closes IPC channels
- Waits 1s for graceful exit
- Sends SIGTERM (waits 1s)
- Sends SIGKILL if needed

The supervisor's health check (from #1464) detects the process death and cleans up the `run()` coroutine.

## Why It Works

`mx.eval()` is a blocking C call that cannot be interrupted from Python. The runner subprocess is the isolation boundary — killing the process is the correct way to reclaim resources. Process death guarantees all tensors/GPU buffers are freed by the OS. The health check infrastructure (concurrent `_health_check` loop) detects the death within ~1 second and lets the supervisor's `run()` coroutine return, preventing orphaned tasks in the Worker's task group.

### Manual Testing

- Delete an instance while model is loading
- Verify runner process is killed within seconds
- Verify memory is reclaimed

### Automated Testing
- Existing `test_runner_supervisor.py` tests cover process death detection
- Depends on health check PR #1464 for the detection mechanism

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -132,7 +132,7 @@ markers = [
 env = [
   "EXO_TESTS=1"
 ]
-addopts = "-m 'not slow'"
+addopts = "-m 'not slow' --ignore=tests/start_distributed_test.py"
 filterwarnings = [
     "ignore:builtin type Swig:DeprecationWarning",
 ]


AlexCheema:

Small fix (+9/-4) for orphaned runner processes:
- When an instance is deleted while its runner is loading (`mx.eval()` blocking), the `Shutdown` task can't be processed
- Previously: 3s timeout fires, runner removed from dict, but **process keeps running** as an orphan holding GPU/memory
- Now: on timeout, calls `supervisor.shutdown()` which sends SIGTERM → waits 1s → SIGKILL, ensuring the process is actually killed
- Also renames `runner` → `supervisor` variable for clarity and adds a warning log

Correct approach — `mx.eval()` is a blocking C call that can't be interrupted from Python, so process termination is the right mechanism. Clean diff.

AlexCheema (commented):
## Code Review: PR #1466 — Forcefully terminate runner on Shutdown timeout

### Overall Assessment
Small, focused fix that addresses a real bug: when a runner doesn't respond to a Shutdown task within 3 seconds, the process was left dangling. The fix is correct and well-scoped.

### Analysis

**The Bug (before this PR):**
```python
case Shutdown(runner_id=runner_id):
    runner = self.runners.pop(runner_id)
    try:
        with fail_after(3):
            await runner.start_task(task)
    except TimeoutError:
        # Only emits TimedOut event — never calls shutdown()!
        await self.event_sender.send(...)
    finally:
        runner.shutdown()
```

The `finally: runner.shutdown()` was always called, but the problem was that `runner.shutdown()` attempts a graceful shutdown first (`join(1)` → `terminate()` → `kill()`). When the runner is already unresponsive (timeout on `start_task`), the graceful path wastes time.

**The Fix:**
```python
except TimeoutError:
    logger.warning(...)
    supervisor.shutdown()  # Force terminate BEFORE emitting event
    await self.event_sender.send(...)
finally:
    supervisor.shutdown()  # Still called in finally for normal path
```

### Strengths

1. **Correct**: `supervisor.shutdown()` is idempotent — calling it twice (once in `except`, once in `finally`) is safe because the second call finds the process already dead.

2. **Good logging**: The warning message includes `runner_id` for debugging.

3. **Variable rename** `runner` → `supervisor`: Accurate — the object is a `RunnerSupervisor`, not a runner.

### Observations

4. **`supervisor.shutdown()` is called twice on timeout path**: Once in the `except TimeoutError` block and once in the `finally` block. While safe (idempotent), the `finally` call on the timeout path is redundant. Not a bug, just slightly redundant work (two `join(1)` calls, two `is_alive()` checks). Negligible cost.

5. **3-second timeout is hardcoded**: The `fail_after(3)` is a magic number. For consistency with other timeout constants in the codebase (e.g., `HEALTH_CHECK_INTERVAL_SECONDS`, `HEARTBEAT_STALE_CHECKS`), consider extracting to a named constant. Minor nit.

6. **`pyproject.toml` change**: The `--ignore=tests/start_distributed_test.py` addition seems unrelated to the runner shutdown fix. Should this be in a separate commit/PR?

### Verdict
**Approve.** Clean, correct fix for a real process lifecycle bug. The double `shutdown()` call is harmless.

AlexCheema:
## Code Review: PR #1466 — Forcefully terminate runner on Shutdown timeout

### Core Fix — Correct approach
The fix addresses a real bug: when a runner is blocked in `mx.eval()` and can't process a Shutdown task, the 3s timeout fires but the process was left running as an orphan. Calling `supervisor.shutdown()` in the `except TimeoutError` block to force-kill the process is the right approach.

The `runner` → `supervisor` rename and added warning log are good improvements.

### Bug: `shutdown()` is not idempotent — second call will crash

On the timeout path, `shutdown()` is called **twice**: once in `except TimeoutError`, once in `finally`. The second call will raise an unhandled exception:

1. `self._cancel_sender.send(TaskId("CANCEL_CURRENT_TASK"))` ([runner_supervisor.py:104](https://github.com/exo-explore/exo/blob/main/src/exo/worker/runner/runner_supervisor.py)) raises `ClosedResourceError` because `MpSender.send()` checks `self._state.closed.is_set()` ([channels.py:119](https://github.com/exo-explore/exo/blob/main/src/exo/utils/channels.py)) — the first `shutdown()` already closed this sender.

2. Even if that were caught, `MpSender.close()` calls `self._state.buffer.put(_MpEndOfStream())` on an already-closed `multiprocessing.Queue` ([channels.py:135](https://github.com/exo-explore/exo/blob/main/src/exo/utils/channels.py)), which raises `ValueError`.

**Suggested fix** — either:
- **(a)** Add an idempotency guard to `shutdown()`: `if self._shutdown: return` at the top (with a `_shutdown: bool = False` field)
- **(b)** Restructure try/except/finally to avoid the double call — e.g., remove `shutdown()` from `finally` and call it explicitly in both the happy path and the except path

### Minor: unrelated `pyproject.toml` change

The `--ignore=tests/start_distributed_test.py` addition to pytest `addopts` is unrelated to the runner shutdown fix. Consider splitting into a separate commit for clean history.

---

AlexCheema:
## Code Review — PR #1466: fix: forcefully terminate runner on Shutdown timeout

**CI: All passing**

### Overview

+9/-4 across 2 files. Fixes orphaned runner processes when `Shutdown` task times out (e.g., runner blocked in `mx.eval()` during loading). Adds `supervisor.shutdown()` in the `except TimeoutError` block so the process is killed before emitting the `TimedOut` event.

### Bug: `shutdown()` is called twice on the timeout path

On the timeout path, `shutdown()` is called in `except` (line ~235) AND in `finally` (line ~243). The second call crashes:

1. `self._cancel_sender.send(TaskId("CANCEL_CURRENT_TASK"))` (runner_supervisor.py:188) raises `ClosedResourceError` because `MpSender.send()` checks `self._state.closed.is_set()` (channels.py:119) and the first `shutdown()` already called `self._cancel_sender.close()`

2. Even if that were caught, `self._task_sender.close()` (runner_supervisor.py:187) calls `buffer.put(_MpEndOfStream())` on an already-closed Queue → `ValueError`

The happy path (no timeout) is fine — `shutdown()` is called once in `finally`.

Fix options:
- **Add an idempotency guard**: `if self._shutdown: return` at the top of `shutdown()`
- **Remove from finally**, call explicitly in both the try success path and except path
- **Guard on process state**: `if supervisor.runner_process.is_alive(): supervisor.shutdown()`

### Minor

- **pyproject.toml change is unrelated** — `--ignore=tests/start_distributed_test.py` is the same change as PR #1478. Should be in a separate commit or dropped if #1478 merges first.

### What's good

- Correct diagnosis: `fail_after(3)` + `finally: shutdown()` was indeed leaving the process alive when `start_task()` timed out, because `shutdown()` was called *after* the timeout but `start_task()` never actually sent the shutdown command
- Rename from `runner` → `supervisor` improves clarity
- Calling `shutdown()` *before* emitting the `TimedOut` event is the right order
- Warning log with runner_id is useful for debugging

### Verdict

The fix correctly addresses orphaned runner processes. However, the double `shutdown()` call on the timeout path will raise `ClosedResourceError` or `ValueError` from already-closed IPC channels. Needs either an idempotency guard on `shutdown()` or restructured control flow to avoid the double call.

AlexCheema:
## Code Review: PR #1466 — fix: forcefully terminate runner on Shutdown timeout

### Summary
When a runner is blocked in `mx.eval()` (loading model weights) and can't respond to a `Shutdown` task within 3 seconds, the process is now forcefully terminated instead of being left as an orphan.

### Review

**The fix is correct:**
```python
case Shutdown(runner_id=runner_id):
    supervisor = self.runners.pop(runner_id)
    try:
        with fail_after(3):
            await supervisor.start_task(task)
    except TimeoutError:
        logger.warning(f"Runner {runner_id} did not respond to Shutdown within 3s, forcefully terminating")
        supervisor.shutdown()
        await self.event_sender.send(TaskStatusUpdated(..., task_status=TaskStatus.TimedOut))
    finally:
        supervisor.shutdown()
```

**Key observation:** `supervisor.shutdown()` is called in both the `except` and `finally` blocks. This means it's called **twice** on timeout. The shutdown method does: close channels → join(1) → SIGTERM → SIGKILL. Calling it twice should be idempotent (channels already closed, process already dead), but worth verifying the `shutdown()` method handles double-invocation gracefully.

### Issues

**1. `pyproject.toml` change is unrelated**
Same as PR #1478 — adds `--ignore=tests/start_distributed_test.py`. This should be in a separate PR.

**2. Double `shutdown()` on timeout path**
As noted above, `supervisor.shutdown()` is called in `except TimeoutError` AND in `finally`. The finally block always runs, so on the timeout path, shutdown is called twice. On the success path, shutdown is called once (only in finally). This works but is unintentional redundancy — consider removing the shutdown from the `except` block since `finally` handles it.

### Verdict
Good fix for a real resource leak. `mx.eval()` is a blocking C call that can't be interrupted from Python — killing the process is the correct approach. The double-shutdown is the only concern but is harmless.

LGTM.

rltakashige:
Closing as stale and unnecessary

Issue #1425, opened 2026-02-09
[BUG] error downloading shard on startup


bdols:
## Describe the bug

unable to get past GETTING-STARTED instructions

## To Reproduce

1. npm install && npm run build && uv run exo

## Expected behavior

able to download models, tried this on mac and linux, with the same results. I also have tailscale running and I stopped it locally.

## Actual behavior

A clear and concise description of what actually happened.

[ 09:20:38.2369PM | INFO    ] Unpausing API
[ 09:20:38.2380PM | INFO    ] Running on http://0.0.0.0:52415 (CTRL + C to quit)
[ 09:20:38.2384PM | INFO    ] Running on http://0.0.0.0:52415 (CTRL + C to quit)
[ 09:20:38.3469PM | WARNING ] Error downloading shard: FileNotFoundError
[ 09:20:38.3472PM | WARNING ] Error downloading shard: FileNotFoundError
[ 09:20:38.3515PM | WARNING ] Error downloading shard: FileNotFoundError
...
[ 09:20:38.4047PM | WARNING ] Error downloading shard: FileNotFoundError
[...]

PR #1477, opened 2026-02-15
Enhance internet connectivity checks and error handling in download

jayy-77:
## Motivation

[#1425](error downloading shard on startup)

### DIFF
--- a/src/exo/download/coordinator.py
+++ b/src/exo/download/coordinator.py
@@ -110,14 +110,36 @@ class DownloadCoordinator:
             tg.start_soon(self._check_internet_connection)
 
     def _test_internet_connection(self) -> None:
-        try:
-            socket.create_connection(("1.1.1.1", 443), timeout=3).close()
-            self.shard_downloader.set_internet_connection(True)
-        except OSError:
-            self.shard_downloader.set_internet_connection(False)
-        logger.debug(
-            f"Internet connectivity: {self.shard_downloader.internet_connection}"
-        )
+        # Try multiple hosts to be resilient against firewall/DNS issues
+        test_hosts = [
+            ("1.1.1.1", 443),  # Cloudflare DNS
+            ("8.8.8.8", 443),  # Google DNS
+            ("api.github.com", 443),  # GitHub API (likely needed for downloads)
+        ]
+        
+        connected = False
+        last_error = None
+        
+        for host, port in test_hosts:
+            try:
+                socket.create_connection((host, port), timeout=5).close()
+                connected = True
+                break
+            except OSError as e:
+                last_error = e
+                logger.debug(f"Connectivity check failed for {host}:{port}: {e}")
+                continue
+        
+        self.shard_downloader.set_internet_connection(connected)
+        
+        if connected:
+            logger.debug("Internet connectivity: ONLINE")
+        else:
+            logger.warning(
+                f"Internet connectivity: OFFLINE (all checks failed). "
+                f"Last error: {last_error}. "
+                f"Check firewall/proxy settings if downloads fail."
+            )
 
     async def _check_internet_connection(self) -> None:
         first_connection = True
--- a/src/exo/download/download_utils.py
+++ b/src/exo/download/download_utils.py
@@ -264,7 +264,9 @@ async def fetch_file_list_with_cache(
             )
             return local_file_list
         raise FileNotFoundError(
-            f"No internet connection and no cached file list for {model_id}"
+            f"Cannot download {model_id}: No internet connection detected and no cached files found. "
+            f"Please check your network connection, firewall settings, or proxy configuration. "
+            f"Ensure you can reach huggingface.co and api.github.com."
         )
 
     try:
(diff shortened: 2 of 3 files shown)


AlexCheema:
## Code Review: Enhance internet connectivity checks and error handling in download

Addresses #1425 where a false-negative connectivity check (1.1.1.1:443 blocked by VPN/firewall) causes all downloads to fail.

### Issues

**1. Event loop blocking regression (must fix)**

`_test_internet_connection()` is a **synchronous** method doing blocking socket I/O, called from async contexts (`run()` at startup, `_check_internet_connection()` every 10s when offline).

- **Before**: 1 host × 3s timeout = max **3s** blocking
- **After**: 3 hosts × 5s timeout = max **15s** blocking (5x worse)

At startup, the entire node freezes for up to 15s. During periodic rechecks when offline, the event loop blocks for 15s every 10s (blocking ~60% of the time).

**Fix**: Run `_test_internet_connection` off the event loop via `anyio.to_thread.run_sync()` or `asyncio.get_event_loop().run_in_executor()`.

**2. DNS resolution for `api.github.com` is not bounded by socket timeout (must fix)**

`socket.create_connection()`'s `timeout` parameter only affects the TCP `connect()` call, **not** DNS resolution. `getaddrinfo()` is synchronous and unaffected by the timeout. If DNS is broken but raw IP connectivity works:

- Host 1 (1.1.1.1): 5s TCP timeout → fail
- Host 2 (8.8.8.8): 5s TCP timeout → fail
- Host 3 (api.github.com): **30-60s DNS timeout** → fail
- **Total: 40-70 seconds** of event loop blocking at startup

**Fix**: Use IP addresses only for connectivity checks, or use async DNS resolution.

**3. Code duplication (should fix)**

The try/except/retry pattern is duplicated verbatim between `ensure_shard()` and `get_shard_download_status_for_shard()` (~20 lines each). Should extract to a helper method.

### What's Good

**4. Retry logic is the correct fix for #1425**

When `skip_internet=True` fails with `FileNotFoundError` (no cache), retrying with `skip_internet=False` is the right approach. This handles the case where the connectivity check returns a false negative but HuggingFace is actually reachable (e.g., traffic routed through VPN). The user gets either a successful download or an actionable network error instead of a confusing "no cache" message.

**5. Improved error messages are helpful**

The new messages are actionable: "Check your network connection, firewall settings, or proxy configuration. Ensure you can reach huggingface.co."

### Minor

- **Timeout 3→5s per host**: 3s is already generous for a TCP handshake. Combined with 3 hosts, 5s makes the worst case much worse.
- **`logger.warning` on every recheck**: Fires every 10-25s when offline, flooding logs. Consider rate-limiting or using `debug` after the first warning.
- **PR description is empty template**: No description of changes, test plan, or verification.

### Verdict

The retry logic correctly addresses #1425, and the improved error messages add real value. However, the multi-host connectivity check introduces a significant event loop blocking regression (up to 15-70s depending on DNS) that must be fixed before merging. The existing 3s blocking was already marginal — making it 5-20x worse is not acceptable.

**Required**: Run connectivity checks off the event loop, use IP addresses only (or async DNS), reduce per-host timeout back to 3s.

---

AlexCheema:
## Code Review: PR #1477 — Enhance internet connectivity checks and error handling in download

### Summary
Expands the internet connectivity check to try multiple hosts, improves error messages for download failures, and adds retry logic when the internet connection status is wrong.

### Review

**Overlaps with PR #1516:**
This PR and #1516 both modify `_test_internet_connection()` to try multiple endpoints. Differences:
- #1516: tries `1.1.1.1`, `8.8.8.8`, `1.0.0.1` (all DNS servers, port 443)
- #1477: tries `1.1.1.1`, `8.8.8.8`, `api.github.com` (adds a hostname, uses 5s timeout)

#1477 is more comprehensive (includes error messages, retry logic, download improvements) but also larger scope.

**Connectivity check improvements:**
- Adds `api.github.com` as a test host — good because it tests DNS resolution too ✅
- Increased timeout from 3s to 5s per host ✅
- Better logging (shows which host succeeded/failed) ✅

**Error message improvements (`download_utils.py`):**
- More actionable error messages mentioning firewall/proxy ✅
- Falls back to local files when network fails ✅

**Retry logic in `ensure_shard` (`impl_shard_downloader.py`):**
When download fails with `skip_internet=True`, retries with `skip_internet=False` to get the actual network error. This is clever — it means the user sees a real network error instead of "no cached files found". ✅

### Issues

**1. Retry can trigger an unnecessary network request**
The retry in `ensure_shard` always retries on `FileNotFoundError` when `not self.internet_connection`. But the error might genuinely be "model doesn't exist on HuggingFace" — not a connectivity issue. The retry would waste time and bandwidth.

**2. Overlaps significantly with PR #1516**
Both PRs modify the same function. Only one should be merged, and the other needs rebasing. This one (#1477) is more comprehensive.

### Verdict
Good improvements to error handling and diagnostics. The retry-for-real-error pattern is clever. Main concern is the overlap with #1516 and the potential for unnecessary retries.

LGTM with the caveats above.

PR #1481, opened 2026-02-16
yay

### LINKED ISSUE #1473: [BUG] `/place_instance` API silently ignores `sharding` and `instance_meta` parameters — always falls back to Pipeline/MlxRing
### Describe the bug

The `PlaceInstanceParams` Pydantic validator in `src/exo/shared/types/api.py` uses `mode="plain"` which disables automatic string-to-enum conversion. When JSON strings like `"Tensor"` or `"MlxJaccl"` are sent via the API, the validator checks `isinstance(v, (Sharding, InstanceMeta))` which always fails for raw strings, causing `PydanticUseDefault()` — silently falling back to `Pipeline`/`MlxRing` defaults.

This means it is **impossible** to request tensor parallelism via the `/place_instance` API.

### To Reproduce

```bash
# This should place the model with Tensor sharding + MlxJaccl backend:
curl -X POST http://localhost:52415/place_instance \
  -d '{"model_id": "mlx-community/DeepSeek-V3.1-8bit", "sharding": "Tensor", "instance_meta": "MlxJaccl"}'

# But it always creates a Pipeline + MlxRing instance instead.
# Check via: curl http://localhost:52415/state
# → MlxRingInstance with PipelineShardMetadata, not MlxJacclInstance with TensorShardMetadata
```

### Expected behavior

The API should accept `"Tensor"` and `"MlxJaccl"` as string values and convert them to the corresponding `Sharding.Tensor` and `InstanceMeta.MlxJaccl` enums.

### Root cause

In `src/exo/shared/types/api.py`:

```python
class PlaceInstanceParams(BaseModel):
    model_id: ModelId
    sharding: Sharding = Sharding.Pipeline
    instance_meta: InstanceMeta = InstanceMeta.MlxRing
    min_nodes: int = 1

    @field_validator("sharding", "instance_meta", mode="plain")
    @classmethod
    def use_default(cls, v: object):
        if not v or not isinstance(v, (Sharding, InstanceMeta)):
            raise PydanticUseDefault()  # ← string "Tensor" is NOT isinstance Sharding!
        return v
```

The `mode="plain"` disables Pydantic's default str→Enum coercion. The validator then rejects all string values.

### Suggested fix

```python
@field_validator("sharding", "instance_meta", mode="plain")
@classmethod
def use_default(cls, v: object, info):
    if not v:
        raise PydanticUseDefault()
    if isinstance(v, (Sharding, InstanceMeta)):
        return v
    if isinstance(v, str):
        enum_cls = Sharding if info.field_name == "sharding" else InstanceMeta
        try:
            return enum_cls(v)
        except ValueError:
            pass
    raise PydanticUseDefault()
```

### Impact

This affects **all models**, not just GLM-5. Any user trying to request Tensor sharding or MlxJaccl backend via the REST API will silently get Pipeline/MlxRing instead. The dashboard's placement preview (`/instance/previews`) correctly enumerates all combinations, but the actual placement API ignores the user's choice.

### Environment

- Exo: built from source (git main, Feb 2026)
- macOS 26.2
- Python 3.13

Evanev7:
closes #1473

our field validator for PlaceInstance was wrong - we can just rely on default behaviour here anyway!

### DIFF
--- a/src/exo/shared/types/api.py
+++ b/src/exo/shared/types/api.py
@@ -3,8 +3,7 @@ from collections.abc import Generator
 from typing import Annotated, Any, Literal
 from uuid import uuid4
 
-from pydantic import BaseModel, Field, field_validator
-from pydantic_core import PydanticUseDefault
+from pydantic import BaseModel, Field
 
 from exo.shared.models.model_cards import ModelCard, ModelId
 from exo.shared.types.common import CommandId, NodeId
@@ -227,13 +226,6 @@ class PlaceInstanceParams(BaseModel):
     sharding: Sharding = Sharding.Pipeline
     instance_meta: InstanceMeta = InstanceMeta.MlxRing
     min_nodes: int = 1
-
-    @field_validator("sharding", "instance_meta", mode="plain")
-    @classmethod
-    def use_default(cls, v: object):
-        if not v or not isinstance(v, (Sharding, InstanceMeta)):
-            raise PydanticUseDefault()
-        return v
 
 
 class CreateInstanceParams(BaseModel):


AlexCheema (approved):
nice fix

PR #1484, opened 2026-02-16
fix: RDMA health detection — only warn for directly-connected Thunderbolt pairs

AlexCheema:
## Summary

- Adds proactive RDMA device health detection (`ibv_alloc_pd` polling) with dashboard warning banner
- Fixes the RDMA UNHEALTHY warning to only show "disconnect and reconnect Thunderbolt 5 cable" for node pairs that **actually have a direct Thunderbolt 5 connection** (RDMA edge in topology), not all connected pairs
- Previously, the warning iterated over ALL topology edges (Socket + RDMA), showing cable warnings for every pair (1-2, 1-3, 1-4, 2-3, 2-4, 3-4) as if they were all-to-all connected via Thunderbolt — they are not

## Test plan

- [ ] Verify RDMA UNHEALTHY warning only appears for node pairs with `sourceRdmaIface` edges in topology
- [ ] Verify socket-only pairs do not get cable disconnect warnings
- [ ] Dashboard builds successfully (`cd dashboard && npm run build`)

### DIFF
--- a/src/exo/master/reconcile.py
+++ b/src/exo/master/reconcile.py
@@ -241,7 +241,9 @@ def try_place_for_meta_instance(
                 update={"meta_instance_id": meta_instance.meta_instance_id}
             )
         return PlacementResult(
-            events=list(get_transition_events(current_instances, target_instances, tasks or {})),
+            events=list(
+                get_transition_events(current_instances, target_instances, tasks or {})
+            ),
             error=None,
         )
     except ValueError as e:


AlexCheema:

Large PR (+3860) with two distinct parts:
1. **RDMA health detection**: Adds proactive `ibv_alloc_pd` polling to detect unhealthy RDMA devices, surfaced as a dashboard warning banner
2. **Fix RDMA cable warning scope**: The UNHEALTHY warning now only shows "disconnect and reconnect Thunderbolt 5 cable" for node pairs that actually have a direct Thunderbolt connection (RDMA edge in topology), not for all connected pairs — previously it was showing cable warnings for socket-only pairs too

Also includes MetaInstance declarative layer for ensuring model instances stay running, connection health checking that verifies topology edges still exist, and master event flow simplification.

Note: This is a large change touching both Rust and Python. The RDMA warning fix is clearly correct and well-scoped; the MetaInstance additions are substantial new functionality bundled in.

AlexCheema (commented):
## Code Review: PR #1484 — RDMA Health Detection

### Overall Assessment
Solid PR that correctly narrows RDMA warnings to Thunderbolt 5-connected pairs. The dashboard filtering logic is correct and well-implemented. The removal of `MlxIbvInstance` is clean throughout.

### Strengths

1. **Dashboard RDMA edge filtering is correct** (`+page.svelte` ~lines 306-338): Filtering `data.edges` by `e.sourceRdmaIface` correctly identifies actual TB5 connections vs socket-only pairs. The `seenPairs` deduplication with sorted pair keys prevents duplicate warnings.

2. **Clean MlxIbv removal**: All references to `MlxIbvInstance` and `MlxIbv` are consistently removed across dashboard components (`ChatSidebar.svelte`, `ModelCard.svelte`, `app.svelte.ts`, `+page.svelte`), type definitions, and the runtime selector. No stale references found.

3. **`mp.set_start_method("spawn", force=True)`** in `main.py`: Good fix — prevents `RuntimeError` when multiprocessing context is already set (e.g. in test environments or when EXO.app bundles set it).

4. **MetaInstance + ProcessManager architecture** (`process_managers/__init__.py`): The `ProcessManager` Protocol is clean and composable. `_apply_and_broadcast` eliminates the hacky loopback processor.

### Issues & Questions

5. **Scope is much larger than the title suggests**: The title says "RDMA health detection — only warn for directly-connected Thunderbolt pairs" but the PR includes the entire MetaInstance declarative layer, placement changes (smallest→largest cycles), process manager refactoring, JACCL side channel handling, and instance retry logic (+3860/-221 lines). Consider whether this should be split or if the title/description should reflect the full scope.

6. **Placement behavior change** (`placement.py`): `get_smallest_cycles` → `get_largest_cycles` changes the default placement strategy from "use fewest nodes" to "use most nodes". This is a significant behavior change that could affect performance characteristics for users. Worth documenting in the PR description.

7. **Reconcile loop frequency** (`master/main.py`): Changed from `await anyio.sleep(10)` to `await anyio.sleep(1)`. The `MetaInstanceReconciler` calls `ModelCard.load()` for every unsatisfied meta-instance on each cycle. If this involves network I/O (HuggingFace API), the 1-second loop could generate excessive requests when placement persistently fails.

8. **Dashboard dismissed state is ephemeral** (`+page.svelte`): `rdmaHealthDismissed` is a `$state(false)` — dismissing the RDMA warning doesn't persist across page refreshes. For a warning about a hardware issue that persists until cable reconnection, users may want it to stay dismissed. Consider using `localStorage`.

9. **MlxJaccl as sole RDMA path** (`placement.py` ~line 117-120): The new code raises `ValueError` if `MlxJaccl` is requested but no RDMA cycles exist. The old code silently fell back. This is correct for explicit RDMA requests but verify the dashboard correctly handles this error path.

### Minor

10. **`nodeRdmaDeviceHealth` not populated in cluster test**: From testing on the Mac Minis (earlier in this session), `nodeRdmaDeviceHealth` was never populated in the state response. The info_gatherer RDMA probing may need the RDMA devices to actually exist on the system — verify this works on TB5-equipped machines.

### Verdict
The code is correct for the RDMA warning filtering use case. The broader MetaInstance/reconciliation changes are architecturally sound but deserve their own review focus given the scope. **Approve** for the RDMA detection fix; the MetaInstance layer is solid but large.

AlexCheema:
## Code Review — PR #1484: RDMA health detection & MetaInstance declarative layer

**CI: ALL PASSING** (aarch64-darwin, x86_64-linux, aarch64-linux)

### Overview

Large PR — 31 files, +3860/-221 (~1750 lines of tests). Bundles several features:
- **RDMA device health probing** via `ibv_alloc_pd` ctypes (info_gatherer.py)
- **RDMA warning scoping** — only warn for node pairs with actual Thunderbolt 5 edges (`sourceRdmaIface`), not all topology edges
- **MetaInstance declarative layer** — server-side placement + auto-retry + re-placement
- **ProcessManager pattern** — replaces ad-hoc `_plan` loop with composable reconcilers (InstanceHealth, NodeTimeout, MetaInstance)
- **Loopback elimination** — removes `_loopback_processor` channel, `_apply_and_broadcast` replaces scattered event indexing
- **MlxIbvInstance removal** — consolidates RDMA to MlxJaccl only
- **JACCL SideChannel** — FIFO-based pipe relay for tensor-parallel all_gather
- **Runner cleanup** — apply_instance_deleted and apply_node_timed_out now remove stale runners from state

### Should fix

**1. localStorage backward compatibility for MlxIbv**

`applyLaunchDefaults()` (line 531) loads `selectedInstanceType = defaults.instanceType` without validation. Users upgrading from the old dashboard may have `"MlxIbv"` stored in localStorage. This sets `selectedInstanceType` to an invalid value — the RDMA radio appears unselected, launches could silently use the wrong type.

Fix: validate before applying:
```js
if (defaults.instanceType !== "MlxRing" && defaults.instanceType !== "MlxJaccl")
  defaults.instanceType = "MlxRing";
```

**2. Test bug: `test_meta_instance_model_is_frozen` always passes**

```python
try:
    meta.model_id = ModelId("something-else")
    raise AssertionError("Should have raised")
except Exception:
    pass  # Expected — frozen model
```

`AssertionError` is a subclass of `Exception`, so the `except` catches its own failure sentinel. This test passes regardless of whether the frozen check works. Use `pytest.raises(ValidationError)` instead.

**3. JACCL SideChannel events pollute the disk event log**

`JacclSideChannelData` and `JacclSideChannelGathered` carry `Base64Bytes` binary payloads and are persisted to disk via `_apply_and_broadcast`. During active inference, these fire continuously. They're correctly marked as pass-through in `apply.py` (no state mutation), but still get appended to the disk event log and broadcast.

Consider: skip disk persistence for these pass-through events, or use a separate channel.

### Non-blocking concerns

**4. FIFO open has no timeout** — If the child process crashes before opening its FIFO end, `os.open(fifo_c2p, O_RDONLY)` blocks forever (line ~5019 of diff). Add `move_on_after()` around the FIFO open task group.

**5. MetaInstance.node_ids uses mutable `list` inside frozen model** — `meta_instance.node_ids.append(...)` silently succeeds. Use `tuple[NodeId, ...] | None` for real immutability.

**6. `min_nodes` has no `ge=1` constraint** — `min_nodes: int = 1` accepts 0 or negative. Fix: `min_nodes: int = Field(default=1, ge=1)`.

**7. `_jaccl_pending` not cleaned on instance deletion** — If an instance is deleted mid-JACCL gather, the pending entry stays forever. Minor memory leak.

**8. Reconcile loop 10x more frequent (1s vs old 10s)** — Each iteration may call `ModelCard.load()` for unsatisfied meta-instances. Consider caching model cards or using backoff on repeated placement failures.

**9. `rdmaHealthDismissed` is session-permanent** — Once dismissed, warning never returns even if new unhealthy pairs appear. The `jacclDismissedError` uses a smarter string-comparison pattern. Consider the same for RDMA health.

### Architecture positives

- MetaInstance abstraction is well-designed — declarative constraints with auto-retry and re-placement
- ProcessManager protocol is clean and composable — each reconciler is independently testable
- Loopback elimination simplifies Master event flow significantly
- RDMA edge filtering (the core fix) is correct — `sourceRdmaIface` check targets actual TB5 connections
- `ibv_alloc_pd` probing: correct ctypes usage with proper cleanup, threaded execution, 15s timeout
- Runner cleanup on deletion/timeout fixes real stale-state bugs
- Event sourcing purity maintained — all apply functions remain pure
- Dashboard MetaInstance UI: launch flow moved from imperative to declarative — architecturally sound
- Duplicate node detection via `friendly_name` in NodeTimeoutReconciler is smart

### Test coverage

1754 lines across 4 test files. Quality is excellent. Gaps:
- `_jaccl_connections_healthy` has zero test coverage (only MlxRing tested)
- MetaInstanceReconciler primary workflow (unsatisfied → placed) untested
- NodeTimeoutReconciler basic timeout path untested (only duplicate detection)
- No test verifies `meta_instance_id` is tagged on newly created instances

### Verdict

Approve with minor fixes: (1) localStorage MlxIbv migration guard, (2) frozen model test bug. The MetaInstance pattern and ProcessManager architecture are well-designed. RDMA health probing and edge filtering are correct and well-implemented.

AlexCheema:
# Code Review: PR #1484 -- RDMA health detection + MetaInstance declarative layer

**PR**: exo-explore/exo#1484
**Author**: AlexCheema
**Branch**: alexcheema/rdma-health-detection -> main
**Lines**: +3866 / -221
**CI Status**: All checks passing (aarch64-darwin, aarch64-linux, x86_64-linux)

---

## 1. Executive Summary

Despite the title "fix: RDMA health detection -- only warn for directly-connected Thunderbolt pairs", this PR is a major architectural overhaul. It introduces:

1. **MetaInstance declarative layer** -- a new abstraction for persistent model deployment intent
2. **ProcessManager reconciliation protocol** -- replaces ad-hoc `_plan()` with composable reconcilers
3. **JACCL SideChannel pipe relay** -- named-pipe-based IPC replacing TCP SideChannel
4. **Instance retry logic** -- automatic runner restart with exponential failure tracking
5. **Node duplicate detection** -- evicts stale peer IDs on quick restarts
6. **RDMA device health probing** -- ctypes-based ibv_alloc_pd polling
7. **Dashboard overhaul** -- MetaInstance-centric UI with warning banners
8. **MlxIbv removal** -- clean elimination of the deprecated instance type

The scope is **dramatically larger** than the title suggests. 44 commits spanning ~3866 additions across 31 files is not a "fix" -- it is a feature branch containing multiple independent concerns. This makes it harder to review, harder to bisect, and harder to revert individual changes.

---

## 2. Architecture and Design Decisions

### 2.1 MetaInstance Abstraction (Positive)

The MetaInstance is a well-conceived declarative layer. Instead of imperatively creating instances, the user declares "I want a model running with these constraints" and the system continuously reconciles toward that state. This is a proven pattern (Kubernetes Deployments, etc.).

**Strengths:**
- Clean separation between intent (`MetaInstance`) and realization (`Instance`)
- `meta_instance_id` embedded directly on `BaseInstance` avoids a separate binding map
- Failure tracking (consecutive_failures, last_failure_error) lives on MetaInstance itself
- The reconciler is idempotent against state, so the 1s tick is safe

**File**: `src/exo/shared/types/meta_instance.py`

### 2.2 ProcessManager Protocol (Positive)

```python
@runtime_checkable
class ProcessManager(Protocol):
    async def reconcile(self, state: State) -> Sequence[Event]: ...
```

This is a clean, composable abstraction. Each reconciler (InstanceHealthReconciler, NodeTimeoutReconciler, MetaInstanceReconciler) is independently testable. The elimination of the `_loopback_processor` and unification through `_apply_and_broadcast` significantly simplifies the master's control flow.

**File**: `src/exo/master/process_managers/__init__.py`

### 2.3 _apply_and_broadcast Unification (Positive)

The new `_apply_and_broadcast` method applies events synchronously to state before awaiting the broadcast send. The docstring correctly identifies that Python's cooperative scheduling prevents interleaving between state read and write. This eliminates a class of bugs where the old loopback path could see stale state.

**File**: `src/exo/master/main.py`

### 2.4 JACCL SideChannel Pipe Relay (Concern)

The switch from TCP SideChannel to named pipes (FIFOs) relayed through exo's event-sourced control plane is architecturally significant. The motivation (errno=57 from Thunderbolt RDMA instability) is valid, but routing potentially large binary payloads through the gossipsub event bus has implications:

- **Bandwidth**: Every all_gather round sends data through the master as an event. For large tensors, this could saturate the pub/sub channel.
- **Latency**: Adding master relay adds at least one network round-trip per all_gather.
- **Base64 encoding overhead**: The `Base64Bytes` type adds ~33% overhead to the wire format.
- **Sequence tracking**: `_jaccl_pending` is an in-memory dict with no eviction. If a runner dies mid-gather, entries leak.

**Files**: `src/exo/worker/runner/runner_supervisor.py`, `src/exo/shared/types/events.py`

### 2.5 Reconcile Loop Frequency Change (Concern)

The reconcile interval changed from 10s to 1s. The `MetaInstanceReconciler.reconcile()` calls `ModelCard.load()` for every unsatisfied meta-instance on each tick. If `ModelCard.load()` involves network I/O (e.g., HuggingFace hub), this could generate excessive requests when placement persistently fails. Consider:
- Caching `ModelCard.load()` results
- Adding backoff for unsatisfied meta-instances that have placement errors
- Using exponential backoff per-meta-instance

**File**: `src/exo/master/process_managers/meta_instance.py`

---

## 3. Error Handling

### 3.1 CreateMetaInstance Command Race Handling (Good)

The `CreateMetaInstance` handler eagerly applies `MetaInstanceCreated`, then awaits `ModelCard.load()`, then re-checks `find_unsatisfied_meta_instances` before attempting placement. This correctly handles the race where the reconciler could place during the await.

**File**: `src/exo/master/main.py` lines ~305-340

### 3.2 DeleteMetaInstance Cascade (Good)

The cascade delete is handled atomically in the command processor, generating `MetaInstanceDeleted` + `InstanceDeleted` events in a single batch. This fixes the race described in the commit message where the API handler's stale state read could miss backing instances.

### 3.3 Stale JACCL Pending Entries (Bug Risk)

In `_handle_jaccl_side_channel`, when an instance is deleted while a gather is in progress, the entries in `self._jaccl_pending[iid]` are never cleaned up. The check `instance = self.state.instances.get(iid)` returns None and logs a warning, but the pending data stays. Over time (with frequent instance churn), this could leak memory.

**Recommendation**: Clean up `_jaccl_pending` entries when `InstanceDeleted` is processed.

**File**: `src/exo/master/main.py` lines ~493-530

### 3.4 FIFO Cleanup on Error Paths (Mostly Good)

The `_close_pipe_fds` method properly cleans up FIFOs, file descriptors, and temp directories with `contextlib.suppress(OSError)`. However:
- If the child process crashes before opening the FIFO, the parent's `os.open(fifo_c2p, os.O_RDONLY)` in `open_read()` will block indefinitely in a background thread. There is no timeout on the FIFO open operation. This could cause a resource leak if the runner process dies during startup.

**File**: `src/exo/worker/runner/runner_supervisor.py`

### 3.5 `instance_runners_failed` Early Return on Missing Status (Correct)

The function returns `(False, None)` when a runner hasn't reported yet (status is None). This prevents false positives during instance startup. The logic correctly requires ALL runners to be terminal with at least one `RunnerFailed`.

**File**: `src/exo/master/reconcile.py`

---

## 4. Type Safety

### 4.1 MetaInstance is Frozen (Good)

`MetaInstance` extends `FrozenModel` and is `@final`. Mutation is only through `model_copy(update=...)` in the apply handlers.

### 4.2 Base64Bytes Custom Type (Good, but Complex)

The `Base64Bytes` annotated type with `BeforeValidator`/`PlainSerializer` correctly handles JSON round-tripping of bytes through TaggedModel's wrap validator. The comment explains the "why" well.

### 4.3 Dashboard Type Casts (Concern)

Several places in the dashboard use unsafe casts:
```typescript
const inst = instance as {
    shardAssignments?: {
        runnerToShard?: Record<string, unknown>;
        nodeToRunner?: Record<string, string>;
    };
};
```
and:
```typescript
const rd = runnerData as { errorMessage?: string };
```

These bypass TypeScript's type checking. While pragmatic for working with the dynamically-typed state response, they could hide bugs. Consider defining a proper `Instance` interface for the dashboard.

### 4.4 MetaInstanceId Type (Good)

`MetaInstanceId` is a proper `NewType`-style class extending `Id`, following the project's convention for type-safe identifiers.

### 4.5 Missing `@final` on Some Event Classes

`MetaInstanceCreated` and `MetaInstanceDeleted` lack the `@final` decorator, while `MetaInstancePlacementFailed` and `InstanceRetrying` have it. This inconsistency should be addressed -- all leaf event types should be `@final` per the project's code style.

**File**: `src/exo/shared/types/events.py`

### 4.6 MetaInstance.node_ids Uses Mutable list Inside Frozen Model

`MetaInstance` is a `FrozenModel`, but `node_ids: list[NodeId] | None = None` uses a mutable `list`. This means `meta_instance.node_ids.append(...)` would silently succeed, violating the immutability guarantee. Use `tuple[NodeId, ...] | None` for real immutability. (The commit history mentions a change from `frozenset` to `list` for JSON serialization, but `tuple` would also work.)

**File**: `src/exo/shared/types/meta_instance.py`

### 4.7 min_nodes Has No Lower Bound Constraint

`min_nodes: int = 1` accepts 0 or negative values. Add `min_nodes: int = Field(default=1, ge=1)` to prevent invalid states.

**File**: `src/exo/shared/types/meta_instance.py`

---

## 5. Edge Cases and Potential Bugs

### 5.1 Node Duplicate Detection False Positive Risk

`_find_stale_duplicate_nodes` groups by `friendly_name` and evicts all but the most recent `last_seen`. If two legitimately different machines happen to have the same hostname (e.g., both named "Mac mini"), the older one would be force-evicted. The "Unknown" name is skipped, which is good, but common default names like "Mac mini" could match unintentionally.

**Severity**: Medium -- could cause production issues in clusters with identically-named machines.
**Mitigation**: Consider using a more robust identity signal (e.g., hardware UUID from `SystemProfiler`) in addition to friendly_name.

**File**: `src/exo/master/process_managers/node_timeout.py`

### 5.2 Dead Node Constraint Relaxation

`try_place_for_meta_instance` intersects `node_ids` with live topology nodes and drops dead ones. If ALL pinned nodes are dead, it drops the constraint entirely:
```python
if alive:
    required_nodes = alive
# else: required_nodes stays None -> no constraint
```
This means a MetaInstance pinned to nodes A,B,C could be placed on completely different nodes D,E,F if A,B,C all die. This might be surprising behavior. Consider whether falling back to "any available nodes" is always desirable vs. keeping the MetaInstance in a "placement failed" state.

**File**: `src/exo/master/reconcile.py`

### 5.3 Retry Counter Reset Timing

`apply_instance_created` resets `consecutive_failures` to 0 but preserves `last_failure_error`. The comment says this is "for UI display". However, this means a successfully created instance that later fails will show the PREVIOUS failure's error message during the "PLACING" state after the first failure. This could be confusing.

### 5.4 _create_runner Terminal Peer Guard

The worker's `_create_runner` now checks for terminal peer runners:
```python
has_terminal_peer = any(
    isinstance(all_runners.get(peer_rid), (RunnerFailed, RunnerShutdown))
    for peer_rid in instance.shard_assignments.node_to_runner.values()
    if peer_rid != runner_id
)
if has_terminal_peer:
    continue
```
This prevents oscillation but introduces a dependency on the master emitting `InstanceRetrying` promptly. If the master's reconcile loop is delayed (e.g., under load), runner creation could be blocked for up to 1 second. This seems acceptable for the 1s reconcile interval.

### 5.5 Placement Error Spam

`MetaInstanceReconciler` emits `MetaInstancePlacementFailed` only when `result.error != meta_instance.placement_error`. However, if the error message includes timestamps or dynamic data, it would emit a new event every tick (1s). The current implementation seems safe since error messages are static strings from `ValueError`.

### 5.6 getBackingInstanceId Linear Scan

`getBackingInstanceId()` in the dashboard does a linear scan of all instances to find the backing one for a MetaInstance. This is called inside `unifiedDisplayItems` which is a derived value, so it runs on every state update. With many instances, this could become slow. Consider building a reverse map.

**File**: `dashboard/src/routes/+page.svelte`

### 5.7 localStorage Backward Compatibility for MlxIbv

`applyLaunchDefaults()` loads `selectedInstanceType = defaults.instanceType` without validation. Users upgrading from the old dashboard may have `"MlxIbv"` stored in localStorage. This sets `selectedInstanceType` to an invalid value -- the RDMA radio appears unselected, and launches could silently use the wrong type.

**Recommendation**: Validate before applying:
```js
if (defaults.instanceType !== "MlxRing" && defaults.instanceType !== "MlxJaccl")
  defaults.instanceType = "MlxRing";
```

### 5.8 Test Bug: test_meta_instance_model_is_frozen Always Passes

```python
try:
    meta.model_id = ModelId("something-else")
    raise AssertionError("Should have raised")
except Exception:
    pass  # Expected -- frozen model
```

`AssertionError` is a subclass of `Exception`, so the `except` catches its own failure sentinel. This test passes regardless of whether the frozen check works. Use `pytest.raises(ValidationError)` or catch a more specific exception.

### 5.9 JACCL SideChannel Events Pollute Disk Event Log

`JacclSideChannelData` and `JacclSideChannelGathered` carry `Base64Bytes` binary payloads and are persisted to disk via `_apply_and_broadcast`. During active inference, these fire continuously. They are correctly marked as pass-through in `apply.py` (no state mutation), but still get appended to the disk event log and broadcast to all workers.

**Recommendation**: Skip disk persistence for pass-through events, or use a separate channel for JACCL data.

### 5.10 rdmaHealthDismissed Is Session-Permanent

Once dismissed, the RDMA health warning never returns even if new unhealthy pairs appear. The `jacclDismissedError` uses a smarter string-comparison pattern (re-appears when a new error arrives). Consider the same approach for RDMA health.

---

## 6. Test Coverage

### 6.1 Strengths

The test coverage is genuinely comprehensive (~1754 lines of tests across 4 new files):

- **`test_reconcile.py`** (856 lines): Covers `instance_satisfies_meta_instance`, `instance_connections_healthy` (single/multi node, edge removal, IP change), `find_unsatisfied_meta_instances` (satisfied/unsatisfied/mix/edge break/idempotent/exclusive binding), all apply handlers, and the `InstanceHealthReconciler` (retry/delete/no-meta/network failure).

- **`test_meta_instance_edge_cases.py`** (585 lines): Covers frozen model validation, create/delete roundtrip, nonexistent ID safety, multiple MetaInstances for same model, retry counter reset, full retry cycle through MAX_INSTANCE_RETRIES, exact limit boundary, error message collection, friendly name inclusion, backward compatibility (no meta_instance_id), and serialization roundtrip.

- **`test_apply_instance_deleted.py`** (142 lines): Covers runner cleanup on instance deletion, cross-instance isolation, multi-node cleanup, and unknown instance safety.

- **`test_apply_node_timed_out.py`** (171 lines): Covers runner cleanup on node timeout, cross-node isolation, multi-instance cleanup, and the stale runner scenario.

### 6.2 Gaps

- **No tests for JACCL SideChannel relay**: The pipe relay logic (`_pipe_relay`, `notify_gathered`, `_handle_jaccl_side_channel`) has zero test coverage. This is complex async code with manual binary framing (struct pack/unpack) and cross-process coordination.

- **No tests for `_probe_rdma_pd_health`**: The ctypes-based RDMA probing is untested. While it is platform-specific (macOS only), mocking the ctypes calls would catch regressions in the probing logic.

- **No tests for `MetaInstanceReconciler.reconcile`**: While `find_unsatisfied_meta_instances` and `try_place_for_meta_instance` are individually tested, the reconciler's integration (local instance tracking to prevent double-placement, error emission deduplication) is not.

- **No tests for `NodeTimeoutReconciler` beyond duplicate detection**: The basic timeout path (node inactive > 30s) is not directly tested in the new reconciler form.

- **`_jaccl_connections_healthy` has zero test coverage** -- only MlxRing health checking is tested.

- **No test verifies `meta_instance_id` is tagged on newly created instances** from the placement flow.

- **No dashboard tests**: The RDMA warning filtering logic, MetaInstance display, and retry status derivation are complex enough to warrant unit tests (e.g., with Vitest).

---

## 7. Code Style Consistency

### 7.1 Consistent with Project Conventions

- Pydantic models use `FrozenModel` / `CamelCaseModel` / `TaggedModel` as appropriate
- Pure functions for state transitions in `apply.py`
- `@final` on leaf types (mostly -- see 4.5)
- Descriptive names, no abbreviations
- Type annotations throughout

### 7.2 Test Helper Duplication

`_model_card()`, `_topology()`, `_meta_instance()`, and `_instance()` helper functions are duplicated between `test_reconcile.py` and `test_meta_instance_edge_cases.py`. The latter file even has a comment: "Helpers (copied from test_reconcile.py for independence)". Consider extracting these into a shared test fixture module to reduce maintenance burden.

### 7.3 Dashboard Code Duplication

The instance display logic (both in the main grid and the topology view) is heavily duplicated. The `{#each unifiedDisplayItems as item}` blocks with their `@const` chains appear twice with nearly identical code. This is fragile -- a bug fix in one location may not be applied to the other.

### 7.4 Commit History

44 commits with a mixture of features, fixes, and reverts (e.g., "Remove permanent retry blocking, allow continuous retry batches" reverting an earlier commit's approach). The commit messages individually are well-written, but the history reads like a development journal rather than a clean, logical progression. A squash or interactive rebase to ~5-10 logical commits would make the history more useful for future debugging.

---

## 8. Specific Issues by File

### 8.1 `src/exo/master/main.py`

- **Line ~311**: `MetaInstanceCreated` is applied eagerly, then `ModelCard.load()` is awaited, then satisfaction is rechecked. If `ModelCard.load()` raises, the MetaInstance exists in state with no placement attempt. The reconciler will eventually pick it up, so this is self-healing, but the error is silently swallowed by the outer `except ValueError`.

### 8.2 `src/exo/master/reconcile.py`

- **`instance_connections_healthy` match exhaustiveness**: The match statement has `MlxRingInstance` and `MlxJacclInstance` cases but no default/wildcard. If a new instance type is added, this function will implicitly return `None` (falsy), causing all instances of the new type to be considered unhealthy. Adding an explicit `case _: return True` or raising would be safer.

### 8.3 `src/exo/worker/runner/runner_supervisor.py`

- **`_pipe_relay` asserts**: The function starts with `assert self._pipe_read_fd is not None`. In production, asserts can be disabled with `-O`. Consider using `if ... is None: raise RuntimeError(...)` instead.

- **`_close_pipe_fds` closes `_child_pipe_fds`**: This field is assigned `None` at init and never set to anything else in this diff. The cleanup code for it appears to be dead code.

### 8.4 `src/exo/utils/info_gatherer/info_gatherer.py`

- **`_probe_rdma_pd_health` runs synchronously**: It is correctly dispatched via `to_thread.run_sync`. However, if `ibv_alloc_pd()` hangs (which can happen with broken RDMA drivers), the thread will be blocked. The `fail_after(15)` timeout in the caller only cancels the anyio task, not the blocked thread. Consider adding a thread-level timeout mechanism.

### 8.5 `dashboard/src/routes/+page.svelte`

- **`matchesSelectedRuntime` simplification**: Changed from a branch expression to `runtime === selectedInstanceType`. This is correct and cleaner. Good fix.

- **`selectedInstanceType = "MlxIbv"` -> `"MlxJaccl"`**: The RDMA radio button now correctly sets `MlxJaccl`. Previous behavior where it set `MlxIbv` was silently falling back to Ring.

---

## 9. Recommendations

### Should fix before merge

1. **localStorage MlxIbv migration guard** (5.7) -- users upgrading will have stale values
2. **Fix `test_meta_instance_model_is_frozen`** (5.8) -- test always passes, never validates frozen behavior
3. **Add `@final` to `MetaInstanceCreated` and `MetaInstanceDeleted`** (4.5) -- consistency with other event types

### Should fix soon (follow-up)

4. **Add FIFO open timeout** (3.4) -- prevents indefinite blocking if child dies before connecting
5. **Clean up `_jaccl_pending` on instance deletion** (3.3) -- prevents memory leak
6. **Skip disk persistence for JACCL events** (5.9) -- prevents disk bloat during inference
7. **Use `tuple` instead of `list` for `MetaInstance.node_ids`** (4.6) -- enforces immutability
8. **Add `ge=1` constraint on `min_nodes`** (4.7) -- prevents invalid states
9. **Add JACCL relay tests** -- the pipe relay is the most fragile untested code
10. **Add `match` exhaustiveness to `instance_connections_healthy`** (8.2)

### Nice to have

11. **Extract dashboard display logic** -- the duplicated `{#each unifiedDisplayItems}` blocks should share a component
12. **Extract test helpers** -- move shared test fixtures to a common module
13. **Consider duplicate node detection robustness** (5.1) -- using only `friendly_name` is fragile
14. **Add `ModelCard.load` caching** -- to prevent excessive calls in the 1s reconcile loop
15. **Improve `rdmaHealthDismissed` to re-trigger** (5.10) -- like `jacclDismissedError`

---

## 10. Verdict

The architecture is sound and well-tested for the core state management paths. The MetaInstance abstraction, ProcessManager protocol, and event-sourced reconciliation are significant improvements to the codebase. The RDMA warning fix (the stated purpose of the PR) is correct.

However, the PR's scope is far larger than indicated, and several subsystems (JACCL relay, RDMA probing, FIFO management) lack test coverage. The FIFO blocking risk and `_jaccl_pending` memory leak are the most concerning practical issues.

**Recommendation**: Address the three "should fix before merge" items (localStorage migration, frozen test bug, missing @final). The "should fix soon" items should be tracked and addressed in follow-ups. The PR should either be split or the title/description should be updated to reflect the actual scope.

Issue #1356, opened 2026-02-02
Responses API: implement resumable streaming and cancellation


AlexCheema:
## Context

The Responses API streaming events now include `sequence_number` fields per the OpenAI spec, but we don't yet implement the features that depend on them.

## TODO

- **Resumable streaming**: Support the `starting_after` query parameter on `GET /v1/responses/{id}?stream=true&starting_after={sequence_number}` to resume a dropped stream. This requires persisting stream events (at least for the duration of the response).
- **Response cancellation**: Support `POST /v1/responses/{id}/cancel` to stop an in-progress response.

## Notes

- `sequence_number` is already emitted on all streaming events (monotonically increasing per response)
- No event persistence exists yet — streams are currently fire-and-forget async generators

Evanev7:
with cancellation merged, do you want to update the responses api to include this behaviour?

PR #1486, opened 2026-02-16
Conform to Responses API resumable streaming and cancellation

jayy-77:
## Motivation

[#1356](Responses API: implement resumable streaming and cancellation)

### DIFF
--- a/src/exo/master/adapters/responses.py
+++ b/src/exo/master/adapters/responses.py
@@ -1,6 +1,6 @@
 """OpenAI Responses API adapter for converting requests/responses."""
 
-from collections.abc import AsyncGenerator
+from collections.abc import AsyncGenerator, Callable
 from itertools import count
 from typing import Any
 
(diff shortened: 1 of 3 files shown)


AlexCheema:
## Review Findings

### Critical Bugs (2) — will crash at runtime

**1. Missing `ResponsesStreamEvent` import in `responses.py`**
`src/exo/master/adapters/responses.py:194` — the new `on_event` parameter uses `Callable[[ResponsesStreamEvent], None]` but `ResponsesStreamEvent` is never imported. This will raise `NameError` at runtime when `generate_responses_stream()` is called.

**Fix:** Add `ResponsesStreamEvent` to the import block from `exo.shared.types.openai_responses` (line 10–32).

**2. Wrong kwarg for `TaskCancelled` in `cancel_openai_response`**
`src/exo/master/api.py:1340` — `TaskCancelled(command_id=command_id)` passes `command_id`, but `TaskCancelled` requires `cancelled_command_id` (it's a separate field from the auto-generated base `command_id`). This will raise a Pydantic validation error at runtime when cancelling a response.

**Fix:** `TaskCancelled(cancelled_command_id=command_id)`

---

### basedpyright strict-mode errors (2)

**3. Unused import `defaultdict` in `responses_store.py`**
`src/exo/master/responses_store.py:11` — `from collections import defaultdict` is imported but never used. basedpyright `reportUnusedImport`.

**4. Protected member access `_new_event` in `responses_store.py`**
`src/exo/master/responses_store.py:58` — `ResponsesStore.mark_cancelled()` accesses `stored._new_event`, which is a protected member of `_StoredResponse`. basedpyright `reportPrivateUsage`.

**Fix:** Add a public method like `notify_waiters()` to `_StoredResponse` and call that instead.

---

### Summary

Bugs 1 and 2 are blockers — both will crash at runtime. The basedpyright issues will also fail CI (`uv run basedpyright` must pass with 0 errors per project rules).

PR #1487, opened 2026-02-16
update our tool parsers

Evanev7:
add a new abstraction for tool parsers that will make it easier to write our own going forward.

also, mlx-lm has upstreamed and improved some parsers, so we can remove our custom ones in favour of theirs for now until we revisit our parsers for pytorch compat.

### DIFF
--- a/src/exo/master/adapters/responses.py
+++ b/src/exo/master/adapters/responses.py
@@ -144,8 +144,8 @@ async def collect_responses_response(
             for tool in chunk.tool_calls:
                 function_call_items.append(
                     ResponseFunctionCallItem(
-                        id=f"fc_{tool.id}",
-                        call_id=f"call_{tool.id}",
+                        id=tool.id,
+                        call_id=tool.id,
                         name=tool.name,
                         arguments=tool.arguments,
                     )
--- a/src/exo/worker/engines/mlx/generator/generate.py
+++ b/src/exo/worker/engines/mlx/generator/generate.py
@@ -306,7 +306,7 @@ def mlx_generate(
     max_stop_len = max((len(s) for s in stop_sequences), default=0)
 
     mx_barrier(group)
-    logger.info("Ready to prefill")
+    logger.info("Starting prefill")
 
     # Prefill cache with all tokens except the last one
     prefill_tps, prefill_tokens, ssm_snapshots_list = prefill(
--- a/src/exo/worker/engines/mlx/utils_mlx.py
+++ b/src/exo/worker/engines/mlx/utils_mlx.py
@@ -353,7 +353,13 @@ def load_tokenizer_for_model_id(
             return list(hf_tokenizer.model.encode(text, allowed_special="all"))  # pyright: ignore[reportUnknownMemberType,reportUnknownArgumentType]
 
         hf_tokenizer.encode = _patched_encode
-        return TokenizerWrapper(hf_tokenizer, eos_token_ids=eos_token_ids)
+        return TokenizerWrapper(
+            hf_tokenizer,
+            eos_token_ids=eos_token_ids,
+            tool_call_start="<|tool_calls_section_begin|>",
+            tool_call_end="<|tool_calls_section_end|>",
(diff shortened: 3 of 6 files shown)


rltakashige (approved):
Thinks this is good. Jake will also review and then we can merge

AlexCheema:
## Review: PR #1487 — Update our tool parsers

**Status:** Already merged (approved by JakeHillion and rltakashige). This is a post-merge review for the record.

### Summary

This PR introduces a `ToolParser` abstraction (`tool_parsers.py`) and refactors tool call parsing in the runner. It removes custom Kimi and GLM tokenizer patches in favor of upstream `mlx_lm` parsers, while keeping a custom Kimi parser specifically for `utils_mlx.py` where the tokenizer is constructed manually.

### What's good

1. **Clean abstraction.** The `ToolParser` dataclass with `start_parsing`, `end_parsing`, and `parse_tool_calls` is a nice, minimal interface that decouples parsing logic from the streaming generator.

2. **Significant code reduction.** ~253 lines deleted vs ~178 added. The removed `patch_kimi_tokenizer`, `patch_glm_tokenizer`, `filter_kimi_tokens`, and `_validate_single_tool` functions were duplicating work now handled upstream by `mlx_lm`.

3. **Tool parser initialization moved to model load time** (lines 207-215 in `runner.py`), rather than being re-evaluated per generation. This is the right place for it.

4. **Typo fix.** "toll call parsing" -> "tool call parsing" (line 795).

5. **Responses API fix.** Removing the `fc_` and `call_` prefixes from tool call IDs in `responses.py` — the ID should pass through as-is.

6. **Better error handling.** Using `model_copy(update=...)` instead of constructing a new `GenerationResponse` for partial tool calls is cleaner and preserves all fields.

### Observations and potential concerns

1. **`startswith`/`endswith` vs exact token match.** The old code used `== tool_call_start` (exact single-token match). The new code uses `response.text.startswith(tool_parser.start_parsing)` and `endswith(tool_parser.end_parsing)`. This is more flexible but could match unintended tokens if a model produces tokens that happen to start/end with those strings. In practice this seems fine since these are specific control tokens, but worth noting the behavioral change.

2. **Start token is now included in `tool_call_text_parts`.** Previously the start token was consumed (`continue` after detecting it). Now it's appended to `tool_call_text_parts` (line 770) and later stripped by `removeprefix`/`removesuffix` in `make_mlx_parser`. This works but means the raw parser receives text with delimiters, which `_parse_kimi_tool_calls` handles correctly since it parses internal `<|tool_call_begin|>` markers.

3. **`_parse_json_calls` has redundant stripping.** It calls `removeprefix("<tool_call>")` and `removesuffix("</tool_call>")`, but `make_mlx_parser` already does the same stripping. If `json_tool_parser` is used directly (not via `make_mlx_parser`), the stripping is needed. If it goes through `make_mlx_parser`, it would double-strip (harmlessly, since the second strip is a no-op). This is fine but slightly confusing — `json_tool_parser` uses `_parse_json_calls` directly, not through `make_mlx_parser`, so the stripping is correct there.

4. **`infer_tool_parser` is defined but not called anywhere** in this PR. It's presumably intended for future use to auto-detect tool parsers from chat templates. The `TODO / example code` comment on `_parse_json_calls` suggests this is scaffolding.

5. **Bare `except Exception` in `make_mlx_parser`.** Silently returning `None` on any exception is pragmatic for tool parsing (you'd rather fall back to text than crash), but could hide bugs during development. A `logger.debug` or `logger.opt(exception=True).debug(...)` there would help with debugging.

6. **`_parse_kimi_tool_calls` uses `from typing import Any`** — this import is already present at the module level of `utils_mlx.py` so the local import of `json` inside the function is the only one needed. Minor: the function-local `import regex as re` means it's imported on every call; since this function is called at tokenizer construction time (not per-token), this is acceptable.

### Verdict

Good refactor. The `ToolParser` abstraction is clean and extensible, the upstream delegation for GLM is sensible, and the code is significantly simpler. The behavioral changes (startswith vs exact match, token accumulation) are reasonable tradeoffs.

PR #1488, opened 2026-02-16
Add --bootstrap-peer flag for SSH deployment

AlexCheema:
## Summary
- Adds `--bootstrap-peer IP:PORT` CLI flag that bypasses mDNS peer discovery, enabling exo to be launched via SSH without a GUI session (macOS TCC blocks mDNS multicast from SSH)
- The flag is additive — mDNS still runs alongside for local discovery
- Bootstrap peers are re-dialed on the existing retry loop, so dropped connections recover automatically

## Changes
- `rust/networking/src/discovery.rs` — store bootstrap addresses, dial on startup and in retry loop
- `rust/networking/src/swarm.rs` — thread `bootstrap_peers` through to `Behaviour`
- `rust/exo_pyo3_bindings/src/networking.rs` — accept optional `bootstrap_peers` in PyO3 constructor
- `rust/exo_pyo3_bindings/exo_pyo3_bindings.pyi` — updated type stub
- `src/exo/routing/router.py` — pass `bootstrap_peers` to `NetworkingHandle`
- `src/exo/main.py` — parse `--bootstrap-peer` CLI arg into multiaddr format

## Test plan
- [x] `cargo build --release` — compiles cleanly
- [x] `uv run ruff check` — all checks passed
- [x] `uv run pytest` — 261 passed, 1 skipped
- [ ] Manual: `uv run exo --bootstrap-peer 192.168.1.100:5000` accepts flag and starts node

### DIFF
--- a/rust/exo_pyo3_bindings/exo_pyo3_bindings.pyi
+++ b/rust/exo_pyo3_bindings/exo_pyo3_bindings.pyi
@@ -130,7 +130,7 @@ class Multiaddr:
 
 @typing.final
 class NetworkingHandle:
-    def __new__(cls, identity: Keypair) -> NetworkingHandle: ...
+    def __new__(cls, identity: Keypair, bootstrap_peers: builtins.list[builtins.str] = ...) -> NetworkingHandle: ...
     async def connection_update_recv(self) -> ConnectionUpdate:
         r"""
         Receives the next `ConnectionUpdate` from networking.
--- a/rust/exo_pyo3_bindings/src/networking.rs
+++ b/rust/exo_pyo3_bindings/src/networking.rs
@@ -342,8 +342,10 @@ impl PyNetworkingHandle {
     // ---- Lifecycle management methods ----
 
     #[new]
-    fn py_new(identity: Bound<'_, PyKeypair>) -> PyResult<Self> {
+    #[pyo3(signature = (identity, bootstrap_peers=vec![]))]
+    fn py_new(identity: Bound<'_, PyKeypair>, bootstrap_peers: Vec<String>) -> PyResult<Self> {
         use pyo3_async_runtimes::tokio::get_runtime;
+        use std::str::FromStr;
 
         // create communication channels
         let (to_task_tx, to_task_rx) = mpsc::channel(MPSC_CHANNEL_SIZE);
@@ -353,9 +355,20 @@ impl PyNetworkingHandle {
         // get identity
         let identity = identity.borrow().0.clone();
 
+        // parse bootstrap peer multiaddrs
+        let parsed_peers: Vec<libp2p::Multiaddr> = bootstrap_peers
+            .into_iter()
+            .map(|s| libp2p::Multiaddr::from_str(&s))
+            .collect::<Result<_, _>>()
+            .pyerr()?;
+
+        if !parsed_peers.is_empty() {
+            log::info!("RUST: bootstrap peers: {:?}", parsed_peers);
+        }
+
         // create networking swarm (within tokio context!! or it crashes)
         let swarm = get_runtime()
-            .block_on(async { create_swarm(identity) })
+            .block_on(async { create_swarm(identity, parsed_peers) })
             .pyerr()?;
 
         // spawn tokio task running the networking logic
--- a/rust/networking/src/discovery.rs
+++ b/rust/networking/src/discovery.rs
@@ -106,6 +106,9 @@ pub struct Behaviour {
     managed: managed::Behaviour,
     mdns_discovered: HashMap<PeerId, BTreeSet<Multiaddr>>,
 
+    /// Addresses provided via --bootstrap-peer, re-dialed on the retry loop.
+    bootstrap_addrs: Vec<Multiaddr>,
+
     retry_delay: Delay, // retry interval
 
     // pending events to emmit => waker-backed Deque to control polling
@@ -113,18 +116,34 @@ pub struct Behaviour {
 }
 
 impl Behaviour {
-    pub fn new(keypair: &identity::Keypair) -> io::Result<Self> {
-        Ok(Self {
+    pub fn new(keypair: &identity::Keypair, bootstrap_peers: Vec<Multiaddr>) -> io::Result<Self> {
+        let mut behaviour = Self {
             managed: managed::Behaviour::new(keypair)?,
             mdns_discovered: HashMap::new(),
+            bootstrap_addrs: bootstrap_peers,
             retry_delay: Delay::new(RETRY_CONNECT_INTERVAL),
             pending_events: WakerDeque::new(),
-        })
+        };
+
+        // Immediately dial all bootstrap peers
+        for addr in &behaviour.bootstrap_addrs.clone() {
+            behaviour.dial_addr(addr.clone());
+        }
+
+        Ok(behaviour)
     }
 
     fn dial(&mut self, peer_id: PeerId, addr: Multiaddr) {
         self.pending_events.push_back(ToSwarm::Dial {
             opts: DialOpts::peer_id(peer_id).addresses(vec![addr]).build(),
+        })
+    }
+
+    /// Dial by address only — PeerId is resolved via Noise handshake on connect.
+    /// Used for bootstrap peers where only IP:port is known.
+    fn dial_addr(&mut self, addr: Multiaddr) {
+        self.pending_events.push_back(ToSwarm::Dial {
+            opts: DialOpts::unknown_peer_id().address(addr).build(),
         })
     }
 
@@ -362,12 +381,15 @@ impl NetworkBehaviour for Behaviour {
             Poll::Pending => {}
         }
 
-        // retry connecting to all mDNS peers periodically (fails safely if already connected)
+        // retry connecting to all mDNS + bootstrap peers periodically (fails safely if already connected)
         if self.retry_delay.poll_unpin(cx).is_ready() {
             for (p, mas) in self.mdns_discovered.clone() {
                 for ma in mas {
                     self.dial(p, ma)
                 }
+            }
+            for addr in self.bootstrap_addrs.clone() {
+                self.dial_addr(addr)
             }
             self.retry_delay.reset(RETRY_CONNECT_INTERVAL) // reset timeout
         }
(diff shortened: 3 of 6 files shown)


AlexCheema:
Closing — the existing solution of launching exo via GUI session (`open /tmp/run_exo.command`) already handles mDNS peer discovery correctly. The --bootstrap-peer flag is unnecessary.

AlexCheema:
## Review: PR #1488 — Add --bootstrap-peer flag for SSH deployment

**Status:** Closed (not merged). This is a review of the diff for the record.

**Note:** This PR bundles several independent features/changes. The title says "Add --bootstrap-peer flag" but the diff includes much more.

### Scope

The PR contains at least 5 distinct changes:

1. **`--bootstrap-peer` CLI flag** — The advertised feature. Bypasses mDNS for SSH deployment.
2. **MetaInstance declarative layer** — A major new abstraction (`MetaInstance`, `ProcessManager` protocol, reconciliation loop, new API endpoints, new event types, dashboard integration).
3. **Placement strategy change** — `get_smallest_cycles` → `get_largest_cycles` — reverses the placement heuristic to prefer larger node groups.
4. **Dashboard cleanup** — Removes `MlxIbv` references throughout the dashboard.
5. **Download coordinator fix** — Detects already-completed downloads on startup (coordinator.py).

These should ideally be separate PRs for reviewability.

### 1. Bootstrap peer feature (the stated goal)

**Files:** `main.py`, `router.py`, Rust networking (`discovery.rs`, `swarm.rs`, `networking.rs`, `.pyi`)

This part is clean and well-implemented:
- `--bootstrap-peer IP:PORT` is parsed and converted to multiaddr format (`/ip4/{ip}/tcp/{port}`)
- Threaded through PyO3 bindings to Rust `create_swarm`
- Bootstrap peers are dialed on startup and re-dialed in the existing retry loop
- `dial_addr` uses `DialOpts::unknown_peer_id()` which is correct for address-only bootstrap

**Concerns:**
- Only supports a single `--bootstrap-peer`. Users wanting multiple bootstrap peers would need to extend this (though `append` is used, the argparse only accepts one value).
- IPv6 addresses with colons would break the `rsplit(":", 1)` parsing in `main.py`. Consider using a more robust parser or documenting IPv4-only.
- The `for addr in &behaviour.bootstrap_addrs.clone()` in `discovery.rs:new()` clones unnecessarily — could iterate `&self.bootstrap_addrs` directly since it's within the constructor.

### 2. MetaInstance declarative layer

**Files:** `master/main.py`, `master/process_managers/`, `master/reconcile.py`, new event types, API endpoints, dashboard

This is a significant architectural addition:
- Introduces a `ProcessManager` protocol with `reconcile(state) -> Sequence[Event]` — clean abstraction.
- Three reconcilers: `InstanceHealthReconciler`, `NodeTimeoutReconciler`, `MetaInstanceReconciler`.
- Replaces the old `_plan()` loop and `_loopback_processor` with `_apply_and_broadcast()` — this is a good simplification.
- The `_apply_and_broadcast` method applies events synchronously to state before broadcasting, which is cleaner than the old loopback channel pattern.

**Concerns:**
- The reconcile loop runs every 1 second (`await anyio.sleep(1)`) vs the old 10 seconds. This is more responsive but increases CPU overhead. Worth monitoring.
- `MetaInstanceReconciler` does `await ModelCard.load()` inside reconciliation, which means network I/O (HuggingFace API) on every reconcile cycle for unsatisfied meta instances. This could be expensive if the HF API is slow/down.
- The cascade delete in `DeleteMetaInstance` iterates `self.state.instances` while generating events that will modify instances — this is safe because events are applied after the loop, but it's worth a comment.

### 3. Placement strategy reversal

**`get_smallest_cycles` → `get_largest_cycles`** — This silently reverses a core placement heuristic. Previously exo would prefer the smallest sufficient group of nodes; now it prefers the largest. This is a significant behavioral change that affects all users, not just bootstrap-peer users. It deserves its own PR with rationale explained.

### 4. Dashboard `MlxIbv` removal

Straightforward cleanup removing dead `MlxIbvInstance` references. No concerns.

### 5. Tautological condition

Line 252 in `+page.svelte`:
```typescript
: runtime === "MlxJaccl" || runtime === "MlxJaccl";
```
This is a tautology — `MlxJaccl || MlxJaccl`. Previously it was `MlxIbv || MlxJaccl`. Looks like a find-replace error. Should just be `runtime === "MlxJaccl"`.

### 6. Other notable changes

- `mp.set_start_method("spawn", force=True)` — Adding `force=True` suppresses the error if `set_start_method` was already called. This can mask bugs where multiple entry points configure multiprocessing differently.
- `pyproject.toml` adds `--ignore=tests/start_distributed_test.py` to pytest defaults — should be explained.
- JACCL side channel handling (`_handle_jaccl_side_channel`) is new infrastructure that isn't used by the bootstrap-peer feature.

### Verdict

The `--bootstrap-peer` feature itself is well-implemented and useful. However, bundling it with the MetaInstance layer, placement reversal, and JACCL side channels makes this PR very hard to review as a unit. The tautological condition on line 252 is a bug that would ship if merged as-is.

AlexCheema:
## Code Review: PR #1488 — "Add --bootstrap-peer flag for SSH deployment"

> **Note:** This PR is closed. Posting review for reference / future work.

### Scope vs. Title

The title says "bootstrap-peer flag" but this PR contains **at least 6 independent features**:

1. `--bootstrap-peer` CLI flag (Rust + Python plumbing)
2. **MetaInstance declarative layer** — a major new abstraction (types, events, commands, API endpoints, reconciler, apply handlers)
3. **Process manager framework** — `ProcessManager` protocol + 3 reconcilers (instance health, node timeout, meta-instance)
4. **JACCL SideChannel pipe relay** — named-pipe IPC for RDMA all-gather
5. **Placement strategy change** — `get_smallest_cycles` → `get_largest_cycles`
6. **MlxIbv removal** — all dashboard + backend references removed

Each of these deserves its own PR for reviewability and safe rollback. The MetaInstance layer alone touches 15+ files.

---

### Bug: Duplicate Condition in Dashboard

In `+page.svelte`, the `matchesSelectedRuntime` function:
```typescript
: runtime === "MlxJaccl" || runtime === "MlxJaccl";
```
This is `X || X` — a tautology from find-replace of `MlxIbv`. Should be just `runtime === "MlxJaccl"` (single check, no `||`).

---

### Bootstrap Peer Feature (the advertised change)

**Files:** `main.py`, `router.py`, Rust networking (`discovery.rs`, `swarm.rs`, `networking.rs`, `.pyi`)

Clean and well-implemented:
- `--bootstrap-peer IP:PORT` parsed and converted to multiaddr `/ip4/{ip}/tcp/{port}`
- Threaded through PyO3 bindings → Rust `create_swarm`
- Bootstrap peers dialed on startup and re-dialed in existing retry loop
- `dial_addr` uses `DialOpts::unknown_peer_id()` — correct for address-only bootstrap

**Concerns:**
- Only supports a single `--bootstrap-peer`. The argparse takes one value, though `bootstrap_peers` list is used internally. Consider `nargs="+"` or `action="append"` for multi-peer.
- IPv6 addresses with colons break `rsplit(":", 1)` in `main.py`. Consider documenting IPv4-only or using a more robust parser.
- In `discovery.rs:new()`, `for addr in &behaviour.bootstrap_addrs.clone()` clones unnecessarily — could iterate `&behaviour.bootstrap_addrs` directly since we're still in the constructor.

---

### MetaInstance Declarative Layer

**Architecture (positive):**
- `ProcessManager` protocol with `reconcile(state) -> Sequence[Event]` is a clean abstraction
- `_apply_and_broadcast` replacing the old loopback channel is a nice simplification — state updates are synchronous before any await, preventing interleaving
- Failure tracking (`consecutive_failures`, `placement_error`) is well-separated from constraints
- `find_unsatisfied_meta_instances` uses binding (`meta_instance_id`) not just constraint matching — supports multiple instances of same model

**Concerns:**
- The reconcile loop runs every **1 second** (down from 10s) and calls `ModelCard.load()` (potentially a network call to HF API) for every unsatisfied meta-instance on every tick. Under failure conditions this could be expensive. Consider exponential backoff for repeated placement failures, or caching model cards.
- In `_handle_jaccl_side_channel`, `_jaccl_pending` entries for deleted instances are never cleaned up — small but unbounded memory leak if instances are deleted while side-channel rounds are pending.
- The `_update_meta_instance` helper returns the whole `meta_instances` mapping but some callers also construct their own mapping inline — minor inconsistency.

---

### Placement Strategy Change: Smallest → Largest Cycles

`get_smallest_cycles` → `get_largest_cycles` is a significant **behavioral change**. Previously placement preferred the smallest cycle that fit the model (use fewer nodes). Now it prefers the largest cycle (use more nodes). This trades compute efficiency for memory headroom but fundamentally changes cluster behavior for all users. The rationale isn't documented in the PR description or commit message.

---

### Other Notable Changes

- **`mp.set_start_method("spawn", force=True)`** — `force=True` suppresses errors if already set. Can mask issues where another library configures multiprocessing differently. Worth a comment explaining why.
- **`pyproject.toml`** adds `--ignore=tests/start_distributed_test.py` — silently excludes a test file. Not mentioned in PR description.
- **Dashboard `getMetaInstancePlacingStatus`** — retry position formula `((failures - 1) % 3) + 1` is coupled to `MAX_INSTANCE_RETRIES = 3` in Python. If that constant changes, dashboard breaks silently. Consider exposing max retries from the API.
- **`InstanceDeleted.failure_error`** default `None` — existing event logs on disk won't have this field. Pydantic handles it, but worth noting for log replay.

---

### Test Coverage

Excellent: ~1520 new test lines across `test_reconcile.py` and `test_meta_instance_edge_cases.py`:
- Lifecycle edge cases, retry counter boundaries, runner failure detection
- Connection health checks (node missing, edge removed, IP changed)
- Backward compatibility (orphan instances without `meta_instance_id`)
- Cascade delete with task cancellation, serialization roundtrip
- Error handling in reconciler (model load failures, dedup)

Gap: no tests for JACCL pipe relay logic (inherently hard to unit test).

---

### Summary

The bootstrap-peer feature itself is clean and useful. The MetaInstance layer is architecturally sound with excellent test coverage. Main concerns: **(1)** PR scope — should be split into separate PRs, **(2)** duplicate condition bug in dashboard, **(3)** aggressive 1s reconcile interval with potential network calls, **(4)** undocumented placement strategy reversal from smallest→largest cycles.

PR #1498, opened 2026-02-17
fix: recover stuck runners and finalize cancel tasks

mlpy0:
## Problem

Nodes running tensor-parallel inference get permanently stuck at 100% GPU after request cancellations. Two issues compound:

### 1. CancelTask never completes (state leak)

When the worker receives a `CancelTask`, it forwards the cancel to the runner supervisor but never emits a `TaskStatusUpdated(Complete)` event for the cancel task itself. The task stays `Pending` in `state.tasks` forever, causing unbounded growth.

**Before:** `state.tasks` grows by one entry per cancellation, never shrinking.

### 2. Runner deadlock in tensor-parallel mode (100% GPU)

In tensor-parallel (multi-node) inference, the runner process uses `mx_any()` — a collective all-reduce — to check if any peer wants to cancel the current generation. This is a blocking collective: all peers must participate.

When one node receives a cancel and the other is mid-generation step, the cancelling node exits the generation loop while the other node blocks on `mx_any()` waiting for all peers. This creates a permanent deadlock: the runner stays in `RunnerRunning` state, pinning GPU at 100% indefinitely.

There was no timeout or recovery mechanism for this scenario.

## Root Cause Analysis

**Reproduction:** Load a 70B model with tensor-parallel sharding across 2 nodes → send concurrent requests → cancel mid-stream → GPU stays at 100% on both nodes.

**Detection:** Observed via `GET /state` — `nodeSystem.gpuUsage` pinned at 1.0, `tasks` dict growing with stale `CancelTask` entries, runner status stuck at `RunnerRunning`.

**Trigger:** Any SSE stream cancellation (user closes browser tab, client timeout, concurrent request overlap) during tensor-parallel inference.

### Cancel task finalization (`main.py`)

After forwarding the cancel to the runner supervisor, emit `TaskStatusUpdated(Complete)` for the cancel task itself. This ensures the cancel task is properly removed from `state.tasks`.

### Stuck runner detection (`runner_supervisor.py`)

- Track `cancel_sent_at` timestamp when `cancel_task()` is called
- Expose `is_stuck_after_cancel` property: returns `True` if runner is still in `RunnerRunning` state >30 seconds after cancel was sent
- Reset the timestamp when runner transitions away from `RunnerRunning`

### Automatic recovery (`plan.py`)

- Add `_recover_stuck_runner()` as the **first** check in the planner's decision chain (highest priority)
- When a stuck runner is detected, return a `Shutdown` task for that runner's instance
- The existing `Shutdown` handler performs graceful termination → SIGTERM → SIGKILL, then the instance auto-recreates

### Test compatibility (`conftest.py`)

- Add `is_stuck_after_cancel` attribute to `FakeRunnerSupervisor` test double

## Recovery Flow

```
Cancel sent → cancel_sent_at = now()
  │
  ├─ Runner responds normally → status changes → cancel_sent_at = None ✓
  │
  └─ Runner deadlocks (mx_any stuck) → RunnerRunning persists
       │
       └─ 30s elapsed → is_stuck_after_cancel = True
            │
            └─ Planner detects → issues Shutdown task
                 │
                 └─ Worker kills runner process (SIGTERM/SIGKILL)
                      │
                      └─ Instance auto-recreates → GPU returns to idle ✓
```

## Testing

**Validated on 2× Mac Studio M4 Max cluster with Llama-3.3-70B tensor-parallel:**
- Soak test: 4 requests completed, 0 errors, 0% error rate
- GPU post-test: 1.1% on both nodes (not stuck)
- State clean: 0 CancelTasks in state (no leaks)
- Health: 4/4 checks passed

### DIFF
--- a/src/exo/worker/main.py
+++ b/src/exo/worker/main.py
@@ -241,6 +241,11 @@ class Worker:
                     cancelled_task_id=cancelled_task_id, runner_id=runner_id
                 ):
                     await self.runners[runner_id].cancel_task(cancelled_task_id)
+                    await self.event_sender.send(
+                        TaskStatusUpdated(
+                            task_id=task.task_id, task_status=TaskStatus.Complete
+                        )
+                    )
                 case ImageEdits() if task.task_params.total_input_chunks > 0:
                     # Assemble image from chunks and inject into task
                     cmd_id = task.command_id


Evanev7:
hiya, thanks for the PR.
the state leak is valid (though minor), but the runner does not exit the generation loop early as far as im aware.

> Runner deadlock in tensor-parallel mode (100% GPU)
> 
> In tensor-parallel (multi-node) inference, the runner process uses mx_any() — a collective all-reduce — to check if any peer wants to cancel the current generation. This is a blocking collective: all peers must participate.
> 
> When one node receives a cancel and the other is mid-generation step, the cancelling node exits the generation loop while the other node blocks on mx_any() waiting for all peers. This creates a permanent deadlock: the runner stays in RunnerRunning state, pinning GPU at 100% indefinitely
> 
> There was no timeout or recovery mechanism for this scenario.

when one node receives a cancel and the other is mid-generation step, all nodes should block on mx_any until they are all ready to exit the task, and this should happen at deterministic intervals (every n tokens, n is synchronised after warmup) so all runners should complete the mx_any at the same time. how confident are you that this is where the runners are getting stuck? we've seen the gpu lock from many other places, and this doesn't appear to be the solution.

mlpy0:
Thanks for the review and the detailed feedback.

You're right about `mx_any` — I re-read the generation loop more carefully and the collective ensures all ranks participate and exit together. My description of "one node exits while the other blocks" was inaccurate. The cancel propagates through `all_sum` so all peers see the same result.

What we observed on our 2× Mac Studio cluster (Llama-3.3-70B tensor-parallel):
- After stream cancellations, runners would occasionally stay in `RunnerRunning` with GPU pinned at 100%
- `state.tasks` accumulated stale `CancelTask` entries that were never completed

I attributed this to the cancel mechanism, but given your point about `mx_any` synchronization, the stuck runner likely has a different root cause — possibly one of the other GPU lock scenarios you've seen.

The two changes are somewhat independent:

1. **CancelTask finalization** (main.py) — the handler never emits `TaskStatusUpdated(Complete)`, so cancel tasks leak in `state.tasks`. This is valid regardless of the stuck runner question.

2. **Stuck runner recovery** (runner_supervisor.py + plan.py) — a general safety net that detects any runner stuck in `RunnerRunning` 30s after a cancel was sent and restarts it. Useful as defense-in-depth even if the root cause is elsewhere.

Happy to split these into separate PRs if that's easier to review, or adjust the approach based on your feedback. Would also be interested to hear about the other GPU lock causes you've seen — might help us track down what's actually happening on our cluster.

AlexCheema:
## Code Review — PR #1498: Recover stuck runners and finalize cancel tasks

**CI: No checks reported on branch** (needs CI run before merge)
**Author: mustafalpyilmaz** | 4 files, +32/-1

### Overview

Fixes two issues in tensor-parallel inference cancellation:
1. **CancelTask state leak** — CancelTask entries accumulate in `state.tasks` forever because no `TaskStatusUpdated(Complete)` is emitted
2. **Stuck runner recovery** — adds 30s timeout detection + automatic Shutdown for runners stuck in `RunnerRunning` after a cancel

### Important context: Evanev7's feedback

Evanev7 (project member) questioned the PR's `mx_any()` deadlock diagnosis, noting that `mx_any()` is a synchronized collective that all ranks participate in at deterministic intervals — so one node shouldn't exit the loop while the other blocks. The author conceded the `mx_any` analysis was inaccurate, but maintained both fixes are independently valid:
- The CancelTask leak is a real state bug regardless of root cause
- The stuck runner recovery is useful as defense-in-depth even if the deadlock source is elsewhere (e.g., the MLX GPU fence issue fixed in PR #1515)

This context matters: the PR fixes real symptoms but the root cause of GPU-stuck-at-100% may be the MLX fence bug rather than `mx_any()` asymmetric cancel.

---

### CancelTask finalization (main.py) — Correct

```python
case CancelTask(cancelled_task_id=cancelled_task_id, runner_id=runner_id):
    await self.runners[runner_id].cancel_task(cancelled_task_id)
    await self.event_sender.send(
        TaskStatusUpdated(task_id=task.task_id, task_status=TaskStatus.Complete)
    )
```

`task.task_id` is the CancelTask's own ID (not the cancelled task). This is correct:
- `plan_step()` line 198 emits `TaskCreated` for ALL tasks (including CancelTask) before the match block
- Without this fix, CancelTask stays Pending in `state.tasks` forever — unbounded growth
- Follows the same pattern as `CreateRunner` (lines 204–208) which also emits Complete immediately

Completion is emitted regardless of whether `cancel_task()` succeeded or the cancel pipe timed out (0.5s). Correct — the CancelTask should be finalized either way.

### Stuck runner detection (runner_supervisor.py) — Correct

Three changes:
1. **`cancel_sent_at: float | None`** — set to `anyio.current_time()` (monotonic clock) on `cancel_task()`. Good — not wall clock.
2. **`is_stuck_after_cancel` property** — requires all three: `isinstance(status, RunnerRunning)` AND `cancel_sent_at is not None` AND `>30s elapsed`. All necessary conditions.
3. **Reset in `_forward_events()`** — clears `cancel_sent_at` when runner transitions away from `RunnerRunning`. Correct.

**No race conditions**: Both write sites (`cancel_task()` and `_forward_events()`) run in the same anyio task group — cooperative concurrency, no concurrent mutation.

### Recovery logic (plan.py) — Correct

`_recover_stuck_runner()` placed **first** in the `plan()` chain (highest priority). Returns `Shutdown` for the first stuck runner. The existing Shutdown handler (main.py:227–239) escalates: close channels → CANCEL_CURRENT_TASK → join 1s → SIGTERM → SIGKILL. This terminates even completely deadlocked runners.

After Shutdown, the instance still exists in state but the runner is gone — `_create_runner()` recreates it in the next plan cycle.

### Edge cases — all safe

- **Long inference, no cancel**: `cancel_sent_at` is None → no false trigger
- **Cancel succeeds quickly**: runner transitions → `cancel_sent_at` reset
- **Runner after recovery**: fresh supervisor starts with `cancel_sent_at=None`, `status=RunnerIdle`
- **Supervisor popped on Shutdown**: dead supervisor never checked again

---

### Suggestions (non-blocking)

1. **Extract timeout constant**: `30` is hardcoded in the property. Consider `STUCK_RUNNER_TIMEOUT_SECONDS = 30` alongside existing `PREFILL_TIMEOUT_SECONDS` and `DECODE_TIMEOUT_SECONDS` at the top of `runner_supervisor.py`.

2. **Unit test**: A test for `_recover_stuck_runner()` returning `Shutdown` when `is_stuck_after_cancel=True` would be trivial and valuable:
   ```python
   runners = {rid: FakeRunnerSupervisor(..., is_stuck_after_cancel=True)}
   assert isinstance(_recover_stuck_runner(runners), Shutdown)
   ```

3. **CI required**: No checks reported on the branch. Must verify CI passes before merge.

4. **Consider splitting**: As the author offered — the CancelTask finalization fix is clear-cut and could merge independently. The stuck runner recovery is more of a safety net whose necessity may diminish now that PR #1515 (MLX fence fix) has merged.

### Verdict

**LGTM with minor suggestions.** The CancelTask finalization is a valid state leak fix — straightforward and follows existing patterns. The stuck runner recovery is well-implemented defense-in-depth (correct monotonic time tracking, no race conditions, leverages existing Shutdown escalation). The `mx_any()` deadlock diagnosis was questioned by Evanev7 and conceded by the author, but both fixes stand on their own merits. Needs CI run before merge.

AlexCheema:
## Code Review: PR #1498 — `fix: recover stuck runners and finalize cancel tasks`

### Overview

Fixes two compounding bugs that cause GPU deadlocks during tensor-parallel inference cancellations:
1. **CancelTask state leak** — CancelTask never emits `TaskStatusUpdated(Complete)`, growing `state.tasks` unboundedly
2. **Runner deadlock** — In tensor-parallel mode, `mx_any()` (collective all-reduce) deadlocks when one node cancels while another blocks on the collective

**Files changed:** `main.py`, `plan.py`, `runner_supervisor.py`, `conftest.py`

### Verdict: Looks good ✅

Well-reasoned fix with a clear root cause analysis. The 3-part approach (finalize cancel tasks, detect stuck runners, auto-recover) is sound. A few observations below.

---

### Analysis

**1. CancelTask finalization (`main.py`) — Correct ✅**

```python
case CancelTask(...):
    await self.runners[runner_id].cancel_task(cancelled_task_id)
    await self.event_sender.send(
        TaskStatusUpdated(task_id=task.task_id, task_status=TaskStatus.Complete)
    )
```

This correctly marks the *CancelTask itself* (`task.task_id`) as complete, not the cancelled task (`cancelled_task_id`). Follows the same pattern as `CreateRunner` (line 204–208) which also emits `Complete` immediately after its work.

**Minor concern:** If `self.runners[runner_id]` raises `KeyError` (runner removed between plan and execution), or `cancel_task()` raises an unexpected exception, the `TaskStatusUpdated(Complete)` is never sent and the CancelTask remains stuck. Consider wrapping in try/finally:
```python
case CancelTask(cancelled_task_id=cancelled_task_id, runner_id=runner_id):
    try:
        await self.runners[runner_id].cancel_task(cancelled_task_id)
    finally:
        await self.event_sender.send(
            TaskStatusUpdated(task_id=task.task_id, task_status=TaskStatus.Complete)
        )
```
This is a pre-existing concern (the `KeyError` path existed before this PR), so not a blocker.

**2. Stuck runner detection (`runner_supervisor.py`) — Correct ✅**

- `cancel_sent_at` set on `cancel_task()`, cleared on non-`RunnerRunning` status transitions
- `is_stuck_after_cancel` requires all three: `RunnerRunning` status, cancel was sent, and 30+ seconds elapsed
- Uses `anyio.current_time()` (monotonic event-loop clock) — correct for timeout measurement

The 30-second timeout is reasonable: long enough to avoid false positives from legitimate slow cancellations, short enough to recover within a reasonable timeframe.

**3. Recovery via planner (`plan.py`) — Correct ✅**

`_recover_stuck_runner` placed *before* `_cancel_tasks` in the evaluation chain. This is the right ordering: if a runner is stuck, sending more cancel tasks won't help — recovery (shutdown + restart) must take priority.

The `Shutdown` handler in `main.py:227–239` already handles unresponsive runners robustly: `start_task` with 3s timeout → `shutdown()` in `finally` (close channels → join → SIGTERM → join → SIGKILL). This guarantees the deadlocked process is terminated.

**4. Test double (`conftest.py`) — Correct ✅**

`FakeRunnerSupervisor` gets `is_stuck_after_cancel: bool = False` — matches the property interface without needing `anyio.current_time()` in test context.

---

### Edge Cases Considered

| Scenario | Behavior | OK? |
|----------|----------|-----|
| Cancel sent, runner responds normally | `_forward_events` sees non-`RunnerRunning` → `cancel_sent_at = None` | ✅ |
| Cancel sent, runner already not `RunnerRunning` | `is_stuck_after_cancel` returns `False` (status check fails) | ✅ |
| Multiple cancels in sequence | `cancel_sent_at` overwritten with latest; 30s window resets | ✅ |
| `is_stuck_after_cancel` called from sync `plan()` | Works — `anyio.current_time()` resolves via running event loop | ✅ |
| Runner shut down by `_recover_stuck_runner` | `runners.pop()` removes it from dict; won't be checked again | ✅ |

### Summary

- **Correctness:** ✅ Both the state leak and the deadlock recovery are properly addressed
- **Design:** ✅ Clean separation — detection in supervisor, recovery in planner, finalization in worker
- **Safety:** ✅ 30s timeout avoids false positives; shutdown uses escalating SIGTERM→SIGKILL
- **Testing:** The soak test results (4/4 requests, 0 errors, GPU back to 1.1%) are encouraging. Consider adding a unit test for `_recover_stuck_runner` in the planner tests if one doesn't already exist

Evanev7:
Alright, I'm hesitant to merge the timeout-based shutdown but obviously the task status part is an improvement so I'll merge that with this PR. We're looking into other solutions for gpu timeout (like #1515) so i'm going to leave it for now - do open an issue if this reoccurs after this merges!

PR #1500, opened 2026-02-17
Ciaran/image edit api

ciaranbor:
## Motivation

  - Image editing previously ignored input image dimensions, always defaulting to 1024x1024
  - Size dropdown was hidden in edit mode, giving users no control over output dimensions
  - Portrait/landscape presets used non-standard aspect ratios (1024x1365 / 1365x1024)

## Changes

  - Added "auto" size option that uses input image dimensions for edits, defaults to 1024x1024 for generation
  - Introduced ImageSize Literal type and normalize_image_size() validator (replaces raw str size fields)
  - Updated portrait/landscape presets to standard 1024x1536 / 1536x1024
  - Made size selector visible in edit mode (previously hidden)
  - Default size changed from "1024x1024" to "auto"

## Why It Works

  - "auto" reads actual input image dimensions via PIL at generation time, so edits preserve the original aspect ratio
  - Pydantic field_validator on both ImageGenerationTaskParams and ImageEditsTaskParams normalizes None → "auto", keeping the API backward-compatible

### Manual Testing

  - Verify image edits output at the input image's native resolution when size is "auto"
  - Verify size dropdown appears and works in both generate and edit modes

### DIFF
--- a/dashboard/src/lib/components/ImageParamsPanel.svelte
+++ b/dashboard/src/lib/components/ImageParamsPanel.svelte
@@ -225,7 +225,7 @@
 
         <!-- Dropdown Panel - fixed positioning to escape overflow:hidden -->
         <div
-          class="fixed bg-exo-dark-gray border border-exo-yellow/30 rounded shadow-lg shadow-black/50 z-[9999] max-h-48 overflow-y-auto min-w-max"
+          class="fixed bg-exo-dark-gray border border-exo-yellow/30 rounded shadow-lg shadow-black/50 z-[9999] max-h-48 overflow-y-auto overflow-x-hidden min-w-max"
           style="bottom: calc(100vh - {sizeDropdownPosition()
             .top}px + 4px); left: {sizeDropdownPosition().left}px;"
         >
@@ -310,7 +310,7 @@
 
         <!-- Dropdown Panel - fixed positioning to escape overflow:hidden -->
         <div
-          class="fixed bg-exo-dark-gray border border-exo-yellow/30 rounded shadow-lg shadow-black/50 z-[9999] max-h-48 overflow-y-auto min-w-max"
+          class="fixed bg-exo-dark-gray border border-exo-yellow/30 rounded shadow-lg shadow-black/50 z-[9999] max-h-48 overflow-y-auto overflow-x-hidden min-w-max"
           style="bottom: calc(100vh - {qualityDropdownPosition()
             .top}px + 4px); left: {qualityDropdownPosition().left}px;"
         >


AlexCheema:
## Code Review: PR #1500 — Image edit API auto-size

### Overall
Clean, well-scoped PR. The core feature (auto-size from input image dimensions) is correctly implemented, Pydantic validators maintain backward compatibility, and there's a nice bonus fix for frozen model mutation. CI is all green.

### Good

1. **Auto-size resolution is correct.** In `generate.py`, when the task is `ImageEditsTaskParams` and `size == "auto"`, the input image is decoded, PIL reads its actual dimensions, and `width`/`height` are overwritten. Ordering is correct (parse_size fallback first, then auto-detection overwrites).

2. **`normalize_image_size` validator is well-designed.** Maps `None → "auto"` for backward compat, validates against `get_args(ImageSize)` so it stays in sync with the Literal type automatically, and is shared between both task param classes and the API form handlers.

3. **Frozen model mutation fix (bonus).** Both `image_generations()` and `bench_image_generations()` previously did `payload.model = ...` (direct mutation of a frozen Pydantic model) before `model_copy()`. Now consolidated into a single `model_copy(update={...})`. Good catch.

### Should Consider

4. **"AUTO" is misleading for generation mode.** For image edits, "auto" reads the input image dimensions — clear and useful. But for generation (no input image), `parse_size("auto")` just returns `(1024, 1024)`. The "AUTO" label in the dropdown implies intelligent size detection when it's really a hardcoded fallback. Consider showing "1024×1024 (default)" for generation mode, or documenting this in a tooltip.

5. **Breaking change: old portrait/landscape sizes removed.** `"1024x1365"` and `"1365x1024"` are replaced with `"1024x1536"` / `"1536x1024"`. API callers using the old sizes will now get a validation error from `normalize_image_size`. Worth noting in release notes if there are external consumers.

6. **Uppercase X in display:** `{size.toUpperCase()}` turns `"1024x1024"` into `"1024X1024"`. Lowercase `x` is more conventional for dimensions. Minor cosmetic point.

### Minor

7. `parse_size()` now takes `ImageSize` (a Literal of known valid strings), so the `try/except` and error `raise` at the bottom are unreachable dead code. Not a bug, just could be simplified.

### Verdict

No blocking issues. Good to merge.

---

AlexCheema:
Code Review: PR #1500 — Ciaran/image edit api
===============================================

Author: ciaranbor
Status: OPEN
Changes: +145 / -105 across 5 files

Overview
--------
Adds an "auto" image size option that preserves input image dimensions for
edits (instead of always defaulting to 1024x1024). Updates portrait/landscape
presets to standard 1024x1536 / 1536x1024. Makes the size selector visible
in edit mode (previously hidden).

Files changed:
  - src/exo/shared/types/api.py — new ImageSize Literal type, normalize_image_size()
    validator, field_validators on both task params models
  - src/exo/master/api.py — use ImageSize type, normalize size in edit endpoints,
    fix model_copy mutation pattern
  - src/exo/worker/engines/image/generate.py — handle "auto" in parse_size(),
    read actual image dimensions for edits
  - dashboard/src/lib/stores/app.svelte.ts — add "auto" to type union, change
    default from "1024x1024" to "auto"
  - dashboard/src/lib/components/ImageParamsPanel.svelte — show size selector in
    edit mode, add "auto" to options, update presets, display toUpperCase()

Correctness — Type System (api.py)
------------------------------------
PASS. The ImageSize Literal type is well-defined:
    "auto" | "512x512" | "768x768" | "1024x768" | "768x1024" |
    "1024x1024" | "1024x1536" | "1536x1024"

normalize_image_size() maps None → "auto" and validates against get_args(ImageSize).
Used in three places:
  1. ImageGenerationTaskParams.normalize_size field_validator (mode="before")
  2. ImageEditsTaskParams.normalize_size field_validator (mode="before")
  3. Directly in image_edits() and bench_image_edits() API endpoints

The field_validators ensure that any None or valid string coming through the
Pydantic model gets normalized. The direct calls in the Form-based edit
endpoints handle the str | None → ImageSize conversion before passing to
_send_image_edits_command(). Correct.

Correctness — Auto Size Resolution (generate.py)
--------------------------------------------------
PASS. The auto-size logic works correctly:

1. parse_size("auto") → returns (1024, 1024) as default
2. For image edits: after saving the input image to a temp file, if
   task.size == "auto", opens the image with PIL and reads actual dimensions
3. This overrides the default (1024, 1024) with the real image dimensions

Flow for edits with "auto":
  parse_size("auto") → (1024, 1024) → decode+save image → Image.open()
  → width, height = img.size → overrides to actual dimensions ✓

Flow for generation with "auto":
  parse_size("auto") → (1024, 1024) → no input image → generates at 1024x1024 ✓

Flow for explicit sizes (any mode):
  parse_size("1024x1536") → (1024, 1536) → used directly ✓

The parse_size() signature change from `str | None` to `ImageSize` correctly
tightens the type. The "auto" check replaces the old falsy check (`if not size_str`).

Correctness — API Endpoints (api.py)
--------------------------------------
PASS. Two improvements beyond the size changes:

1. image_generations(): The model_copy mutation pattern is fixed. Old code
   mutated `payload.model` directly then called `model_copy()`. New code
   applies all changes through model_copy(update={...}) atomically. Cleaner.

2. bench_image_generations(): Same fix — moved `stream=False` and
   `partial_images=0` into the model_copy update dict instead of direct
   mutation.

3. image_edits()/bench_image_edits(): Form parameter `size` changed from
   `str = Form("1024x1024")` to `str | None = Form(None)`. The None is then
   normalized to "auto" via normalize_image_size(). This changes the API
   default behavior: edits now default to input image dimensions instead of
   1024x1024.

4. _send_image_edits_command() signature: `size: str` → `size: ImageSize`.
   Correctly tightened.

Correctness — Dashboard
------------------------
PASS. Changes are consistent:
  - ImageGenerationParams type union updated to match ImageSize
  - Default changed from "1024x1024" to "auto"
  - Size selector no longer hidden in edit mode (removed the `{#if !isEditMode}` wrapper)
  - "auto" added as first option in size dropdown
  - Display uses `.toUpperCase()` for consistent rendering ("AUTO", "512X512", etc.)
  - Added `overflow-x-hidden` to dropdown panels (minor UX fix)

Backward Compatibility
-----------------------
NOTE. Two breaking changes:

1. **Tightened size validation**: The old API accepted any "WIDTHxHEIGHT" string.
   The new API rejects sizes not in the ImageSize Literal. External clients
   passing custom sizes like "2048x2048" or the old "1024x1365" / "1365x1024"
   will get validation errors. This is a deliberate tightening — the dashboard
   only ever offered the predefined sizes, and mflux may not support arbitrary
   dimensions anyway.

2. **Default size change for edits**: Old default was "1024x1024" (always).
   New default is "auto" (uses input image dimensions). External API clients
   that relied on the 1024x1024 default for edits will now get variable-sized
   output matching their input. Clients can explicitly pass size="1024x1024"
   to preserve old behavior.

Both changes are improvements but could break external integrations. Low risk
since this is primarily a dashboard-driven feature.

Edge Cases
----------
- Auto + no input image (generation): Falls back to 1024x1024. Correct.
- Auto + corrupt/invalid image data: Image.open() would raise an exception.
  This is handled by the existing error handling in the generation pipeline.
- None size in JSON body: field_validator normalizes to "auto". Correct.
- Empty string size: Would fail normalize_image_size validation. Correct —
  empty string is not a valid ImageSize.

Nits
----
- The portrait/landscape preset change (1024x1365 → 1024x1536) is bundled
  with the auto-size feature. These are independent changes that could have
  been separate commits, but it's minor.

Verdict
-------
LGTM. Well-structured feature that addresses a real UX issue (image edits
ignoring input dimensions). The type system changes are sound, the auto-size
resolution logic is correct, and the API mutation pattern is improved. The
only notable concern is the tightened size validation which could reject
previously-accepted custom sizes, but this is a reasonable trade-off for
type safety.

AlexCheema:
## Code Review: PR #1500 — `Ciaran/image edit api`

### Overview

Adds an `"auto"` size option for image editing that preserves input image dimensions, introduces a typed `ImageSize` Literal, and makes the size selector visible in edit mode.

**Files changed:** 5 files (+145, -105)

### Verdict: Looks good overall, with one bug and a few items worth discussing

---

### Bug: `width`/`height` not reassigned in correct scope

In `generate.py`, the "auto" dimension extraction for edits works correctly:

```python
width, height = parse_size(task.size)  # "auto" → (1024, 1024) as fallback
...
if isinstance(task, ImageEditsTaskParams):
    image_path = Path(tmpdir) / "input.png"
    image_path.write_bytes(base64.b64decode(task.image_data))
    if task.size == "auto":
        with Image.open(image_path) as img:
            width, height = img.size
```

This **does** reassign `width` and `height` before they're used in the `model.generate()` call below — so the logic is correct. However, it decodes the image twice (once to write to disk, once to read back via PIL). A minor inefficiency — could open from the bytes directly with `Image.open(io.BytesIO(raw_bytes))` and read `.size` without the extra disk round-trip. Not a blocker, just a note.

### Analysis

**1. `ImageSize` Literal type and `normalize_image_size()` — Good ✅**

The `ImageSize` Literal replaces raw `str` fields, giving type safety at the Pydantic boundary. The `normalize_image_size()` standalone function + `field_validator` pattern handles both:
- Form parameters (via explicit `normalize_image_size()` call in the endpoint)
- JSON body params (via Pydantic's `field_validator` on the model)

The `None → "auto"` mapping preserves backward compatibility for callers that omit the size field.

**2. `parse_size()` refactoring — Correct ✅**

Changing the `if not size_str` check to `if size_str == "auto"` is correct since `ImageSize` can't be `None` or empty. The try/except parsing block is now technically unreachable for valid `ImageSize` values (they're all pre-validated), but it's harmless as defense-in-depth.

**3. API backward compatibility — Mostly preserved, one breaking edge**

| Caller behavior | Before | After | Breaking? |
|----------------|--------|-------|-----------|
| Omits `size` entirely | 1024x1024 | "auto" (1024x1024 for gen, native for edits) | No for gen, **behavior change for edits** |
| Sends `size: null` | 1024x1024 | "auto" → same as above | Same |
| Sends `size: "1024x1024"` | 1024x1024 | 1024x1024 | No |
| Sends `size: "1024x1365"` | Accepted | **Rejected** (not in `ImageSize`) | **Yes** |

The old `"1024x1365"` / `"1365x1024"` portrait/landscape presets are removed. External API consumers using these sizes will get validation errors. This seems intentional (correcting to standard `1024x1536` / `1536x1024`), but worth noting in release notes.

**4. `model_copy` refactoring in `api.py` — Good ✅**

Replacing direct field mutations (`payload.model = ...`, `payload.stream = False`) with a single `model_copy(update={...})` is cleaner and avoids mutating the original object. Good cleanup.

**5. Dashboard changes — Correct ✅**

- Size dropdown now visible in edit mode (previously hidden by `{#if !isEditMode}`)
- `"auto"` added as first option, displayed as `"AUTO"` via `.toUpperCase()`
- Default changed from `"1024x1024"` to `"auto"`

**Minor localStorage concern:** Users with previously saved params containing `"1024x1365"` or `"1365x1024"` will have a stale value loaded from localStorage (via `{ ...DEFAULT_IMAGE_PARAMS, ...parsed }`). The dropdown won't show a checkmark for the old value, and the API will reject it. This self-resolves once the user selects a new size, but could cause a confusing first-time error. Consider adding a migration check:
```typescript
if (parsed.size && !sizeOptions.includes(parsed.size)) {
  parsed.size = "auto";
}
```

**6. No max dimension guard for "auto"**

When `size="auto"`, the output dimensions come directly from the input image. A very large input image (e.g., 8000x6000) would attempt generation at that resolution, potentially causing OOM. This is a pre-existing concern (the backend had no dimension limits before), but "auto" makes it easier to trigger. Consider clamping to a max dimension in a follow-up.

---

### Summary

- **Correctness:** ✅ The "auto" size logic works correctly for both generation (1024x1024 fallback) and edits (native input dimensions)
- **Type safety:** ✅ `ImageSize` Literal + `field_validator` is the right pattern
- **Breaking change:** ⚠️ Old `1024x1365`/`1365x1024` sizes are rejected — worth a mention in release notes
- **Suggestions:** Consider localStorage migration for stale size values, and a max dimension guard for "auto" in a follow-up

AlexCheema:
## Code Review: PR #1500 — Ciaran/image edit api

### Summary
Fixes image editing to respect input image dimensions (was always defaulting to 1024x1024), adds an "auto" size option, updates portrait/landscape presets to standard ratios, and makes the size selector visible in edit mode.

### Review

**"auto" size option is a good UX improvement:**
- For edits: uses input image dimensions via PIL ✅
- For generation: defaults to 1024x1024 ✅
- Dashboard default changed from "1024x1024" to "auto" ✅

**`ImageSize` Literal type and `normalize_image_size()`:**
- Adds type safety to size parameters ✅
- Pydantic validator normalizes `None` → `"auto"` for backward compatibility ✅

**Preset updates:**
- `1024x1365` / `1365x1024` → `1024x1536` / `1536x1024` (standard DALL-E ratios) ✅

**Dashboard changes:**
- Size selector now visible in edit mode (was hidden behind `{#if !isEditMode}`) ✅
- `.toUpperCase()` for size display ✅

### Issues

**1. Need to see the full backend diff**
The diff was truncated at 449 lines. The backend changes to `api.py` show the `size: str | None = Form(None)` change for the edits endpoint, and the `model_copy` refactoring for frozen models. Would need to see the full PIL dimension reading logic to verify correctness.

**2. `payload.model_copy(update={...})` is better than direct mutation**
The PR correctly fixes the pattern from `payload.model = ...` (mutating frozen model) to `payload = payload.model_copy(update={...})`. This is a good fix independent of the image sizing feature.

### Verdict
Good UX improvement with proper type safety. The "auto" default is user-friendly.

LGTM (pending full diff review for the PIL dimension logic).

AlexCheema:
## Code Review -- PR #1500: Ciaran/image edit api

**CI:** All 3 checks passing (aarch64-darwin, x86_64-linux, aarch64-linux)

### Overview

Adds an `"auto"` image size option so that image edits preserve the input image's native dimensions instead of always defaulting to 1024x1024. Introduces a typed `ImageSize` Literal, a shared `normalize_image_size()` validator, updates portrait/landscape presets to standard ratios, and makes the size selector visible in edit mode on the dashboard. Also fixes direct mutation of frozen Pydantic models in `image_generations()` and `bench_image_generations()`.

**Files changed:** 5 (+145, -105)

---

### Critical Issues

None found.

### Significant Issues

**1. No upper-bound guard on "auto" dimensions for edits**

When `size="auto"` and the user uploads a very large image (e.g. 8000x6000 from a DSLR), the backend will attempt to generate at that resolution, which will almost certainly OOM or produce garbage output from mflux. The old hardcoded 1024x1024 default acted as an implicit safety cap.

```python
# generate.py
if task.size == "auto":
    with Image.open(image_path) as img:
        width, height = img.size  # could be anything
```

Suggestion: clamp dimensions to a sensible maximum (e.g. 2048 on the longest side, preserving aspect ratio), or at minimum document that "auto" passes through raw input dimensions.

**2. Stale localStorage values will cause API rejection**

Users who previously used `"1024x1365"` or `"1365x1024"` will have those values persisted in `localStorage` under `exo-image-generation-params`. On next load, these stale sizes will be sent to the API and rejected by `normalize_image_size()`. The dashboard should sanitize restored params:

```typescript
// Somewhere in the localStorage restoration logic:
if (parsed.size && !sizeOptions.includes(parsed.size)) {
  parsed.size = "auto";
}
```

This is a real user-facing bug for anyone who used portrait/landscape presets before this change.

### Minor Issues

**3. `parse_size()` error paths are now dead code**

After this PR, `parse_size()` accepts `ImageSize` (a Literal of pre-validated strings). The `try/except` block and the `raise ValueError(...)` at the bottom are unreachable since all valid `ImageSize` values (except `"auto"`, handled above) will always parse correctly. Not a bug -- just dead code that could be cleaned up in a follow-up.

**4. `.toUpperCase()` makes dimension separators look odd**

`"1024x1024".toUpperCase()` produces `"1024X1024"`. The conventional separator for dimensions is lowercase `x`. Consider:

```typescript
{size === "auto" ? "AUTO" : size}
```

**5. Redundant disk round-trip for reading image dimensions**

In `generate.py`, the image bytes are decoded from base64, written to disk, then re-read via `Image.open(image_path)` to get dimensions. Since the bytes are already in memory:

```python
raw_bytes = base64.b64decode(task.image_data)
image_path = Path(tmpdir) / "input.png"
image_path.write_bytes(raw_bytes)
if task.size == "auto":
    with Image.open(io.BytesIO(raw_bytes)) as img:
        width, height = img.size
```

Not a correctness issue, just an unnecessary I/O round-trip.

**6. Backward compatibility: old sizes now rejected**

`"1024x1365"` and `"1365x1024"` are removed from the allowed set. External API consumers using those sizes will get validation errors. Worth mentioning in release notes.

### What's Good

- **`ImageSize` Literal type** provides compile-time safety. Using `get_args(ImageSize)` in the validator keeps validation automatically in sync with the type definition -- no separate list to maintain.
- **`normalize_image_size()` as a shared function** avoids duplication between Pydantic `field_validator` usage and the Form-parameter endpoints.
- **`model_copy()` consolidation** correctly fixes the frozen-model mutation pattern in `image_generations()` and `bench_image_generations()`. The old code did `payload.model = ...` which would fail at runtime with frozen Pydantic models.
- **Auto-size resolution logic is correct.** `parse_size("auto")` returns `(1024, 1024)` as fallback, then for edits the PIL `.size` read overwrites with actual dimensions before they are passed to `model.generate()`. For generation (no input image), the 1024x1024 default applies. The ordering and scoping are right.
- **Dashboard changes are clean.** Removing the `{#if !isEditMode}` wrapper to show the size selector in edit mode is the right UX fix. The `"auto"` option being first in the list makes it the natural default.

### Verdict

Good feature with solid implementation. The `ImageSize` Literal + shared validator pattern is well-designed and the frozen-model mutation fix is a welcome bonus. Two items worth addressing before merge: **(1)** adding a max dimension clamp for "auto" to prevent OOM on large input images, and **(2)** sanitizing stale localStorage values so existing users don't hit an immediate error after upgrading. The rest are minor polish items.

PR #1511, opened 2026-02-17
fix: unblock MpReceiver.close() to prevent shutdown hang

AlexCheema:
## Summary

- `MpReceiver.close()` did not unblock threads stuck on `queue.get()` in `receive_async()`, causing abandoned threads (via `abandon_on_cancel=True`) to keep the Python process alive indefinitely after tests pass
- This caused the `aarch64-darwin` CI jobs in PR #1462 to hang for ~6 hours until the GitHub Actions timeout killed them
- Sends an `_MpEndOfStream` sentinel before closing the buffer, mirroring what `MpSender.close()` already does

## Test plan

- [x] `uv run ruff check` — clean
- [x] `nix fmt` — 0 changed
- [x] `uv run pytest` — 188 passed, 1 skipped in 12s (no hang)

### DIFF
--- a/src/exo/utils/channels.py
+++ b/src/exo/utils/channels.py
@@ -1,3 +1,4 @@
+import contextlib
 import multiprocessing as mp
 from dataclasses import dataclass, field
 from math import inf
@@ -132,7 +133,8 @@ class MpSender[T]:
     def close(self) -> None:
         if not self._state.closed.is_set():
             self._state.closed.set()
-        self._state.buffer.put(_MpEndOfStream())
+        with contextlib.suppress(Exception):
+            self._state.buffer.put_nowait(_MpEndOfStream())
         self._state.buffer.close()
 
     # == unique to Mp channels ==
@@ -204,10 +206,8 @@ class MpReceiver[T]:
     def close(self) -> None:
         if not self._state.closed.is_set():
             self._state.closed.set()
-        try:  # noqa: SIM105
+        with contextlib.suppress(Exception):
             self._state.buffer.put_nowait(_MpEndOfStream())
-        except Exception:
-            pass
         self._state.buffer.close()
 
     # == unique to Mp channels ==


AlexCheema:
## Review: PR #1511 — fix: unblock MpReceiver.close() to prevent shutdown hang

### Summary

Fixes a real hang where abandoned threads blocked on `queue.get()` inside `MpReceiver.receive()` prevent the Python process from exiting. The fix sends an `_MpEndOfStream` sentinel in `MpReceiver.close()` before closing the buffer, mirroring what `MpSender.close()` already does.

### Analysis

**The bug is well-diagnosed.** The hang path is:

1. `receive_async()` runs `receive()` in a thread with `abandon_on_cancel=True`
2. Async task gets cancelled → thread is abandoned but stays blocked on `queue.get()` (line 193)
3. `MpReceiver.close()` calls `buffer.close()`, but `mp.Queue.close()` only prevents new `put()` calls — it does **not** unblock threads waiting on `get()`
4. Abandoned thread holds the process alive indefinitely

**The fix is correct.** Injecting an `_MpEndOfStream` sentinel before `buffer.close()` unblocks the waiting thread, which then sees the sentinel, calls `self.close()`, and raises `EndOfStream` — exactly the existing shutdown protocol.

**The `try/except Exception: pass` is appropriate** for this shutdown path. `put_nowait()` could raise `Full` (buffer at capacity) or `ValueError` (buffer already closed), and neither should prevent cleanup from proceeding. The `# noqa: SIM105` for not using `contextlib.suppress` is fine.

**`put_nowait` vs `put` is the right choice.** `MpSender.close()` uses blocking `put()` (line 135), which makes sense for a sender that needs to guarantee the sentinel lands. For the receiver close path, `put_nowait()` is better — we don't want to block the closing thread if the buffer is full, and best-effort delivery is sufficient here.

### Minor observations

**1. Multiple abandoned threads:** If multiple `receive_async()` calls are abandoned (each spawns a fresh `CapacityLimiter(1)`), only one gets unblocked by the single sentinel. The first thread to consume it calls `self.close()`, which tries to put another sentinel (silently caught), but `buffer.close()` then prevents further puts. In practice this is unlikely to be an issue — the common case is one abandoned thread — but worth noting.

**2. Slight asymmetry with MpSender.close():** `MpSender.close()` doesn't guard its `put(_MpEndOfStream())` with try/except, meaning it would propagate if the buffer is already closed. Both methods could benefit from consistent error handling, but that's a pre-existing concern, not introduced here.

### Verdict

Clean, minimal, correct fix for a real CI-blocking hang. The sentinel approach is consistent with the existing shutdown protocol. LGTM.

AlexCheema:
## Code Review: PR #1511 — fix: unblock MpReceiver.close() to prevent shutdown hang

### Summary

Clean, well-motivated 4-line fix that addresses a real CI hang. The root cause analysis is correct and the fix mirrors the existing `MpSender.close()` pattern.

### The Bug

`RunnerSupervisor._forward_events()` iterates `async for event in self._ev_recv`, which calls `MpReceiver.receive_async()` → `to_thread.run_sync(self.receive, ..., abandon_on_cancel=True)`. The thread inside `receive()` blocks on `self._state.buffer.get()`. When `shutdown()` calls `self._ev_recv.close()`, the old code only set the closed flag and called `buffer.close()` — but `mp.Queue.close()` does **not** interrupt a blocked `get()`. The abandoned thread keeps the process alive indefinitely.

### The Fix

Sending `_MpEndOfStream` via `put_nowait` before closing the buffer is correct. The sentinel unblocks `get()`, and the existing logic in `receive()` (lines 194-196) already handles the sentinel by calling `self.close()` and raising `EndOfStream`.

### Analysis

**`put_nowait` is the right choice over `put`:** A thread can only be stuck on `get()` when the queue is empty, which means `put_nowait` will always succeed in the case that matters. Using blocking `put` in a `close()` method could itself hang if the queue were full.

**Broad `except Exception: pass` — acceptable here:** In a `close()` method, resilience matters more than precision. The exception could be `Full` (queue at capacity — but then no thread is blocked on `get()` anyway), `ValueError` / `OSError` (buffer already closed by sender side), etc. A narrower catch like `except (Full, ValueError, OSError)` would be slightly more self-documenting, but the broad catch is defensible for teardown code.

**Double-close interaction with `MpSender.close()`:** If the sender already called `close()` (put sentinel + closed buffer), the receiver's `put_nowait` will raise (caught and ignored) and `buffer.close()` is idempotent. No issue.

### Minor Observations

**1. Multiple abandoned threads:** Each `receive_async()` call creates its own `CapacityLimiter(1)`, so in theory multiple abandoned threads could accumulate. Only one sentinel is sent, so only one thread would be unblocked. However, `buffer.close()` will cause subsequent `get()` calls to eventually raise, and in practice the `async for` pattern produces at most one abandoned thread at a time. Minor theoretical gap, not a practical concern.

**2. Slight asymmetry with `MpSender.close()`:** The sender's `close()` (line 135) does an unguarded `self._state.buffer.put(_MpEndOfStream())` — no try/except. If the buffer were already closed, the sender's close would propagate the exception. Both sides could benefit from consistent error handling, but that's a pre-existing concern, not introduced by this PR.

### Suggestion (non-blocking)

The existing test in `test_mp_channel.py` only covers the sender-close path. A small test verifying `MpReceiver.close()` unblocks a waiting `receive_async()` would prevent regression, e.g.:

```python
@pytest.mark.anyio
async def test_receiver_close_unblocks():
    """Verify MpReceiver.close() unblocks a thread stuck in receive_async()."""
    from anyio import fail_after, EndOfStream
    _, r = mp_channel[str]()
    with pytest.raises(EndOfStream):
        async with fail_after(1):
            await r.receive_async()
```

But for a 4-line fix that mirrors existing sender behavior and is validated by the full test suite no longer hanging, this is optional.

### Verdict

**LGTM.** The fix is minimal, correct, and directly addresses the 6-hour CI hang. No blocking concerns.

AlexCheema:
## Code Review: PR #1511 — `fix: unblock MpReceiver.close() to prevent shutdown hang`

### Overview

Sends an `_MpEndOfStream` sentinel in `MpReceiver.close()` to unblock threads stuck on `queue.get()`, mirroring what `MpSender.close()` already does.

**Files changed:** `src/exo/utils/channels.py` (+4, -0)

### Verdict: Looks good ✅

---

### Root Cause

1. `receive_async()` dispatches `self.receive()` to a thread via `to_thread.run_sync(..., abandon_on_cancel=True)`
2. `self.receive()` blocks on `self._state.buffer.get()` (a `multiprocessing.Queue`)
3. When `MpReceiver.close()` was called, it set the `closed` flag and called `buffer.close()`
4. But `mp.Queue.close()` does **not** unblock a thread already stuck on `.get()` — it only prevents future operations
5. The abandoned thread (kept alive by `abandon_on_cancel=True`) holds the Python process open indefinitely

### Fix Analysis

```python
def close(self) -> None:
    if not self._state.closed.is_set():
        self._state.closed.set()
    try:
        self._state.buffer.put_nowait(_MpEndOfStream())
    except Exception:
        pass
    self._state.buffer.close()
```

The sentinel unblocks the `buffer.get()` call. The blocked thread receives it, `receive()` sees `isinstance(item, _MpEndOfStream)`, calls `self.close()` (idempotent), and raises `EndOfStream` — thread exits cleanly.

**Symmetry with `MpSender.close()`:** The sender already does this at line 135 (`self._state.buffer.put(_MpEndOfStream())`). The receiver was missing the same step. Now both sides are consistent.

**Broad `except Exception: pass` — Correct for this context:**
- Buffer may already be closed (`ValueError`)
- Buffer may be full (`Full`) — unlikely at shutdown, but possible
- This is best-effort cleanup during shutdown; swallowing errors is appropriate

**`# noqa: SIM105`:** Suppresses ruff's suggestion to use `contextlib.suppress()`. Functionally equivalent; the inline try/except is fine.

### Edge Cases

| Scenario | Behavior |
|----------|----------|
| Thread blocked on `get()` | Sentinel unblocks it → `EndOfStream` → thread exits ✅ |
| No thread blocked | Sentinel sits on queue, `buffer.close()` cleans up ✅ |
| `close()` called twice | First call sets flag + sends sentinel; second call: flag already set, sentinel `put_nowait` may fail → caught by `except` ✅ |
| Sender already closed and sent sentinel | Receiver may consume sender's sentinel first; receiver's sentinel is redundant but harmless ✅ |

### Summary

- **Correctness:** ✅ Mirrors the established `MpSender.close()` pattern
- **Minimal:** ✅ 4 lines, no behavioral changes to normal operation
- **Impact:** Fixes CI hangs caused by zombie threads holding the process open after tests complete

AlexCheema:
## Code Review -- PR #1511: fix: unblock MpReceiver.close() to prevent shutdown hang

**CI status:** All checks passing (aarch64-darwin, aarch64-linux, x86_64-linux -- all SUCCESS)

### Overview

This PR fixes a CI hang where `MpReceiver.close()` failed to unblock threads stuck on `queue.get()` inside `receive_async()`. The fix sends an `_MpEndOfStream` sentinel via `put_nowait()` before closing the buffer, mirroring the pattern already established by `MpSender.close()`.

**Changes:** +4 / -0 in `src/exo/utils/channels.py`

### Root Cause Verification

The diagnosed hang path is correct:

1. `_forward_events()` (line 240) iterates `async for event in self._ev_recv`, which calls `receive_async()` -> `to_thread.run_sync(self.receive, ..., abandon_on_cancel=True)`
2. `receive()` (line 191) falls through to `self._state.buffer.get()` -- a blocking call on `multiprocessing.Queue`
3. When `shutdown()` (line 184) calls `self._ev_recv.close()`, the old code set the `closed` flag and called `buffer.close()`, but `mp.Queue.close()` does **not** unblock a pending `get()` -- it only prevents future `put()` operations
4. The abandoned thread (due to `abandon_on_cancel=True`) stays alive, holding the Python process open indefinitely

### Fix Analysis

The patched `close()`:

```python
def close(self) -> None:
    if not self._state.closed.is_set():
        self._state.closed.set()
    try:  # noqa: SIM105
        self._state.buffer.put_nowait(_MpEndOfStream())
    except Exception:
        pass
    self._state.buffer.close()
```

**Correctness:** The sentinel unblocks `buffer.get()`. The blocked thread receives it, `receive()` sees `isinstance(item, _MpEndOfStream)` (line 192), calls `self.close()` (idempotent), and raises `EndOfStream` -- thread exits cleanly. This mirrors exactly what `MpSender.close()` does at line 133.

**`put_nowait` vs `put`:** Good choice. The sender uses blocking `put()` (line 133), but in the receiver's close path, `put_nowait()` is better -- using blocking `put()` in a `close()` method could itself deadlock if the buffer were somehow full. Since a thread can only be stuck on `get()` when the queue is empty or has been drained, `put_nowait()` will succeed in the case that actually matters.

**Broad `except Exception: pass`:** Appropriate for a shutdown path. The `put_nowait()` could raise `Full` (queue at capacity), `ValueError` (buffer already closed), or `OSError`/`BrokenPipeError` (broken pipe). None of these should prevent cleanup from proceeding. The `# noqa: SIM105` to suppress the "use `contextlib.suppress`" suggestion is fine.

### Issues

**No blocking issues found.**

**Minor observation -- `MpSender.close()` has an unguarded `put()` (line 133):**

```python
# MpSender.close() at line 130-134:
def close(self) -> None:
    if not self._state.closed.is_set():
        self._state.closed.set()
    self._state.buffer.put(_MpEndOfStream())  # no try/except, blocking put
    self._state.buffer.close()
```

The sender's `close()` uses an unguarded blocking `put()` -- if the buffer is already closed (e.g., receiver closed first), this would propagate an exception, and if the buffer is full, this could block. The receiver's new code with `put_nowait` + `try/except` is actually more robust. Not introduced by this PR, but worth noting as a potential follow-up to make the sender consistent.

**Suggestion (non-blocking) -- regression test:**

The existing `test_mp_channel.py` only tests the sender-close path (sender sends messages then calls `close()`; receiver iterates until `StopIteration`). A small test confirming that `MpReceiver.close()` unblocks a waiting `receive_async()` would prevent regression:

```python
@pytest.mark.anyio
async def test_receiver_close_unblocks():
    """Verify MpReceiver.close() unblocks a thread stuck in receive_async()."""
    s, r = mp_channel[str]()
    async def close_after_delay():
        await anyio.sleep(0.1)
        r.close()
    async with anyio.create_task_group() as tg:
        tg.start_soon(close_after_delay)
        with pytest.raises(EndOfStream):
            await r.receive_async()
    s.close()
```

This is optional given that the full test suite no longer hangs, which is itself validation.

### What's Good

- **Minimal and targeted** -- 4 lines that fix a real 6-hour CI hang
- **Mirrors established pattern** -- `MpSender.close()` already does the same sentinel approach, so this makes both sides symmetric
- **Defensive** -- `put_nowait` + `try/except` is more robust than the sender's unguarded blocking `put`
- **Root cause well-diagnosed** in the PR description with clear explanation of the hang path

### Verdict

LGTM. Clean, minimal fix that correctly addresses the root cause. The sentinel pattern is already proven by `MpSender.close()`, and this PR makes `MpReceiver.close()` symmetric. No blocking concerns.

Evanev7:
Can you also add this to the MpSender.close()?

AlexCheema:
## Review Summary

**Verdict: LGTM** — Surgical 4-line fix for a real CI issue (6-hour hangs on `aarch64-darwin`).

### What it does

`MpReceiver.close()` now sends an `_MpEndOfStream` sentinel before closing the buffer, unblocking threads stuck on `queue.get()` in `receive_async()`. This mirrors what `MpSender.close()` already does.

### Analysis

- **`put_nowait()` instead of `put()`**: Correct — a blocking `put()` in `close()` could deadlock if the buffer is full
- **`try/except Exception: pass`**: Correct — the queue might already be closed, buffer full, or pipe broken. The `buffer.close()` right after is the hard shutdown
- **Edge cases**: Multiple `close()` calls are idempotent. No thread waiting means sentinel sits in buffer harmlessly. Race between `close()` and `receive_nowait()` is benign (both paths lead to clean termination)
- **No regressions**: Purely additive (4 lines, 0 deletions). Sentinel pattern already used throughout this file

Already approved by Evanev7. CI 6/6 green. This unblocks #1462 and #1464 which both suffer from the same 6-hour `aarch64-darwin` timeout.

*Automated review — not a merge/approve action.*

PR #1512, opened 2026-02-17
Fix partial download progress showing 0% on restart

### LINKED ISSUE #1042: [BUG] Initial download statuses don't support partial download progress
## Describe the bug

If you go to the dashboard initially it will either show models as 0% downloaded or 100% downloaded, i.e. it doesn't show any partial downloads.

## To Reproduce

1. Partially download a model
2. Quit exo
3. Relaunch exo
4. Go to model download page on dashboard
5. It will show 0% downloaded

## Expected behavior

Should show the partial download progress of each model.

## Actual behavior

It shows 0%

## Environment

Any environment.

## Additional context

N/A

AlexCheema:
## Summary
- Fixes partially downloaded models showing 0% progress after restarting exo
- Root cause: `_emit_existing_download_progress()` checked `downloaded_bytes_this_session` (always 0 in a new session) instead of `downloaded_bytes` (actual data on disk) to decide if a download was pending
- One-line fix: check `downloaded_bytes.in_bytes == 0` instead of `downloaded_bytes_this_session.in_bytes == 0`

Closes #1042

## Test plan
- [x] `basedpyright` passes with 0 errors
- [x] `ruff check` passes
- [x] `nix fmt` clean (0 changed)
- [x] `pytest` passes (260 passed, 1 skipped)
- [ ] Partially download a model, quit exo, relaunch, verify the downloads page shows actual progress instead of 0%

### DIFF
--- a/src/exo/download/coordinator.py
+++ b/src/exo/download/coordinator.py
@@ -398,7 +398,17 @@ class DownloadCoordinator:
                             ),
                         )
                     elif progress.status in ["in_progress", "not_started"]:
-                        if progress.downloaded_this_session.in_bytes == 0:
+                        if (
+                            progress.downloaded.in_bytes
+                            >= progress.total.in_bytes
+                            > 0
+                        ):
+                            status = DownloadCompleted(
+                                node_id=self.node_id,
+                                shard_metadata=progress.shard,
+                                total=progress.total,
+                            )
+                        elif progress.downloaded.in_bytes == 0:
                             status = DownloadPending(
                                 node_id=self.node_id,
                                 shard_metadata=progress.shard,


AlexCheema:
Code Review: PR #1512 — Fix partial download progress showing 0% on restart
=============================================================================

Author: AlexCheema
Status: OPEN (auto-merge enabled, squash)
Changes: +1 / -1 in src/exo/download/coordinator.py

Overview
--------
One-line fix: in `_emit_existing_download_progress()`, changes the condition
that decides whether a partially-downloaded model should be reported as
DownloadPending vs DownloadOngoing. Previously checked
`downloaded_bytes_this_session.in_bytes == 0` (always 0 on fresh session);
now checks `downloaded_bytes.in_bytes == 0` (actual bytes on disk).

Correctness
-----------
PASS. Root cause is clear and the fix is correct.

The function is called on startup to emit the current download state. The logic
at line 316-335 handles `status in ["in_progress", "not_started"]`:
  - If downloaded_bytes == 0 → DownloadPending (nothing on disk yet)
  - If downloaded_bytes > 0 → DownloadOngoing (partial data exists)

The old code checked `downloaded_bytes_this_session` which is always 0 at
startup since no downloading has occurred in the current session. This caused
all partial downloads to appear as DownloadPending (0%), even if gigabytes
of data existed on disk.

The fix is minimal and surgical. `downloaded_bytes` reflects the actual on-disk
state from the downloader's progress tracking.

Edge Cases
----------
- Fresh download (nothing on disk): downloaded_bytes.in_bytes == 0 → still
  correctly emits DownloadPending. No behavior change.
- Completed download: handled by the `status == "completed"` branch above
  (line 304-315). Not affected.
- Corrupted partial download: would still show as DownloadOngoing if any bytes
  exist. This is pre-existing behavior and not this PR's concern.

Nits
----
None.

Verdict
-------
LGTM. Clean, correct one-line fix. The root cause is obvious once identified —
`downloaded_bytes_this_session` vs `downloaded_bytes` is a classic session-scoped
vs persistent-state confusion.

AlexCheema:
This needs to be tested very carefully.
I would think the correct fix here is to introduce another state which is not `DownloadOngoing` which represents a download that was partially downloaded but not downloading right now. That, or we can add more information to `DownloadPending` e.g. `downloaded_bytes` etc...

AlexCheema:
## Code Review -- PR #1512: Fix partial download progress showing 0% on restart

**CI Status:** All 3 checks passing (aarch64-darwin, x86_64-linux, aarch64-linux)

### Overview

One-line fix in `_emit_existing_download_progress()` (`coordinator.py:327`). On restart, partially downloaded models showed 0% because the condition checked `downloaded_bytes_this_session` (reset to 0 every session) instead of `downloaded_bytes` (actual bytes on disk). The fix changes the check to use the persistent counter.

### Analysis

**The fix is correct for the stated bug.** The scan code at `download_utils.py:789` explicitly sets `downloaded_this_session=Memory.from_bytes(0)` for every file on startup, so `downloaded_bytes_this_session` is always 0 at the time `_emit_existing_download_progress()` first runs. Checking `downloaded_bytes` instead uses the real on-disk total (populated via `get_downloaded_size()` at line 783).

The branching logic after the fix:
- `status == "complete"` -> `DownloadCompleted` (line 307)
- `downloaded_bytes >= total_bytes > 0` -> `DownloadCompleted` (line 317, catches completions missed by status)
- `downloaded_bytes == 0` -> `DownloadPending` (the fixed line)
- `downloaded_bytes > 0` -> `DownloadOngoing` with progress data (line 336)

All four branches are correct.

### Significant Issue

**`DownloadOngoing` on restart will show speed=0 and ETA=0, which may be misleading.** When a download is partially complete but *not actively downloading*, the `DownloadOngoing` event will carry `DownloadProgressData` with `speed=0` and `eta_ms=0` (because `downloaded_bytes_this_session=0` means `all_speed=0` in `build_repo_download_progress` at `download_utils.py:580`). The dashboard will render this as an active download with 0 B/s speed and no ETA.

This is arguably better than the current bug (showing 0%), but the author (AlexCheema) raised the same concern in the comments: a model that is partially downloaded but not currently downloading is semantically neither "pending" nor "ongoing" -- it is *paused* or *stalled*. The `DownloadOngoing` type implies active transfer.

This is a pre-existing design limitation, not a regression introduced by this PR. The fix is strictly an improvement over the status quo. Whether to address the semantic gap with a new state (e.g., `DownloadPaused`) is a separate design decision.

### Edge Cases

| Scenario | Before fix | After fix |
|---|---|---|
| Restart, 50% downloaded | DownloadPending (0%) | DownloadOngoing (50%, speed=0) |
| Restart, 0% downloaded | DownloadPending | DownloadPending |
| Complete download | DownloadCompleted | DownloadCompleted |
| Active download mid-session | DownloadOngoing | DownloadOngoing (no change -- skipped by `active_downloads` guard) |

All correct.

### What's Good

- Minimal, surgical fix -- exactly one token changed (`downloaded_bytes` vs `downloaded_bytes_this_session`)
- Root cause clearly identified and explained in the PR description
- All pre-commit checks pass
- Closes a real user-facing bug (#1042)

### Verdict

LGTM. The fix is correct and strictly improves the user experience. The semantic question of whether to introduce a `DownloadPaused` state (as noted by the author in the comments) is worth considering as a follow-up, but it should not block this fix.

PR #1517, opened 2026-02-17
Handle MessageTooLarge error in router to prevent node crash

### LINKED ISSUE #1296: [BUG] Large messages cause routing layer RuntimeError
## Problem Description

When a request sends a large amount of characters (1,000,000+), the `_networking_publish` method attempts to send a message that exceeds maximum size limit allowed by the networking layer, resulting in an unhandled `RuntimeError: MessageTooLarge` This cascades from a TaskGroup exception all the way up to a LifespanFailureError, crashing the runner completely.

Input validation could prevent catastrophic errors.

## To Reproduce

1. Load model (mlx-community/gpt-oss-120b-MXFP4-Q8) with Tensor+RDMA
2. Send request with large context. Tested with examples from LongBench v2
3. Node instance shuts down runner.

## Expected behavior

Message is chunked for processing or, ideally, a 413 error is returned for payloads exceeding length.

## Actual behavior

The message request hangs and the nodes stop responding / drop from cluster.

## Environment Details

- macOS Version: Tahoe 26.2
- EXO Version: 1.0.66
- Hardware:
  - Devices: 3x Mac Mini M4, 64GB (Only 2 used for model instance)
- Interconnection:
  - Thunderbolt ~~4~~ 5 cables
  - 10GbE Ethernet networking

## Additional context
<details> 
  <summary>Stack trace</summary>

```
[ 2026-01-26 16:47:57.952 | INFO     | exo.worker.runner.runner_supervisor:start_task:143 ] Finished task Shutdown(task_id='4ee5039f-69b0-4af7-b363-ca461f57c50e' task_status=<TaskStatus.Pending: 'Pending'> instance_id='49dc288d-4673-4682-8055-a61bf6ae457a' runner_id='6cd5b195-46fb-4461-ad41-5a064fdf8184')
[ 2026-01-26 16:49:12.830 | ERROR    | logging:handle:1681 ] unhandled exception during asyncio.run() shutdown
task: <Task finished name='Task-5' coro=<Lifespan.handle_lifespan() done, defined at hypercorn/asyncio/lifespan.py:43> exception=LifespanFailureError('Lifespan failure in shutdown. \'  + Exception Group Traceback (most recent call last):\n  |   File "anyio/_backends/_asyncio.py", line 2325, in run\n  |   File "asyncio/runners.py", line 118, in run\n  |   File "asyncio/base_events.py", line 725, in run_until_complete\n  |   File "anyio/_backends/_asyncio.py", line 2313, in wrapper\n  |   File "exo/main.py", line 139, in run\n  |   File "anyio/_backends/_asyncio.py", line 781, in __aexit__\n  | ExceptionGroup: unhandled errors in a TaskGroup (1 sub-exception)\n  +-+---------------- 1 ----------------\n    | Exception Group Traceback (most recent call last):\n    |   File "exo/routing/router.py", line 149, in run\n    |   File "anyio/_backends/_asyncio.py", line 781, in __aexit__\n    | ExceptionGroup: unhandled errors in a TaskGroup (1 sub-exception)\n    +-+---------------- 1 ----------------\n      | Traceback (most recent call last):\n      |   File "exo/routing/router.py", line 205, in _networking_publish\n      | RuntimeError: MessageTooLarge\n      +------------------------------------\n\nDuring handling of the above exception, another exception occurred:\n\nTraceback (most recent call last):\n  File "starlette/routing.py", line 701, in lifespan\n  File "hypercorn/asyncio/lifespan.py", line 109, in asgi_receive\n  File "asyncio/queues.py", line 186, in get\nasyncio.exceptions.CancelledError\n\'')>
  + Exception Group Traceback (most recent call last):
  |
  |   File "anyio/_backends/_asyncio.py", line 2325, in run
  |
  |   File "asyncio/runners.py", line 118, in run
  |
  |   File "asyncio/base_events.py", line 725, in run_until_complete
  |
  |   File "anyio/_backends/_asyncio.py", line 2313, in wrapper
  |
  |   File "exo/main.py", line 139, in run
  |
  |   File "anyio/_backends/_asyncio.py", line 781, in __aexit__
  |
  +-+---------------- 1 ----------------
    | Exception Group Traceback (most recent call last):
    |
    |   File "exo/routing/router.py", line 149, in run
    |
    |   File "anyio/_backends/_asyncio.py", line 781, in __aexit__
    |
    +-+---------------- 1 ----------------
      | Traceback (most recent call last):
      |
      |   File "exo/routing/router.py", line 205, in _networking_publish
      |
      | RuntimeError: MessageTooLarge
      +------------------------------------


  File "starlette/routing.py", line 701, in lifespan

  File "hypercorn/asyncio/lifespan.py", line 109, in asgi_receive

  File "asyncio/queues.py", line 186, in get

asyncio.exceptions.CancelledError


  File "__main__.py", line 39, in <module>

  File "exo/main.py", line 271, in main

  File "anyio/_core/_eventloop.py", line 74, in run

  File "anyio/_backends/_asyncio.py", line 2324, in run

  File "asyncio/runners.py", line 62, in __exit__

  File "asyncio/runners.py", line 70, in close

  File "asyncio/runners.py", line 212, in _cancel_all_tasks

  File "asyncio/base_events.py", line 725, in run_until_complete

  File "asyncio/base_events.py", line 687, in run_forever

  File "asyncio/base_events.py", line 2051, in _run_once

  File "asyncio/events.py", line 105, in _run

> File "hypercorn/asyncio/lifespan.py", line 56, in handle_lifespan

  File "hypercorn/app_wrappers.py", line 34, in __call__

  File "fastapi/applications.py", line 1135, in __call__

  File "starlette/applications.py", line 107, in __call__

  File "starlette/middleware/errors.py", line 151, in __call__

  File "starlette/middleware/cors.py", line 77, in __call__

  File "starlette/middleware/exceptions.py", line 49, in __call__

  File "fastapi/middleware/asyncexitstack.py", line 18, in __call__

  File "starlette/routing.py", line 716, in __call__

  File "starlette/routing.py", line 725, in app

  File "starlette/routing.py", line 705, in lifespan

  File "hypercorn/asyncio/lifespan.py", line 121, in asgi_send

hypercorn.utils.LifespanFailureError: Lifespan failure in shutdown. '  + Exception Group Traceback (most recent call last):
  |   File "anyio/_backends/_asyncio.py", line 2325, in run
  |   File "asyncio/runners.py", line 118, in run
  |   File "asyncio/base_events.py", line 725, in run_until_complete
  |   File "anyio/_backends/_asyncio.py", line 2313, in wrapper
  |   File "exo/main.py", line 139, in run
  |   File "anyio/_backends/_asyncio.py", line 781, in __aexit__
  +-+---------------- 1 ----------------
    | Exception Group Traceback (most recent call last):
    |   File "exo/routing/router.py", line 149, in run
    |   File "anyio/_backends/_asyncio.py", line 781, in __aexit__
    +-+---------------- 1 ----------------
      | Traceback (most recent call last):
      |   File "exo/routing/router.py", line 205, in _networking_publish
      | RuntimeError: MessageTooLarge
      +------------------------------------

  File "starlette/routing.py", line 701, in lifespan
  File "hypercorn/asyncio/lifespan.py", line 109, in asgi_receive
  File "asyncio/queues.py", line 186, in get
asyncio.exceptions.CancelledError
```
</details>

AlexCheema:
## Summary
- When a message exceeds gossipsub's 1 MiB limit, `_networking_publish` threw an unhandled `RuntimeError("MessageTooLarge")` which propagated through the TaskGroup and crashed the entire node
- Now catches this error, logs the message size, and drops the oversized message gracefully instead of crashing
- A proper `MessageTooLargeError` exception class in the Rust bindings (similar to existing `AllQueuesFullError`) would be a better long-term fix

Fixes #1296

## Test plan
- [ ] Verify node no longer crashes when a large payload (1M+ chars) is sent via the API
- [ ] Verify normal-sized messages continue to route correctly
- [ ] Verify the error is logged with the topic and payload size

### DIFF
--- a/src/exo/routing/router.py
+++ b/src/exo/routing/router.py
@@ -211,6 +211,14 @@ class Router:
                     pass
                 except AllQueuesFullError:
                     logger.warning(f"All peer queues full, dropping message on {topic}")
+                except RuntimeError as e:
+                    if "MessageTooLarge" in str(e):
+                        logger.error(
+                            f"Message too large for gossipsub on topic {topic} "
+                            f"({len(data)} bytes), dropping message"
+                        )
+                    else:
+                        raise
 
 
 def get_node_id_keypair(


AlexCheema:
## PR #1517 Review: "Handle MessageTooLarge error in router to prevent node crash"

Minimal, surgical fix. 1 file changed, +8/-0. CI all green. Fixes #1296.

---

### Summary

When a message exceeds gossipsub's 1 MiB limit, the Rust bindings throw `RuntimeError("MessageTooLarge")`. This was unhandled in `_networking_publish`, causing the exception to propagate through the TaskGroup and crash the entire node. The fix catches this specific error, logs the topic and payload size, and drops the oversized message gracefully.

### What's Done Well

- **Minimal blast radius**: Only adds a catch clause in the exact location where the error occurs. No refactoring, no unrelated changes.
- **Correct error discrimination**: Catches `RuntimeError` but re-raises if it's not `MessageTooLarge`, avoiding masking unrelated errors.
- **Appropriate log level**: Uses `logger.error` (not warning) since dropping a message is a significant event that operators should investigate.
- **Useful diagnostics**: Logs both the topic name and payload size in bytes, which helps identify what produced the oversized message.
- **Good PR description**: Clearly explains the crash path, the fix, and the long-term improvement (proper `MessageTooLargeError` exception class in Rust bindings).

### One Concern

**String matching on exception message is fragile** — `"MessageTooLarge" in str(e)` depends on the exact error string from the Rust bindings. If the Rust side changes the message (e.g., to "Message too large" or "message_too_large"), this catch silently stops working and the crash returns.

The PR description already acknowledges this: "A proper `MessageTooLargeError` exception class in the Rust bindings (similar to existing `AllQueuesFullError`) would be a better long-term fix."

For now, this is acceptable as a pragmatic fix — the Rust error string is under this project's control. But a follow-up to add a typed exception in `exo_pyo3_bindings` would eliminate this fragility.

### Verdict

LGTM. Clean crash prevention fix. The string-matching concern is acknowledged and appropriate for a short-term fix.

AlexCheema:
## Code Review

Clean, well-scoped fix. LGTM.

The change correctly catches the `RuntimeError("MessageTooLarge")` from the Rust gossipsub layer and drops the message gracefully instead of crashing the node. The pattern is consistent with the existing `AllQueuesFullError` and `NoPeersSubscribedToTopicError` handlers in the same try block.

A few notes:

- **String matching is pragmatic** — `"MessageTooLarge" in str(e)` with `else: raise` is acceptable as a short-term fix since the Rust side controls the error message. As noted in the PR description, a proper `MessageTooLargeError` exception class in the PyO3 bindings (like `AllQueuesFullError`) would be the cleaner long-term solution.

- **Logging at `error` level is appropriate** — this represents data loss (dropped message), so operators should be aware.

- **`len(data)` in the log** is helpful for debugging what triggered the limit.

No issues found. CI 6/6 PASS.

AlexCheema:
Code Review: PR #1517 — Handle MessageTooLarge error in router to prevent node crash
=====================================================================================

Author: AlexCheema
Status: OPEN
Changes: +8 / -0 in src/exo/routing/router.py

Fixes #1296.

Overview
--------
When a message exceeds gossipsub's 1 MiB limit, the Rust bindings' `gossipsub_publish`
raises an unhandled `RuntimeError("MessageTooLarge")` that propagates through the
TaskGroup and crashes the entire node. This PR catches that error, logs the message
size, and drops the oversized message instead of crashing.

Root Cause
----------
In the Rust bindings (networking.rs:196-206), `gossipsub_publish` explicitly converts
`PublishError::NoPeersSubscribedToTopic` and `PublishError::AllQueuesFull` into typed
Python exceptions (`PyNoPeersSubscribedToTopicError`, `PyAllQueuesFullError`). However,
`PublishError::MessageTooLarge` falls through to the generic `.pyerr()` conversion at
line 205, which wraps it as a plain `RuntimeError` with the error's Display string.

On the Python side, `_networking_publish()` catches `NoPeersSubscribedToTopicError` and
`AllQueuesFullError` but had no handler for MessageTooLarge, so the RuntimeError
propagated up through the anyio TaskGroup and cancelled the entire Router — killing
the node.

Correctness
-----------
PASS. The fix is the right approach given the current Rust bindings:

    except RuntimeError as e:
        if "MessageTooLarge" in str(e):
            logger.error(
                f"Message too large for gossipsub on topic {topic} "
                f"({len(data)} bytes), dropping message"
            )
        else:
            raise

1. Only catches RuntimeError (not Exception), minimizing the blast radius.
2. String-checks for "MessageTooLarge" and re-raises any other RuntimeError — so
   genuine bugs still crash loudly.
3. Logs the topic and payload size, giving operators enough info to diagnose.
4. Drops the message (correct — gossipsub fundamentally cannot send it).

The string matching is necessarily fragile since the Rust bindings don't expose a
typed exception. However, this is fine because:
  - The string comes from libp2p's PublishError::MessageTooLarge Display impl, which
    is stable across patch versions.
  - If the string format ever changes, the worst case is a re-raised RuntimeError
    (a crash) — the same behavior as without this fix. So the matching can only help,
    never hurt.

The PR body correctly notes that a proper `PyMessageTooLargeError` exception class
(following the established AllQueuesFullError/NoPeersSubscribedToTopicError pattern)
would be the long-term fix. That's a Rust-side change and reasonable to defer.

Edge Cases
----------
- Multiple unhandled PublishError variants (e.g., TransformFailed, Duplicate,
  SigningError) also fall through to RuntimeError via .pyerr(). These are not caught
  by the "MessageTooLarge" string check, so they still propagate and crash — correct,
  since those indicate actual bugs rather than expected conditions.
- Zero-length or near-limit messages: The `len(data)` in the log message correctly
  reports the serialized byte length, which is what gossipsub measures.
- Concurrent publishes: Each publish is processed sequentially in the async for loop,
  so there's no race between the catch and the log.

Nits
----
None.

Verdict
-------
LGTM. Minimal, correct fix for a real crash scenario. The string-matching approach is
the pragmatic choice given that the Rust bindings don't expose a typed exception for
MessageTooLarge. The fix is safe by construction — any string format change would
simply re-raise the RuntimeError, reverting to pre-fix behavior rather than silently
swallowing an unexpected error.

AlexCheema:
## Code Review: PR #1517

**Verdict: LGTM**

Minimal, correct fix for a real node-crash scenario (#1296).

### Root Cause (verified in Rust source)

In `networking.rs:196-206`, `gossipsub_publish` explicitly converts `PublishError::NoPeersSubscribedToTopic` and `AllQueuesFull` into typed Python exceptions, but `MessageTooLarge` falls through to the generic `.pyerr()` path (lib.rs:74), which wraps it as a plain `RuntimeError` with the Debug-derived string `"MessageTooLarge"`.

On the Python side, `_networking_publish()` runs inside a TaskGroup (router.py:158). The unhandled `RuntimeError` cancels the entire TaskGroup → tears down the Router → crashes the node.

### Fix Analysis

```python
except RuntimeError as e:
    if "MessageTooLarge" in str(e):
        logger.error(f"Message too large for gossipsub on topic {topic} ({len(data)} bytes), dropping message")
    else:
        raise
```

1. **Narrow catch**: `RuntimeError` only, not broad `Exception`
2. **String check + re-raise**: Only catches the specific error. Other `PublishError` variants that also arrive as `RuntimeError` (Duplicate, SigningError, TransformFailed) have different Display strings and correctly re-raise
3. **Drop is correct**: gossipsub fundamentally cannot send an oversized message — no retry or chunking would help
4. **Safe by construction**: if the string format ever changes, the `else: raise` fires and we're back to pre-fix behavior (crash), not silent swallowing

### String Matching Fragility

The `"MessageTooLarge"` string comes from Rust's `#[derive(Debug)]` on `PublishError` (libp2p-gossipsub 0.49.2, error.rs:47: `write!(f, "{self:?}")`). Changing this would require renaming the enum variant — a semver-breaking change. Acceptable for now.

The PR correctly identifies a proper `PyMessageTooLargeError` exception class (following the `AllQueuesFullError` pattern in networking.rs:68) as the long-term fix. Reasonable to defer as a separate Rust-side change.

All 6 CI checks pass.

AlexCheema:
## Review: PR #1517 — Handle MessageTooLarge error in router to prevent node crash

### Summary

Catches `RuntimeError("MessageTooLarge")` from gossipsub publish in `_networking_publish()`, logs the topic and payload size, and drops the oversized message instead of letting it propagate through the TaskGroup and crash the node.

### Analysis

**The fix is correct and prevents a real crash.** In the Rust bindings, `gossipsub_publish` maps unhandled `PublishError` variants to `PyRuntimeError` via `.pyerr()` (`lib.rs:75`), which converts the error with `.to_string()`. The `NoPeersSubscribedToTopic` and `AllQueuesFull` variants have dedicated Python exception classes, but `MessageTooLarge` does not — so it surfaces as a generic `RuntimeError`.

**The re-raise for non-`MessageTooLarge` RuntimeErrors is correct.** The `else: raise` ensures other unexpected `RuntimeError`s (e.g., `InsufficientPeers`, `TransformFailed`, `SigningError`, `Duplicate` from gossipsub's `PublishError` enum) are not silently swallowed.

**String matching is acknowledged as fragile.** The PR description notes a proper `MessageTooLargeError` exception class would be better. The current approach depends on libp2p's `PublishError::MessageTooLarge` producing "MessageTooLarge" in its `Display` implementation — this is stable in practice but technically an upstream implementation detail.

**Logging is good.** `len(data)` correctly gives the byte count (the `data` parameter is `bytes` per the pyi stub), and including the topic helps with debugging.

### Minor observation

**Silent message dropping could be surprising to callers.** The oversized message is silently dropped from the perspective of whatever component sent it — no error propagates back through the channel. For the current event-sourcing architecture (fire-and-forget pub/sub), this is fine. But if a caller ever needs to know a message was dropped (e.g., for retry or user feedback), this would need to change. Not a concern for the current use case.

### Verdict

Clean, targeted crash prevention. The string-matching approach is pragmatic for an interim fix, and the re-raise for other `RuntimeError`s is correctly scoped. LGTM.

AlexCheema:
## Code Review: PR #1517 — Handle MessageTooLarge error in router to prevent node crash

### Summary

Catches `RuntimeError("MessageTooLarge")` from gossipsub publish in `_networking_publish()`, logs the topic and payload size, and drops the oversized message instead of letting it crash the node via the TaskGroup.

### Analysis

**The fix prevents a real crash path.** In the Rust bindings (`networking.rs:196-206`), `gossipsub_publish` handles `NoPeersSubscribedToTopic` and `AllQueuesFull` with dedicated Python exception classes, but all other `PublishError` variants (including `MessageTooLarge`) fall through to `.pyerr()` (`lib.rs:74-75`):

```rust
fn pyerr(self) -> PyResult<T> {
    self.map_err(|e| PyRuntimeError::new_err(e.to_string()))
}
```

This produces a generic `RuntimeError`. Since libp2p's `PublishError` delegates `Display` to `Debug`, `to_string()` returns `"MessageTooLarge"` — so the `"MessageTooLarge" in str(e)` check works correctly.

**The `else: raise` is essential and correctly scoped.** Other `PublishError` variants that reach the Python side as `RuntimeError` (e.g., `InsufficientPeers`, `TransformFailed`, `Duplicate`) are not swallowed — they propagate as before.

**Logging is appropriate.** Logs `len(data)` (byte count) and the topic at `error` level — sufficient for debugging without leaking message content.

### Concern: String matching fragility

The check `"MessageTooLarge" in str(e)` depends on:
1. libp2p's `PublishError::MessageTooLarge` display format (currently delegates `Display` → `Debug`, producing the variant name)
2. The `.pyerr()` conversion using `.to_string()`

If libp2p ever switches to `thiserror` with human-readable messages (e.g., `"data must be smaller than max_transmit_size"`), this match would silently stop catching the error and the crash would return. The PR acknowledges this and suggests a proper `MessageTooLargeError` exception class as the long-term fix.

A small improvement for robustness: also match on a likely alternative string, e.g.:
```python
if "MessageTooLarge" in str(e) or "max_transmit_size" in str(e):
```
But this is minor — the proper fix is adding a typed exception on the Rust side, following the `AllQueuesFullError` pattern.

### Minor observation

Oversized messages are silently dropped from the sender's perspective — no error propagates back through the channel. For the current fire-and-forget pub/sub architecture this is fine, but it means callers can't detect or retry. Not a concern for the current use case.

### Verdict

**LGTM.** Clean crash prevention with correctly scoped error handling. The string-matching fragility is acknowledged and acceptable as an interim fix.

AlexCheema:
## Code Review: PR #1517 — Handle MessageTooLarge error in router to prevent node crash

### Summary
Prevents a node crash when a gossipsub message exceeds the 1 MiB limit. Instead of letting the `RuntimeError("MessageTooLarge")` propagate through the TaskGroup and crash everything, it's now caught and logged.

### Issues

**1. String-matching on exception messages is fragile** (`router.py`)
```python
except RuntimeError as e:
    if "MessageTooLarge" in str(e):
```
This is the main concern. If the Rust side changes the error message string (e.g., to "Message too large" or "message_too_large"), this catch silently stops working and the node crashes again. The PR description acknowledges this:

> A proper `MessageTooLargeError` exception class in the Rust bindings (similar to existing `AllQueuesFullError`) would be a better long-term fix

I'd strongly recommend either:
- **Short term**: Adding a `# TODO: Replace with MessageTooLargeError when added to Rust bindings` comment so this doesn't get lost
- **Long term**: Adding the proper exception class in `exo_pyo3_bindings` (just like `AllQueuesFullError` already exists)

**2. Log level: `logger.error` vs `logger.warning`**
The `AllQueuesFullError` case uses `logger.warning`. A too-large message is arguably a similar "expected operational condition" rather than a true error. Consider using `logger.warning` for consistency, or if you want to keep `logger.error`, add a comment explaining why this is more severe than full queues.

**3. No test coverage**
There are no tests for this change. Given that it's a string-match on a RuntimeError, a test that verifies the behavior would help catch regressions, especially if the Rust bindings change. Even a unit test that mocks the publish call to raise `RuntimeError("MessageTooLarge")` would be valuable.

### Verdict
The fix is correct and prevents a real crash. The string-matching approach is pragmatic but fragile. Recommend at minimum adding a TODO comment about migrating to a proper exception class, and consider adding a test.

LGTM with the caveats above — this is better than crashing.

AlexCheema:
## Code Review — PR #1517: Handle MessageTooLarge error in router to prevent node crash

**CI status**: 6/6 checks passed (aarch64-darwin, x86_64-linux, aarch64-linux x2 each)

### Overview

This PR fixes issue #1296, where sending a message exceeding gossipsub's 1 MiB limit causes an unhandled `RuntimeError("MessageTooLarge")` that propagates through the TaskGroup in `_networking_publish`, tears down the Router, and crashes the entire node. The fix adds a targeted `except RuntimeError` clause that string-checks for `"MessageTooLarge"`, logs the topic and payload size, and drops the message.

**Change**: +8/-0 in `src/exo/routing/router.py`

### Root Cause Verification

Confirmed by reading the Rust source. In `networking.rs:196-206`, `gossipsub_publish` explicitly converts `PublishError::NoPeersSubscribedToTopic` and `PublishError::AllQueuesFull` into typed Python exceptions, but `PublishError::MessageTooLarge` falls through to the generic `.pyerr()` path (`lib.rs:74-75`):

```rust
fn pyerr(self) -> PyResult<T> {
    self.map_err(|e| PyRuntimeError::new_err(e.to_string()))
}
```

This wraps it as a plain `RuntimeError` with the `to_string()` output of the `PublishError` variant.

### Analysis

**The fix is correct.** The error handling is well-scoped:

1. Catches only `RuntimeError`, not broad `Exception` -- minimal blast radius.
2. String-checks for `"MessageTooLarge"` and re-raises anything else via `else: raise` -- other `RuntimeError`s (from `PublishError::Duplicate`, `TransformFailed`, etc.) still propagate correctly.
3. Dropping the message is the only correct action -- gossipsub fundamentally cannot send an oversized message; retrying or chunking would not help at this layer.
4. Logging includes both topic name and `len(data)` byte count, giving operators what they need to diagnose the source.
5. The fix is **safe by construction**: if the Rust-side error string ever changes, the `else: raise` fires and we get pre-fix behavior (crash), rather than silently swallowing an unexpected error.

### Issues

**1. String matching on exception message is inherently fragile** (Significant -- but acknowledged)

```python
except RuntimeError as e:
    if "MessageTooLarge" in str(e):
```

The `"MessageTooLarge"` string comes from `PublishError`'s `Display` implementation (which delegates to `Debug` in libp2p-gossipsub 0.49.2, producing the variant name). If libp2p ever migrates to `thiserror` with human-readable messages, or the variant is renamed, this match silently stops catching and the crash returns.

The PR description already acknowledges this and correctly identifies the long-term fix: adding a proper `PyMessageTooLargeError` exception class in the Rust bindings, following the exact pattern already established for `PyAllQueuesFullError` and `PyNoPeersSubscribedToTopicError` at `networking.rs:32-100`. That Rust-side change would be roughly 30 lines and would eliminate the fragility entirely.

**Suggestion**: Add a TODO comment at the catch site so this does not get lost:

```python
except RuntimeError as e:
    # TODO: Replace with MessageTooLargeError when added to Rust bindings
    if "MessageTooLarge" in str(e):
```

**2. Log level inconsistency** (Minor)

The `AllQueuesFullError` case uses `logger.warning`, but this uses `logger.error`. Both represent "dropping a message due to a transport limitation." Using `logger.error` is defensible (it indicates data loss), but the inconsistency is worth a conscious choice. If `error` is intentional, a brief comment explaining why would help.

### What's Good

- **Minimal and surgical**: 8 lines, 1 file, no refactoring or unrelated changes.
- **Follows existing patterns**: The catch clause mirrors the structure of the `NoPeersSubscribedToTopicError` and `AllQueuesFullError` handlers directly above it.
- **Good PR description**: Clearly explains the crash path, the fix, and the long-term improvement. Links to issue #1296 which has a full stack trace confirming the exact crash path at `router.py:205` inside `_networking_publish`.
- **Correct diagnosis**: The stack trace in #1296 shows exactly `RuntimeError: MessageTooLarge` at `router.py:205`, which is precisely where this fix catches it.

### Verdict

LGTM. This is a correct, well-scoped crash prevention fix for a real production issue. The string-matching fragility is the right tradeoff for a short-term fix -- the `else: raise` makes it fail-safe, and the long-term fix (typed Rust exception) is clearly identified. I would suggest adding a TODO comment at the catch site so the follow-up does not get lost, but that is not blocking.

Evanev7:
workin on it myself atm

PR #1521, opened 2026-02-17
fix: --no-downloads no longer blocks model loading

### LINKED ISSUE #1510: Offline/air-gapped mode
### Problem

There is currently no reliable way to run exo in an air-gapped or offline cluster where models are pre-staged on local disks (or a shared read-only NFS mount). The two existing partial mechanisms — `--no-downloads` and `HF_HUB_OFFLINE=1` — each cover only a subset of the problem and conflict with each other.

#### 1. `--no-downloads` breaks model loading on worker nodes

When `--no-downloads` is passed, `DownloadCoordinator` is not created ([main.py#L63-74](https://github.com/exo-explore/exo/blob/main/src/exo/main.py#L63-L74)). This means no `NodeDownloadProgress` events are ever emitted for the node, so `state.downloads` for that node remains empty.

In `plan.py`, the planning functions are evaluated via short-circuit OR:

```python
return (
    _kill_runner(...)
    or _create_runner(...)
    or _model_needs_download(...)   # ← blocks here
    or _init_distributed_backend(...)
    or _load_model(...)             # ← never reached
    or _ready_to_warmup(...)
    or _pending_tasks(...)
)
```

`_model_needs_download()` checks if the runner is `RunnerIdle` and the model has no `DownloadOngoing`, `DownloadCompleted`, or `DownloadFailed` status. With no coordinator running, the download status for the node is always empty, so this function perpetually returns `DownloadModel` — which short-circuits `_load_model()` and prevents the model from ever being loaded into memory.

The `DownloadModel` task is sent to the `DOWNLOAD_COMMANDS` topic, but no coordinator is listening. The `KeyedBackoff` throttles retries, but every `plan_step()` iteration (every 100ms) still evaluates `_model_needs_download()` first, blocking the pipeline indefinitely.

**Reproduction:**
1. Pre-stage a model on disk (e.g., `~/.exo/models/mlx-community--Qwen3-0.6B-4bit/`)
2. Start exo with `--no-downloads`
3. Place an instance via API or dashboard
4. Observe: runner stays `RunnerIdle` forever, model never loads

#### 2. `HF_HUB_OFFLINE=1` does not cover exo's custom HTTP code

`HF_HUB_OFFLINE=1` is a standard `huggingface_hub` library environment variable. However, exo's download pipeline primarily uses raw `aiohttp` for HTTP operations, not the `huggingface_hub` library:

| Code path | Uses `huggingface_hub`? | Blocked by `HF_HUB_OFFLINE=1`? |
|---|---|---|
| `fetch_file_list_with_cache()` → `_fetch_file_list()` | No (raw aiohttp) | **No** |
| `download_file_with_retry()` → `_download_file()` | No (raw aiohttp) | **No** |
| `file_meta()` (HTTP HEAD for verification) | No (raw aiohttp) | **No** |
| `get_weight_map()` → `snapshot_download()` | Yes | Yes |
| `fetch_safetensors_size()` → `model_info()` | Yes | Yes |
| `search_models()` → `list_models()` | Yes | Yes |

When `HF_HUB_OFFLINE=1` is set but the node has no internet, the aiohttp-based calls will attempt connections, time out (30-60 seconds each), retry, and eventually fail. This adds significant latency to startup even when all model files are already present on disk.

#### 3. `internet_connection` auto-detection is partial

The `DownloadCoordinator` pings `1.1.1.1:443` at startup and every 10 seconds. If unreachable, it sets `internet_connection = False`, which propagates `skip_internet=True` to `fetch_file_list_with_cache()`. This correctly prevents the HF tree API call.

However, `skip_internet` is **not** propagated to `download_file_with_retry()` / `file_meta()`. If the download path is triggered (e.g., file size mismatch), HTTP HEAD/GET requests will still be attempted and will time out.

### Use Case

Running exo on an air-gapped Apple Silicon cluster where:
- Models are pre-staged to local NVMe disks via `rsync` or similar tools
- Or models are on a shared read-only NFS mount from a TB5-attached NVMe enclosure
- The cluster has no internet access (or internet access is unreliable/unwanted)
- exo auto-starts at boot via `launchd` and should be ready to serve inference requests immediately

This is a common setup for semi-professional and enterprise deployments where model management is handled by external tooling (Ansible, scripts, NFS infrastructure) rather than by the inference runtime itself.

### Proposed Solution

A unified `--offline` flag (or `EXO_OFFLINE=1` environment variable) that:

1. **Sets `internet_connection = False` immediately** without pinging `1.1.1.1` (avoids 3-second timeout on air-gapped networks)
2. **Propagates `skip_internet` to all download functions**, including `download_file_with_retry()` and `file_meta()`
3. **In `_download_file()`**: if the file exists on disk, returns it immediately without HTTP HEAD verification against HuggingFace
4. **In `_download_file()`**: if the file does not exist, raises an error instead of attempting download
5. **In `_model_needs_download()`**: checks actual file existence on disk instead of relying solely on `DownloadCompleted` events from the coordinator. If all required files exist, allows the pipeline to proceed to `_load_model()`

Alternatively, the `DownloadCoordinator` could be modified to emit synthetic `DownloadCompleted` events based on file existence checks (local `stat()` only, no network), which would unblock `_load_model()` without changes to the planning logic.

### Related Issues

- #1038 — EXO macOS app cannot start without internet connection
- #1257 — Why should all nodes download the model directly from Hugging Face?
- #1081 — Allow the .exo or downloads folder to be mapped to different location, including network shares
- #721 — Feature Request: Add Rsync
- #882 — Xet storage to replace git LFS

### Environment

- exo version: main branch (as of Feb 2026)
- OS: macOS (Apple Silicon)
- Hardware: Multi-node Apple Silicon cluster with Thunderbolt interconnect
- Network: air-gapped (no internet), models pre-staged on local disks via external tooling

AlexCheema:
## Summary
- When `--no-downloads` is passed, `plan()` now skips download checks so pre-staged models can be loaded without `DownloadCompleted` events
- Adds `no_downloads` flag through `Worker` → `plan()` → `_load_model()`, skipping `_model_needs_download()` and the download completeness check
- Preserves the flag across Worker recreation during master re-election

Fixes #1510

## Test plan
- [x] New test `test_plan_loads_model_with_no_downloads_flag` verifies `LoadModel` is returned with empty download status when `no_downloads=True`
- [x] All existing plan tests pass (no regression)
- [x] basedpyright passes with 0 errors
- [x] ruff check passes

### DIFF
--- a/src/exo/main.py
+++ b/src/exo/main.py
@@ -94,6 +94,7 @@ class Node:
                 command_sender=router.sender(topics.COMMANDS),
                 download_command_sender=router.sender(topics.DOWNLOAD_COMMANDS),
                 event_index_counter=event_index_counter,
+                no_downloads=args.no_downloads,
             )
         else:
             worker = None
@@ -225,6 +226,7 @@ class Node:
                         )
                         self._tg.start_soon(self.download_coordinator.run)
                     if self.worker:
+                        no_downloads = self.worker.no_downloads
                         self.worker.shutdown()
                         # TODO: add profiling etc to resource monitor
                         self.worker = Worker(
@@ -239,6 +241,7 @@ class Node:
                                 topics.DOWNLOAD_COMMANDS
                             ),
                             event_index_counter=self.event_index_counter,
+                            no_downloads=no_downloads,
                         )
                         self._tg.start_soon(self.worker.run)
                     if self.api:
--- a/src/exo/worker/main.py
+++ b/src/exo/worker/main.py
@@ -65,6 +65,7 @@ class Worker:
         command_sender: Sender[ForwarderCommand],
         download_command_sender: Sender[ForwarderDownloadCommand],
         event_index_counter: Iterator[int],
+        no_downloads: bool = False,
     ):
         self.node_id: NodeId = node_id
         self.session_id: SessionId = session_id
@@ -74,6 +75,7 @@ class Worker:
         self.event_index_counter = event_index_counter
         self.command_sender = command_sender
         self.download_command_sender = download_command_sender
+        self.no_downloads = no_downloads
         self.event_buffer = OrderedBuffer[Event]()
         self.out_for_delivery: dict[EventId, ForwarderEvent] = {}
 
@@ -182,6 +184,7 @@ class Worker:
                 self.state.tasks,
                 self.input_chunk_buffer,
                 self.input_chunk_counts,
+                no_downloads=self.no_downloads,
             )
             if task is None:
                 continue
--- a/src/exo/worker/plan.py
+++ b/src/exo/worker/plan.py
@@ -51,15 +51,20 @@ def plan(
     tasks: Mapping[TaskId, Task],
     input_chunk_buffer: Mapping[CommandId, dict[int, str]] | None = None,
     input_chunk_counts: Mapping[CommandId, int] | None = None,
+    no_downloads: bool = False,
 ) -> Task | None:
     # Python short circuiting OR logic should evaluate these sequentially.
     return (
         _cancel_tasks(runners, tasks)
         or _kill_runner(runners, all_runners, instances)
         or _create_runner(node_id, runners, instances)
-        or _model_needs_download(node_id, runners, global_download_status)
+        or (
+            None
+            if no_downloads
+            else _model_needs_download(node_id, runners, global_download_status)
+        )
         or _init_distributed_backend(runners, all_runners)
-        or _load_model(runners, all_runners, global_download_status)
+        or _load_model(runners, all_runners, global_download_status, no_downloads)
         or _ready_to_warmup(runners, all_runners)
         or _pending_tasks(runners, tasks, all_runners, input_chunk_buffer or {})
     )
@@ -192,22 +197,25 @@ def _load_model(
     runners: Mapping[RunnerId, RunnerSupervisor],
     all_runners: Mapping[RunnerId, RunnerStatus],
     global_download_status: Mapping[NodeId, Sequence[DownloadProgress]],
+    no_downloads: bool = False,
 ) -> LoadModel | None:
     for runner in runners.values():
         instance = runner.bound_instance.instance
         shard_assignments = instance.shard_assignments
 
-        all_local_downloads_complete = all(
-            nid in global_download_status
-            and any(
-                isinstance(dp, DownloadCompleted)
-                and dp.shard_metadata.model_card.model_id == shard_assignments.model_id
-                for dp in global_download_status[nid]
+        if not no_downloads:
+            all_local_downloads_complete = all(
+                nid in global_download_status
+                and any(
+                    isinstance(dp, DownloadCompleted)
+                    and dp.shard_metadata.model_card.model_id
+                    == shard_assignments.model_id
+                    for dp in global_download_status[nid]
+                )
+                for nid in shard_assignments.node_to_runner
             )
-            for nid in shard_assignments.node_to_runner
-        )
-        if not all_local_downloads_complete:
-            continue
+            if not all_local_downloads_complete:
+                continue
 
         is_single_node_instance = len(instance.shard_assignments.runner_to_shard) == 1
         if is_single_node_instance and isinstance(runner.status, RunnerIdle):
--- a/src/exo/worker/tests/unittests/test_plan/test_download_and_loading.py
+++ b/src/exo/worker/tests/unittests/test_plan/test_download_and_loading.py
@@ -157,6 +157,41 @@ def test_plan_does_not_request_download_when_shard_already_downloaded():
     assert not isinstance(result, plan_mod.DownloadModel)
 
 
+def test_plan_loads_model_with_no_downloads_flag():
+    """
+    When no_downloads=True, plan() should skip download checks and proceed
+    to load the model even with empty global_download_status.
+    """
+    shard = get_pipeline_shard_metadata(model_id=MODEL_A_ID, device_rank=0)
+    instance = get_mlx_ring_instance(
+        instance_id=INSTANCE_1_ID,
+        model_id=MODEL_A_ID,
+        node_to_runner={NODE_A: RUNNER_1_ID},
+        runner_to_shard={RUNNER_1_ID: shard},
+    )
+    bound_instance = BoundInstance(
+        instance=instance, bound_runner_id=RUNNER_1_ID, bound_node_id=NODE_A
+    )
+    runner = FakeRunnerSupervisor(bound_instance=bound_instance, status=RunnerIdle())
+
+    runners = {RUNNER_1_ID: runner}
+    instances = {INSTANCE_1_ID: instance}
+    all_runners = {RUNNER_1_ID: RunnerIdle()}
+
+    result = plan_mod.plan(
+        node_id=NODE_A,
+        runners=runners,  # type: ignore
+        global_download_status={},
+        instances=instances,
+        all_runners=all_runners,
+        tasks={},
+        no_downloads=True,
+    )
+
+    assert isinstance(result, LoadModel)
+    assert result.instance_id == INSTANCE_1_ID
+
+
 def test_plan_does_not_load_model_until_all_shards_downloaded_globally():
     """
     LoadModel should not be emitted while some shards are still missing from


AlexCheema:
## Review: PR #1521 — fix: --no-downloads no longer blocks model loading

### Summary

When `--no-downloads` is passed, `plan()` was still gating model loading on `DownloadCompleted` events that would never arrive (since downloads are disabled). This fix threads a `no_downloads` flag through `Worker` → `plan()` → `_load_model()`, skipping both the download-trigger step and the download-completeness check so pre-staged models can load immediately.

### Analysis

**The fix touches two correct points in the plan chain:**

1. **`_model_needs_download` is skipped** — When `no_downloads=True`, the `None if no_downloads else _model_needs_download(...)` evaluates to `None` (falsy), so the OR chain continues. This prevents emitting `DownloadModel` tasks that would be ignored anyway.

2. **`_load_model` skips download completeness check** — The `all_local_downloads_complete` gate (which checks for `DownloadCompleted` events per node) is bypassed entirely. Without this, models would never load because no download events are emitted.

**Worker recreation preserves the flag** — In `_elect_loop`, `no_downloads` is saved before shutting down the old worker and passed to the new one. This is correct and ensures the flag survives master re-election.

**Failure mode is reasonable** — If `no_downloads=True` but the model isn't actually on disk, `_load_model` will return a `LoadModel` task, the runner will attempt to load from disk, and fail with `RunnerFailed`. This is the expected behavior when the user explicitly opts out of downloads — they're responsible for pre-staging.

**Multi-node case** — The download completeness check normally ensures ALL nodes have the model before any starts loading. With `no_downloads=True`, this cross-node check is skipped entirely. Each node will independently attempt to load from local disk. If one node doesn't have the model pre-staged, it will fail individually while others succeed. This is acceptable — the user should pre-stage on all nodes when using `--no-downloads`.

### Minor style note

The `None if no_downloads else _model_needs_download(...)` pattern in the OR chain works correctly but reads a bit unusually. An alternative would be `_model_needs_download(...) if not no_downloads else None`, which reads more naturally left-to-right, but this is a nit.

### Test coverage

The new test `test_plan_loads_model_with_no_downloads_flag` verifies the core scenario: single-node instance with empty `global_download_status` and `no_downloads=True` returns a `LoadModel` task. Clean and sufficient.

### Verdict

Correct, minimal fix for a clear bug. The flag is threaded cleanly through the stack and preserved across worker recreation. LGTM.

AlexCheema:
## Code Review: PR #1521 — fix: --no-downloads no longer blocks model loading

### Summary

When `--no-downloads` is passed, the `DownloadCoordinator` is disabled (no `DownloadCompleted` events emitted), but `plan()` still gated model loading on those events — so pre-staged models could never load. This fix threads a `no_downloads` flag through `Worker` → `plan()`, skipping both the download-trigger step and the download-completeness check.

### Analysis

**Two correct intervention points in the plan chain:**

1. **`_model_needs_download` is skipped** via `(None if no_downloads else _model_needs_download(...))`. `None` is falsy so the OR chain continues past it. This prevents emitting `DownloadModel` tasks that would have no coordinator to process them.

2. **`_load_model` bypasses the download completeness gate** — The `all_local_downloads_complete` check (which verifies `DownloadCompleted` events exist for every node in the shard assignment) is wrapped in `if not no_downloads`. Without this bypass, `_load_model` would `continue` past every runner since the global download status is empty.

**Worker recreation correctly preserves the flag** (`main.py:228`). `no_downloads` is read from the old worker *before* `shutdown()` is called and passed to the new one, ensuring the flag survives master re-election cycles.

**Failure mode is reasonable.** If `--no-downloads` is set but the model isn't on disk, `LoadModel` is emitted anyway, the runner attempts to load, and fails with `RunnerFailed`. This is the expected contract — the user explicitly opted out of downloads and is responsible for pre-staging.

**Multi-node consideration.** The completeness check normally gates loading until ALL nodes in the shard assignment report `DownloadCompleted`. With `no_downloads=True`, this cross-node gate is skipped entirely — each node will independently attempt to load from local disk. If one node doesn't have the model, it fails individually. This is acceptable: `--no-downloads` is a per-node flag, and in a multi-node setup the user would set it on all nodes with pre-staged models.

### Test

`test_plan_loads_model_with_no_downloads_flag` directly tests the core scenario: single-node instance, empty `global_download_status`, `no_downloads=True` → `LoadModel` returned. Clean and sufficient. Could optionally add a test verifying `_model_needs_download` is also skipped (i.e., no `DownloadModel` task emitted), but the existing test implicitly covers this since `_model_needs_download` comes before `_load_model` in the chain — if it weren't skipped, it would return `DownloadModel` instead of `LoadModel`.

### Verdict

**LGTM.** Clean, correct fix for a clear bug. The flag is threaded through the stack minimally and preserved across worker recreation.

AlexCheema:
## Code Review: PR #1521 — fix: --no-downloads no longer blocks model loading

### Summary
Good fix. When `--no-downloads` is set, the planner was still waiting for `DownloadCompleted` events before issuing `LoadModel` tasks, which blocked pre-staged models from loading. This PR threads a `no_downloads` flag through `Worker` → `plan()` → `_load_model()` to skip download checks.

### Design Review

**Threading the flag through correctly:**
- `Args.no_downloads` → `Worker.__init__` → `plan()` → `_load_model()` ✅
- Preserved across worker recreation during master re-election (`_elect_loop`) ✅
- Default `no_downloads=False` maintains backward compatibility ✅

### Issues

**1. `plan()` short-circuit chain readability** (`plan.py`)
```python
or (
    None
    if no_downloads
    else _model_needs_download(node_id, runners, global_download_status)
)
```
This ternary-in-a-chain is a bit hard to parse on first read. A clearer alternative:
```python
or (not no_downloads and _model_needs_download(node_id, runners, global_download_status))
```
Or even extracting to a local variable. Minor style nit though.

**2. What happens when the model files don't actually exist on disk?**
When `no_downloads=True`, we skip both `_model_needs_download()` and the download completeness check in `_load_model()`. If the model files aren't actually pre-staged on disk, the `LoadModel` task will be issued but presumably fail at the runner level. Is there clear error reporting for this case? It would be good to ensure the runner surfaces a meaningful error rather than a cryptic file-not-found.

### Test Coverage
- New test `test_plan_loads_model_with_no_downloads_flag` ✅ — verifies `LoadModel` is returned with empty download status when `no_downloads=True`
- Good that it checks the exact instance and task type

### Verdict
Clean, well-scoped fix. The flag is threaded correctly and preserved across re-elections. Test coverage is solid. Only minor concern is error handling when model files are actually missing despite `--no-downloads` being set, but that's arguably a user error.

LGTM.

AlexCheema:
## Code Review -- PR #1521: fix: --no-downloads no longer blocks model loading

**CI status:** All 6 checks passed (aarch64-darwin, x86_64-linux, aarch64-linux on both pipeline runs).

---

### Overview

When `--no-downloads` is passed, the `DownloadCoordinator` is never created, so no `DownloadCompleted` events are emitted. But `plan()` was still gating model loading on those events, meaning pre-staged models could never load. This PR threads a `no_downloads: bool` flag through `Worker` -> `plan()` -> `_load_model()`, skipping the download-trigger step and the download-completeness gate.

The fix touches 4 files across 3 layers: `Node.create` / `_elect_loop` (flag origination and preservation), `Worker.__init__` / `plan_step` (flag storage and forwarding), and `plan()` / `_load_model()` (behavioral changes).

---

### Significant Issues

None found. The logic is correct across both single-node and multi-node paths.

For the multi-node `--no-downloads` case, I verified the plan chain ordering is sound: `_model_needs_download` is skipped, `_init_distributed_backend` picks up `RunnerIdle` -> emits `ConnectToGroup` -> runner transitions to `RunnerConnected`, then `_load_model` fires with the download check skipped. No ordering issues.

---

### Minor Issues

**1. Ternary expression in the OR chain reads somewhat awkwardly** (`plan.py:54-58`)

```python
or (
    None
    if no_downloads
    else _model_needs_download(node_id, runners, global_download_status)
)
```

`None` is falsy, so this works correctly in the short-circuit chain. However, the `None if X else expr` pattern requires the reader to mentally invert the condition. A slightly more idiomatic alternative:

```python
or (not no_downloads and _model_needs_download(node_id, runners, global_download_status))
```

This reads left-to-right as "if downloads are enabled, check if model needs download." Both are correct; purely a readability nit.

**2. Test could additionally verify that `DownloadModel` is NOT returned**

The new test `test_plan_loads_model_with_no_downloads_flag` correctly verifies that `LoadModel` is returned. Since `_model_needs_download` comes before `_load_model` in the chain, if the skip weren't working, the test would get `DownloadModel` instead -- so this is implicitly covered. But an explicit assertion like `assert not isinstance(result, plan_mod.DownloadModel)` would make the intent doubly clear.

**3. Default parameter value on `_load_model` may be unnecessary** (`plan.py:200`)

```python
def _load_model(
    ...,
    no_downloads: bool = False,
) -> LoadModel | None:
```

`_load_model` is only called from `plan()`, which always passes `no_downloads` explicitly. The `= False` default means it is possible to call `_load_model` without the flag and silently get the old behavior, which could mask a bug if a new callsite is added. Making it a required parameter (`no_downloads: bool` with no default) would be marginally safer. Very minor since the function is internal.

---

### What's Good

- **Minimal, well-scoped fix.** Only touches the two exact points in the plan chain that need to change: `_model_needs_download` (skip entirely) and `_load_model` (skip completeness gate). No unnecessary changes.
- **Flag preservation across re-election.** The `_elect_loop` correctly reads `no_downloads` from the old worker before `shutdown()` and passes it to the new one. This is easy to miss and was handled correctly.
- **Clean test.** `test_plan_loads_model_with_no_downloads_flag` directly tests the core scenario with empty `global_download_status` and `no_downloads=True`, confirming `LoadModel` is returned. The test follows existing patterns in the file.
- **Backward compatible.** `no_downloads=False` is the default everywhere, so existing behavior is unchanged.
- **Failure mode is explicit.** If `--no-downloads` is set but the model is not on disk, the runner will attempt to load and fail with `RunnerFailed` -- the correct contract for a flag that says "I have pre-staged the model."

---

### Verdict

Clean, correct fix for a clear bug. The flag is threaded through the stack minimally and preserved across worker recreation. All CI checks pass. The issues above are minor style/robustness nits. LGTM.

rltakashige:
Haven't read the PR much, but my understanding was that this was left as is intentionally by @JakeHillion

PR #1525, opened 2026-02-18
feat: add unified --offline mode for air-gapped clusters

### LINKED ISSUE #1510: Offline/air-gapped mode

vskiwi:
## Motivation

Closes #1510

There is currently no reliable way to run exo on an air-gapped or offline cluster where models are pre-staged on local disks. The two existing mechanisms — `--no-downloads` and `HF_HUB_OFFLINE=1` — each cover only a subset of the problem:

1. **`--no-downloads` blocks model loading**: When passed, `DownloadCoordinator` is not created. No `NodeDownloadProgress` events are ever emitted, so `_model_needs_download()` in `plan.py` perpetually returns `DownloadModel`, short-circuiting `_load_model()` and preventing the model from ever being loaded.

2. **`HF_HUB_OFFLINE=1` doesn't cover exo's aiohttp code**: exo's download pipeline primarily uses raw `aiohttp` for HTTP operations (file list fetching, file downloads, HEAD verification), not the `huggingface_hub` library. These calls will attempt connections and time out on air-gapped networks.

3. **`skip_internet` is not propagated to `download_file_with_retry()`**: Even when `internet_connection = False`, the `_download_file()` function still makes HTTP HEAD calls via `file_meta()` to verify local files and unconditionally attempts downloads for missing files.

### `src/exo/main.py`
- Add `--offline` flag to `Args` with env var detection (`EXO_OFFLINE=1`, `HF_HUB_OFFLINE=1`)
- Pass `offline` to `DownloadCoordinator` at creation and re-creation (election loop)

### `src/exo/download/coordinator.py`
- Add `offline: bool = False` field
- In offline mode: set `internet_connection = False` immediately in `__post_init__`, skip `_test_internet_connection()` ping (avoids 3s timeout), skip `_check_internet_connection` periodic loop
- In `_start_download()`: if model is not fully available locally, emit `DownloadFailed` with clear message instead of starting a download task

### `src/exo/download/download_utils.py`
- Add `skip_internet: bool` parameter to `download_file_with_retry()` and `_download_file()`
- When `skip_internet=True` in `_download_file()`: return local file immediately without HTTP HEAD verification; raise `FileNotFoundError` for missing files
- Propagate `skip_internet` from `download_shard()` to `download_file_with_retry()`

### `src/exo/download/tests/test_offline_mode.py` (new)
- 8 tests covering `_download_file`, `download_file_with_retry`, and `fetch_file_list_with_cache` in offline mode

## Why It Works

Unlike `--no-downloads` which disables `DownloadCoordinator` entirely, `--offline` keeps the coordinator running in a restricted mode. The existing `_emit_existing_download_progress()` disk scanner still runs every 60 seconds, emitting `DownloadCompleted` events for pre-staged models. These events flow through the event-sourcing pipeline and populate `state.downloads`, which unblocks `_model_needs_download()` in `plan.py` — no changes to the planning logic required.

```
--offline flag
  → DownloadCoordinator (offline mode)
    → Skip 1.1.1.1 ping, internet_connection = False
    → _emit_existing_download_progress scans disk
      → Emits DownloadCompleted for pre-staged models
        → _model_needs_download sees DownloadCompleted
          → _load_model proceeds normally
```

### Automated Testing
- `ruff check` — passes
- 8 new tests in `test_offline_mode.py` — all pass
- 11 existing download tests in `test_download_verification.py` — all pass (no regressions)

### Manual Testing
1. Pre-stage a model on disk (e.g., `~/.exo/models/mlx-community--Qwen3-0.6B-4bit/`)
2. Start exo with `--offline` (or `EXO_OFFLINE=1`)
3. Place an instance via API or dashboard
4. Verify: model loads into memory and inference works without any network calls

### Environment
- macOS (Apple Silicon), multi-node cluster with Thunderbolt interconnect
- Models pre-staged via rsync / NFS mount

Made with [Cursor](https://cursor.com)

### DIFF
--- a/src/exo/download/coordinator.py
+++ b/src/exo/download/coordinator.py
@@ -220,9 +220,7 @@ class DownloadCoordinator:
                 model_directory=self._model_dir(model_id),
             )
             self.download_status[model_id] = failed
-            await self.event_sender.send(
-                NodeDownloadProgress(download_progress=failed)
-            )
+            await self.event_sender.send(NodeDownloadProgress(download_progress=failed))
             return
 
         # Start actual download
--- a/src/exo/download/tests/test_offline_mode.py
+++ b/src/exo/download/tests/test_offline_mode.py
@@ -202,9 +202,7 @@ class TestFetchFileListOffline:
             "metadata": {},
             "weight_map": {"model.layers.0.weight": "model.safetensors"},
         }
-        async with aiofiles.open(
-            model_dir / "model.safetensors.index.json", "w"
-        ) as f:
+        async with aiofiles.open(model_dir / "model.safetensors.index.json", "w") as f:
             await f.write(json.dumps(index_data))
 
         async with aiofiles.open(model_dir / "model.safetensors", "wb") as f:
@@ -229,6 +227,4 @@ class TestFetchFileListOffline:
         """When skip_internet=True and neither cache nor local files exist,
         raise FileNotFoundError."""
         with pytest.raises(FileNotFoundError, match="No internet"):
-            await fetch_file_list_with_cache(
-                model_id, "main", skip_internet=True
-            )
+            await fetch_file_list_with_cache(model_id, "main", skip_internet=True)


rltakashige:
Hi again! Just commented on your original issue.

src/exo/main.py
```diff
@@ -329,6 +337,13 @@ def parse(cls) -> Self:
             action="store_true",
             help="Disable the download coordinator (node won't download models)",
         )
+        parser.add_argument(
+            "--offline",
+            action="store_true",
+            default=os.getenv("EXO_OFFLINE") == "1"
+            or os.getenv("HF_HUB_OFFLINE") == "1",
+            help="Run in offline/air-gapped mode: skip internet checks, use only pre-staged local models (env: EXO_OFFLINE=1 or HF_HUB_OFFLINE=1)",
```
Evanev7: looks good bar this - no need for an env variable here, the --offline flag is sufficient and i'd prefer this behaviour to be explicit.

PR #1529, opened 2026-02-18
Fix tool calling

rltakashige:
## Motivation

GPT OSS tool calling issues.

## Changes

Fixes those and adds a bunch of evals for tool calling.
Fixes GLM5 prefix caching, where CacheList wasn't getting handled properly.
Extracts a bunch of the setup functionality of exo bench to a harness that can be reused elsewhere, such as in the tool calling eval.

### Automated Testing
Let's run the evals for all models


uv.lock
```diff
@@ -456,6 +457,7 @@ dependencies = [
 
 [package.metadata]
 requires-dist = [
+    { name = "httpx", specifier = ">=0.27.0" },
```
AlexCheema, line 460: should be in pyproject.toml?

AlexCheema: nvm it's from bench/pyproject

Evanev7:
trust this is fine but wondering if this intentionally added the kvcache changes as well?

PR #1533, opened 2026-02-18
Better onboarding UX

AlexCheema:
## Summary
- **Complete onboarding wizard**: 7-step flow guiding new users from Welcome → Your Devices (topology) → Add More Devices (animation) → Choose Model → Download → Load → Chat
- **Native macOS integration**: NSPopover welcome callout anchored to menu bar icon on first launch, polished DMG installer with drag-to-Applications arrow
- **Dashboard UX polish**: auto-download on model select, toast notifications, connection banner, skeleton loading, download progress in header, recommended model tags, sidebar hidden in home state for cleaner first impression
- **Settings & menu bar overhaul**: native Settings window with Advanced tab, onboarding reset, chat sidebar toggle

## Test plan
- [ ] Fresh install: verify onboarding wizard appears and flows Welcome → Topology → Animation → Model → Download → Load → Chat
- [ ] Verify topology shows real device data in onboarding step 2
- [ ] Verify selecting a model in the main dashboard picker auto-triggers download
- [ ] Verify chat sidebar is hidden on home view, appears when chat is active
- [ ] Verify DMG installer has white background with curved arrow
- [ ] Verify NSPopover appears anchored to menu bar icon on first launch

### DIFF
--- a/dashboard/src/lib/components/DeviceIcon.svelte
+++ b/dashboard/src/lib/components/DeviceIcon.svelte
@@ -67,8 +67,8 @@
   const studioMemH = $derived((ramPercent / 100) * studioMemTotalH);
 
   // ── MacBook dimensions (same ratios as TopologyGraph) ──
-  const mbW = $derived(size * 1.6);
-  const mbH = $derived(size * 1.15);
+  const mbW = $derived((size * 1.6 * 0.85) / 1.15);
+  const mbH = $derived(size * 0.85);
   const mbX = $derived(cx - mbW / 2);
   const mbY = $derived(cy - mbH / 2);
 


AlexCheema:
review pls @rltakashige

src/exo/master/api.py
```diff
@@ -1814,3 +1817,13 @@ async def get_trace_raw(self, task_id: str) -> FileResponse:
             media_type="application/json",
             filename=f"trace_{task_id}.json",
         )
+
+    _ONBOARDING_FILE = EXO_CONFIG_HOME / "onboarding_complete"
```
Evanev7: should be in cache home imo

src/exo/worker/engines/mlx/auto_parallel.py
```diff
@@ -609,6 +618,8 @@ def shard_heads(w: mx.array, sh: int = sh, eh: int = eh) -> mx.array:
                 layer.mlp.sharding_group = self.group
 
             mx.eval(layer)
+            if on_layer_loaded is not None:
+                on_layer_loaded(i + 1, total)
```
Evanev7: why is this 1-indexed?

src/exo/worker/runner/llm_inference/runner.py
```diff
@@ -170,11 +173,24 @@ def on_model_load_timeout() -> None:
...
+                    def on_layer_loaded(layers_loaded: int, total: int) -> None:
+                        event_sender.send(
+                            RunnerStatusUpdated(
+                                runner_id=runner_id,
+                                runner_status=RunnerLoading(
+                                    layers_loaded=layers_loaded, total_layers=total
+                                ),
+                            )
+                        )
+
```
Evanev7, line 187: id prefer if we yielded events from the loading rather than a callback, but this is fine

src/exo/master/api.py
Evanev7: also - make this a constant rather than an internal private class value

src/exo/worker/runner/llm_inference/runner.py
Evanev7: given this is a runner status, can you update the runners current_status field?

src/exo/master/api.py
AlexCheema: Done — moved to `EXO_CACHE_HOME` and made it a module-level constant `ONBOARDING_COMPLETE_FILE`.

src/exo/worker/engines/mlx/auto_parallel.py
AlexCheema: Fixed — now 0-indexed.

src/exo/worker/runner/llm_inference/runner.py
AlexCheema: Done — the callback now updates `current_status` via `nonlocal` before sending the event.

src/exo/worker/engines/mlx/auto_parallel.py
```diff
@@ -777,12 +792,12 @@ def shard_model(
         model: nn.Module,
         timeout_seconds: float,
         on_timeout: TimeoutCallback | None,
+        on_layer_loaded: LayerLoadedCallback | None = None,
```
rltakashige: Let's not use a default = None just to be intentional about it.

src/exo/worker/engines/mlx/utils_mlx.py
```diff
@@ -172,12 +175,25 @@ def load_mlx_items(
...
+        # Eval layers one by one for progress reporting
+        try:
+            inner = get_inner_model(model)
+            layers = get_layers(inner)
+            total = len(layers)
+            for i, layer in enumerate(layers):
+                mx.eval(layer)  # type: ignore
+                if on_layer_loaded is not None:
+                    on_layer_loaded(i, total)
+        except ValueError:
```
rltakashige: Why could we get a ValueError and why should we pass?

src/exo/worker/runner/llm_inference/runner.py
rltakashige: @Evanev7 I'm not a huge fan of a variable from a different scope getting magically updated within a function. Are you sure this is the best solution?

Evanev7: so, the other solution i can think of is making the runner a class with some class variables, and updating them through the self parameter. this is essentially what every class with callbacks is doing anyway. while i think maybe we should make the runner into a class, i don't think this magic self updating is avoidable with callbacks and id rather the current status field accurately reflected the runner status the supervisor will see.

src/exo/worker/engines/mlx/auto_parallel.py
AlexCheema: Done in 7bf52a26 — removed `= None` defaults from all `on_layer_loaded`, `on_timeout`, and `timeout_seconds` parameters across `auto_parallel.py` and `utils_mlx.py` so callers must be explicit.

src/exo/worker/engines/mlx/utils_mlx.py
AlexCheema: Good catch — `get_inner_model()` raises `ValueError` if the model lacks a `model`/`transformer`/`language_model` attribute, and `get_layers()` raises if it lacks `layers`/`h`. This means some model architectures won't support layer-by-layer progress reporting. Replaced the silent `pass` with `logger.debug(..., exc_info=True)` in 7bf52a26 so we'll see it in verbose logs when this happens.

src/exo/worker/runner/llm_inference/runner.py
AlexCheema: Agreed that the class refactor is a separate concern. The nonlocal approach works correctly for now and keeps the scope of this PR focused on the onboarding UX.

AlexCheema:
Re: defaulting to the fastest instance type — this is a good point. Currently the dashboard defaults to `MlxRing` (TCP/IP pipeline). We could detect RDMA capability from the topology data (the dashboard already tracks `nodeRdmaCtl` per node) and auto-select `MlxJaccl` when all nodes support it. @AlexCheema what do you think — should we add that logic to this PR or follow up separately?

AlexCheema:
Working on better defaults for these things in other PRs. most importantly we need the work in meta-instances to start to do better stuff here. most users should not have to choose any of this.

the problem doing it now is it will cause other issues e..g what if a user launches an instance before all his devices fully joined the cluster? well, now there's confusion because it defaulted to the best thing for a subset of his devices and will never update again...

src/exo/worker/engines/mlx/utils_mlx.py
Evanev7: use logger.opt(exception=e) instead of exc_info=true, this isn't the standard python logger.

PR #1536, opened 2026-02-18
add system ids

Evanev7:
addresses some election edge cases where a new worker with an old master would get stuck on the old workers buffer index - we now use new system ids each time we instantiate a node, and each event-producing system has a unique system id for its lifespan (until the master moves).

this is undertested currently, and shouldn't be merged until we've got a failing test for this behaviour on main.

### DIFF
--- a/src/exo/download/tests/test_coordinator_ack.py
+++ b/src/exo/download/tests/test_coordinator_ack.py
@@ -1,0 +1,98 @@
+from typing import Any
+
+import anyio
+import pytest
+
+from exo.download.coordinator import DownloadCoordinator
+from exo.download.shard_downloader import NoopShardDownloader
+from exo.shared.models.model_cards import ModelCard, ModelTask
+from exo.shared.types.common import ModelId, NodeId, SessionId
+from exo.shared.types.events import (
+    GlobalForwarderEvent,
+    LocalForwarderEvent,
+    NodeDownloadProgress,
+)
+from exo.shared.types.memory import Memory
+from exo.shared.types.worker.downloads import (
+    DownloadPending,
+)
+from exo.shared.types.worker.shards import PipelineShardMetadata
+from exo.utils.channels import channel
+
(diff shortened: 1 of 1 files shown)


src/exo/main.py
```diff
@@ -213,7 +205,6 @@ async def _elect_loop(self):
                 if result.is_new_master:
                     await anyio.sleep(0)
                     # Fresh counter for new session (buffer expects indices from 0)
```
rltakashige: Unnecessary comment

src/exo/worker/main.py
```diff
@@ -132,7 +131,7 @@ async def _forward_info(self, recv: Receiver[GatheredInfo]):
     async def _event_applier(self):
         with self.global_event_receiver as events:
             async for f_event in events:
-                if f_event.origin != self.session_id.master_node_id:
+                if f_event.session != self.session_id:
```
rltakashige, line 135: This is different right? I thought we only wanted to accept events from the master. Wouldn't this allow any events from any nodes in the same session?

Evanev7: yes - though the master **should** be the only one writing to the global event log in that session. i'll make it a lil more solid though

Evanev7:
bit chunkier now, split the master and workers forwarder event types so we preserve the origin field for the masters events and the system id for the workers.

src/exo/master/tests/test_master.py
```diff
@@ -133,7 +133,7 @@ def _get_events() -> Sequence[IndexedEvent]:
         logger.info("inject a TextGeneration Command")
         await command_sender.send(
             ForwarderCommand(
-                origin=node_id,
+                origin=SystemId("API"),
```
rltakashige, line 136: Magic string - extract to a constant?

rltakashige (approved):
Other than the minor comment, the changes are really good. Agreed that we should wait on some testing

PR #1538, opened 2026-02-18
remove nightly

Evanev7:
we have no good need for rust nightly (nor futures, for that matter)


JakeHillion (approved):
More cleanup, nice!

PR #1541, opened 2026-02-19
simplify the rust ident module

Evanev7:
this is partly dead code, partly narrowing the rust-python boundary in prep for future rewrites. no testing as this is all type safe refactoring.

### DIFF
--- a/rust/exo_pyo3_bindings/src/networking.rs
+++ b/rust/exo_pyo3_bindings/src/networking.rs
@@ -8,7 +8,7 @@
 use crate::r#const::MPSC_CHANNEL_SIZE;
 use crate::ext::{ByteArrayExt as _, FutureExt, PyErrExt as _};
 use crate::ext::{ResultExt as _, TokioMpscReceiverExt as _, TokioMpscSenderExt as _};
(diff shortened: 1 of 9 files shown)


rust/exo_pyo3_bindings/src/networking.rs
```diff
@@ -119,7 +119,7 @@ struct PyConnectionUpdate {
 
     /// Identity of the peer that we have connected to or disconnected from.
     #[pyo3(get)]
-    peer_id: PyPeerId,
+    peer_id: String,
```
JakeHillion, line 122: It feels a bit sad to change this interface to a String, was it painful to use before? Might be better to at least have a `struct PeerId(String)`

Evanev7: imho this is mostly something i'd want to fix on the python end (automatically wrap with NodeId() constructor) rather than Rust, i'll have a look but im not fond of duplicating rust and python types like the connectionmessage

Evanev7: ok, any solution i can find is going to be immediately cast to a string anyway, which feels completely pointless to me. i'm merging as-is.

PR #1543, opened 2026-02-19
feat: Mac Studio RDMA incompatible port warning in dashboard

AlexCheema:
Detects when devices are connected via the TB5 port next to ethernet (en2) on Mac Studio which does not support RDMA. Shows a warning banner with SVG port illustration and fix suggestion.


Evanev7:
contains obvious issues, namely duplicates the on_prefill_progress() call.

Evanev7:
also this PR doesn't seem to do what the title suggests

PR #1547, opened 2026-02-19
fix: suppress closure errors in runnersupervisor and force spawn start method

AlexCheema:
Part 5 of meta-instance split. Independent bug fixes that were bundled in PR #1519, now extracted for easier review.

### DIFF
--- a/src/exo/worker/main.py
+++ b/src/exo/worker/main.py
@@ -1,10 +1,9 @@
-import contextlib
 from collections import defaultdict
 from datetime import datetime, timezone
 from random import random
 
 import anyio
-from anyio import CancelScope, ClosedResourceError, fail_after
+from anyio import CancelScope, fail_after
 from loguru import logger
 
 from exo.download.download_utils import resolve_model_in_path
@@ -120,8 +119,7 @@ class Worker:
             self.command_sender.close()
             self.download_command_sender.close()
             for runner in self.runners.values():
-                with contextlib.suppress(ClosedResourceError):
-                    runner.shutdown()
+                runner.shutdown()
 
     async def _forward_info(self, recv: Receiver[GatheredInfo]):
         with recv as info_stream:
@@ -267,8 +265,7 @@ class Worker:
                             )
                         )
                     finally:
-                        with contextlib.suppress(ClosedResourceError):
-                            runner.shutdown()
+                        runner.shutdown()
                 case CancelTask(
                     cancelled_task_id=cancelled_task_id, runner_id=runner_id
                 ):
--- a/src/exo/worker/runner/runner_supervisor.py
+++ b/src/exo/worker/runner/runner_supervisor.py
@@ -106,13 +106,18 @@ class RunnerSupervisor:
     def shutdown(self):
         logger.info("Runner supervisor shutting down")
         self._tg.cancel_tasks()
-        self._ev_recv.close()
-        self._task_sender.close()
         if not self._cancel_watch_runner.cancel_called:
             self._cancel_watch_runner.cancel()
         with contextlib.suppress(ClosedResourceError):
+            self._ev_recv.close()
+        with contextlib.suppress(ClosedResourceError):
+            self._task_sender.close()
+        with contextlib.suppress(ClosedResourceError):
+            self._event_sender.close()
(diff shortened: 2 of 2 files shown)


src/exo/download/coordinator.py
JakeHillion: This seems like a nice win.

src/exo/main.py
```diff
@@ -258,7 +258,7 @@ def main():
     target = min(max(soft, 65535), hard)
     resource.setrlimit(resource.RLIMIT_NOFILE, (target, hard))
 
-    mp.set_start_method("spawn")
+    mp.set_start_method("spawn", force=True)
```
JakeHillion, line 255: What does this fix that was broken before?

src/exo/worker/main.py
```diff
@@ -116,7 +117,8 @@ async def run(self):
             self.command_sender.close()
             self.download_command_sender.close()
             for runner in self.runners.values():
-                runner.shutdown()
+                with contextlib.suppress(ClosedResourceError):
+                    runner.shutdown()
 
     async def _forward_info(self, recv: Receiver[GatheredInfo]):
```
JakeHillion, line 124: Same question, what error case is this catching?

Evanev7: id like to add - these should be caught inside runner.shutdown(), not outside it.

src/exo/download/coordinator.py
AlexCheema: Thanks! Yeah, the coordinator was doing redundant work tracking download states that the download manager already handles.

src/exo/main.py
AlexCheema: mp.set_start_method("spawn") raises RuntimeError: context has already been set if any import or library initialization has already triggered multiprocessing context setup before main() runs. This happens when MLX/PyTorch calls mp.get_context() during import, or with pytest-xdist, or on some macOS configurations. We hit this crash on the Mac Mini cluster — all 4 nodes failed. force=True is safe here because we always set it to spawn (the right choice on macOS).

src/exo/worker/main.py
AlexCheema: RunnerSupervisor.shutdown() calls .close() on four channels. If the runner subprocess has already crashed, the _forward_events() loop exits and closes _event_sender as part of its own cleanup. When the finally block then calls runner.shutdown(), it tries to close _event_sender again, raising ClosedResourceError. contextlib.suppress(ClosedResourceError) is the right fix because shutdown is best-effort cleanup — if channels are already closed, the runner is already torn down.

AlexCheema: Good point — moving the suppress(ClosedResourceError) inside runner.shutdown() itself is cleaner since it is the shutdown method that should handle already-closed channels gracefully. Will update.

src/exo/download/coordinator.py
JakeHillion: Let me be more specific: please split this download manager change out into a separate PR with a better name and description.

src/exo/download/coordinator.py
```diff
@@ -338,7 +338,17 @@ async def _emit_existing_download_progress(self) -> None:
...
+                        if (
+                            progress.downloaded_bytes.in_bytes
+                            >= progress.total_bytes.in_bytes
+                            > 0
+                        ):
+                            status = DownloadCompleted(
+                                node_id=self.node_id,
+                                shard_metadata=progress.shard,
+                                total_bytes=progress.total_bytes,
+                            )
```
Evanev7: move this section to be an additional condition (so `if progress.status == "complete" or (progress.downloaded_bytes >= progress.total_bytes > 0)`)

src/exo/download/coordinator.py
AlexCheema: Moved into separate PR and changed it to fix the root cause: https://github.com/exo-explore/exo/pull/1582

src/exo/download/coordinator.py
AlexCheema: Fixed the root cause here: https://github.com/exo-explore/exo/pull/1582

Evanev7 (approved):
update the pr description before you merge, these are completely reasonable

PR #1548, opened 2026-02-19
Add thinking in thinking blocks, and fix DeepSeek interleaved tool calls

rltakashige:
## Motivation

OpenCode shows <think> tags and not thinking blocks as we aren't following the API specs properly.

Claude was also getting horrible prefix cache hits because it sends headers.

## Changes

Handle thinking tokens properly by placing them in think tags for each of the API endpoints.
Also support DeepSeekV3.2 tool calling properly as a minor feature.
Strips Claude headers at the API level.

### Manual Testing
Tested OpenCode manually
Needs testing with Claude.

### Automated Testing
All CI and tests passing - added a new e2e test for DeepSeekV32 tool parsing.

### DIFF
--- a/dashboard/src/routes/+page.svelte
+++ b/dashboard/src/routes/+page.svelte
@@ -114,6 +114,74 @@
   });
   let tb5InfoDismissed = $state(false);
 
+  // Detect Mac Studio nodes using RDMA on en2 (the port next to ethernet — RDMA doesn't work there)
+  const macStudioEn2RdmaWarning = $derived.by(() => {
+    const edges = data?.edges;
+    const ids = tbIdentifiers;
+    const rdmaCtl = rdmaCtlData;
+    if (!edges || !ids || !rdmaCtl) return null;
+
+    const affectedConnections: Array<{
+      nodeId: string;
+      nodeName: string;
+      peerNodeId: string;
+      peerNodeName: string;
+      rdmaIface: string;
+    }> = [];
+
+    const isMacStudio = (node: (typeof data.nodes)[string] | undefined) =>
+      node?.system_info?.model_id === "Mac Studio";
+
+    for (const edge of edges) {
+      if (!edge.sourceRdmaIface && !edge.sinkRdmaIface) continue;
+
+      const sourceNode = data?.nodes?.[edge.source];
+      if (
+        isMacStudio(sourceNode) &&
+        edge.sourceRdmaIface === "rdma_en2" &&
+        rdmaCtl[edge.source]?.enabled
+      ) {
+        affectedConnections.push({
+          nodeId: edge.source,
+          nodeName:
+            sourceNode?.friendly_name || edge.source.slice(0, 8) + "...",
+          peerNodeId: edge.target,
+          peerNodeName:
+            data?.nodes?.[edge.target]?.friendly_name ||
+            edge.target.slice(0, 8) + "...",
+          rdmaIface: "en2",
+        });
+      }
+
+      const sinkNode = data?.nodes?.[edge.target];
+      if (
+        isMacStudio(sinkNode) &&
+        edge.sinkRdmaIface === "rdma_en2" &&
+        rdmaCtl[edge.target]?.enabled
+      ) {
+        affectedConnections.push({
+          nodeId: edge.target,
+          nodeName: sinkNode?.friendly_name || edge.target.slice(0, 8) + "...",
+          peerNodeId: edge.source,
+          peerNodeName:
+            sourceNode?.friendly_name || edge.source.slice(0, 8) + "...",
+          rdmaIface: "en2",
+        });
+      }
+    }
+
+    // Deduplicate by nodeId
+    const seen = new Set<string>();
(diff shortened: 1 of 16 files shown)


Evanev7:
Pretty sure mlx also has patches for deepseek

Evanev7:
are we sure we need to be vendoring this?

rltakashige:
> are we sure we need to be vendoring this?

Yup - switched to using more of mlx lm's code and vendoring less. The main thing is parsing.

AlexCheema:
## Test Report: PR #1548 — `leo/fix-thinking-api`

**Cluster**: 4x Mac mini M4 Pro (24GB each), macOS, mDNS peer discovery
**Model**: `mlx-community/Qwen3-30B-A3B-4bit` (17.6 GB, thinking-capable)
**Branch**: `leo/fix-thinking-api` (commit `d7e288b2`)

### Screenshots

**4-node cluster dashboard** — all Mac Minis discovered and model placed:
4-node cluster dashboard

### Results Summary

| Test | Result | Notes |
|------|--------|-------|
| Cluster formation (4 nodes) | ✅ **PASS** | All 4 nodes discovered via mDNS |
| Model placement (4-node pipeline) | ✅ **PASS** | Placed and runner ready |
| OpenAI streaming (`/v1/chat/completions`) | ⚠️ **PARTIAL** | Thinking tokens in `content` as `<think>`, NOT in `reasoning_content` |
| Anthropic streaming (`/v1/messages`) | ⚠️ **PARTIAL** | Thinking tokens in `text` block, NOT in `thinking` block |

### Issue 1: Missing Qwen3 Thinking Parser

The PR adds `is_thinking: bool` to `TokenChunk` and `GenerationResponse`, and the adapters correctly route `is_thinking=True` chunks. However, **Qwen3 models have no thinking parser**. The runner only has parsers for:
- **GPT-OSS** (`parse_gpt_oss`): detects `analysis` channel
- **DeepSeek V3.2** (`parse_deepseek_v32`): detects `<think>...</think>` tags

**Qwen3 outputs `<think>...</think>` tags as raw text** and the runner never sets `is_thinking=True`.

**Observed API response** (OpenAI format, streaming):
```json
{"delta": {"content": "<think>", "reasoning_content": null}}
{"delta": {"content": "Okay, the user is asking...", "reasoning_content": null}}
{"delta": {"content": "</think>", "reasoning_content": null}}
{"delta": {"content": "\n\n2 + 2 = **4**.", "reasoning_content": null}}
```

**Expected** (if Qwen3 parser existed):
```json
{"delta": {"content": null, "reasoning_content": "<think>"}}
{"delta": {"content": null, "reasoning_content": "Okay, the user is asking..."}}
{"delta": {"content": null, "reasoning_content": "</think>"}}
{"delta": {"content": "2 + 2 = **4**.", "reasoning_content": null}}
```

**Suggested fix**: Add a `parse_qwen3_thinking()` generator similar to `parse_deepseek_v32()` that:
1. Detects `<think>` token and starts buffering as thinking
2. Sets `is_thinking=True` on tokens between `<think>` and `</think>`
3. Transitions to `is_thinking=False` after `</think>`

This would be ~30 lines, following the same pattern as `parse_deepseek_v32()`.

### Issue 2: Anthropic Messages API — Same root cause

**Observed** (`/v1/messages`, streaming):
```
event: content_block_start
data: {"type":"content_block_start","index":0,"content_block":{"type":"text","text":""}}

event: content_block_delta
data: {"type":"content_block_delta","index":0,"delta":{"type":"text_delta","text":"<think>"}}
...thinking tokens as text_delta...
data: {"type":"content_block_delta","index":0,"delta":{"type":"text_delta","text":"</think>"}}
...response tokens as text_delta...
```

**Expected** (with Qwen3 parser):
```
event: content_block_start  → thinking block at index 0
...thinking_delta events...
event: content_block_stop   → close thinking block
event: content_block_start  → text block at index 1
...text_delta events...
```

The `claude.py` adapter has the correct logic (`ClaudeThinkingBlock`, `ClaudeThinkingDelta`), but it never fires because `chunk.is_thinking` is always `False` for Qwen3.

### What the PR does well

- Adds `is_thinking: bool` to `TokenChunk` and `GenerationResponse` — clean extension point
- Claude adapter correctly creates `ClaudeThinkingBlock`/`ClaudeThinkingDelta`
- OpenAI adapter correctly routes to `reasoning_content`
- DeepSeek V3.2 DSML encoding is comprehensive with full tool call parsing
- New `parse_deepseek_v32()` generator properly sets `is_thinking=True`

### Recommendation

Add a Qwen3 thinking parser — this is the most commonly used thinking model in exo clusters, so the PR's improvements won't be visible to most users without it.

rltakashige:
image
GPT OSS Claude approves.

Evanev7:
I wonder if this can be cleaned up with the tool parser abstraction, there's a lot of duped code between here and the regular tool parser.

Evanev7:
im happy to merge now and take a look later though

Evanev7:
seems like broadly the tool parser has been duped because we don't have single-token <think> and <tool call> tokens. is that actually true?

PR #1549, opened 2026-02-19
move messaging into rust

Evanev7:
the main body of the rust refactor. fixes the tokio panic on shutdown. simplifies the networking module significantly. doesn't touch lp2p behaviour

### DIFF
--- a/rust/exo_pyo3_bindings/src/networking.rs
+++ b/rust/exo_pyo3_bindings/src/networking.rs
@@ -1,26 +1,24 @@
-#![allow(
-    clippy::multiple_inherent_impl,
-    clippy::unnecessary_wraps,
-    clippy::unused_self,
-    clippy::needless_pass_by_value
-)]
+use std::pin::Pin;
+use std::sync::Arc;
 
 use crate::r#const::MPSC_CHANNEL_SIZE;
 use crate::ext::{ByteArrayExt as _, FutureExt, PyErrExt as _};
-use crate::ext::{ResultExt as _, TokioMpscReceiverExt as _, TokioMpscSenderExt as _};
+use crate::ext::{ResultExt as _, TokioMpscSenderExt as _};
 use crate::ident::PyKeypair;
+use crate::networking::exception::{
+    PyAllQueuesFullError, PyMessageTooLargeError, PyNoPeersSubscribedToTopicError,
+};
 use crate::pyclass;
-use libp2p::futures::StreamExt as _;
-use libp2p::gossipsub;
-use libp2p::gossipsub::{IdentTopic, Message, MessageId, PublishError};
-use libp2p::swarm::SwarmEvent;
-use networking::discovery;
-use networking::swarm::create_swarm;
+use futures_lite::{Stream, StreamExt as _};
+use libp2p::gossipsub::PublishError;
+use networking::swarm::{FromSwarm, ToSwarm, create_swarm};
+use pyo3::exceptions::PyRuntimeError;
 use pyo3::prelude::{PyModule, PyModuleMethods as _};
 use pyo3::types::PyBytes;
-use pyo3::{Bound, Py, PyErr, PyResult, PyTraverseError, PyVisit, Python, pymethods};
-use pyo3_stub_gen::derive::{gen_stub_pyclass, gen_stub_pyclass_enum, gen_stub_pymethods};
-use std::net::IpAddr;
+use pyo3::{Bound, Py, PyAny, PyErr, PyResult, Python, pymethods};
+use pyo3_stub_gen::derive::{
+    gen_methods_from_python, gen_stub_pyclass, gen_stub_pyclass_complex_enum, gen_stub_pymethods,
+};
 use tokio::sync::{Mutex, mpsc, oneshot};
 
 mod exception {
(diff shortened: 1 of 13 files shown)


rust/exo_pyo3_bindings/src/networking.rs
```diff
@@ -1,26 +1,21 @@
...
-    clippy::unnecessary_wraps,
-    clippy::unused_self,
-    clippy::needless_pass_by_value
-)]
+use std::sync::Arc;
 
 use crate::r#const::MPSC_CHANNEL_SIZE;
 use crate::ext::{ByteArrayExt as _, FutureExt, PyErrExt as _};
-use crate::ext::{ResultExt as _, TokioMpscReceiverExt as _, TokioMpscSenderExt as _};
+use crate::ext::{ResultExt as _, TokioMpscSenderExt as _};
```
JakeHillion, line 6: nit: is this a widely used pattern? I haven't seen importing traits as _ before, I have always imported them directly.

rust/networking/examples/chatroom.rs
JakeHillion, line 1: Are these examples actually helpful? Can we move them to integration tests if they are so they run? Otherwise, delete it.

rust/networking/src/swarm.rs
```diff
@@ -15,8 +18,144 @@ pub type Swarm = libp2p::Swarm<Behaviour>;
...
+    Subscribe {
+        topic: String,
+        result_sender: oneshot::Sender<Result<bool, gossipsub::SubscriptionError>>,
+    },
+    Publish {
+        topic: String,
+        data: Vec<u8>,
+        result_sender: oneshot::Sender<Result<gossipsub::MessageId, gossipsub::PublishError>>,
+    },
+}
```
JakeHillion, line 36: Could you add a comment explaining the `result_sender` architecture? It's not immediately clear at first why we need to do this over getting the return from the async processor directly.

rust/networking/src/swarm.rs
```diff
@@ -15,8 +18,144 @@ pub type Swarm = libp2p::Swarm<Behaviour>;
...
+                _ = result_sender.send(result)
+            }
+        }
+    }
+}
+impl Stream for Swarm {
+    type Item = FromSwarm;
+    fn poll_next(
+        mut self: std::pin::Pin<&mut Self>,
+        cx: &mut std::task::Context<'_>,
```
JakeHillion: nit: s/cx/ctx/

rust/networking/src/swarm.rs
```diff
@@ -15,8 +18,144 @@ pub type Swarm = libp2p::Swarm<Behaviour>;
...
+                        data,
+                    })),
+                    SwarmEvent::Behaviour(BehaviourEvent::Discovery(
+                        discovery::Event::ConnectionEstablished { peer_id, .. },
+                    )) => Poll::Ready(Some(FromSwarm::Discovered { peer_id })),
+                    SwarmEvent::Behaviour(BehaviourEvent::Discovery(
+                        discovery::Event::ConnectionClosed { peer_id, .. },
+                    )) => Poll::Ready(Some(FromSwarm::Expired { peer_id })),
+                    // continue to re-poll after consumption
+                    _ => continue,
```
JakeHillion: How many entries are we missing here? I'm really not a fan of wildcard continues, but if it's dozens of cases then it makes sense.

rust/networking/src/swarm.rs
```diff
@@ -15,8 +18,144 @@ pub type Swarm = libp2p::Swarm<Behaviour>;
...
+                    .behaviour_mut()
+                    .gossipsub
+                    .publish(gossipsub::IdentTopic::new(topic), data);
+                // send response oneshot (or exit if connection closed)
+                _ = result_sender.send(result)
+            }
+        }
+    }
+}
+impl Stream for Swarm {
```
JakeHillion: I opened a PR, #1597, to do this in a bit of a different way. Rather than implementing poll it implements a high level stream, which I think is a bit harder to get wrong. Let me know what you think.

src/exo/routing/connection_message.py
```diff
@@ -1,37 +1,15 @@
...
-                return ConnectionMessageType.Connected
-            case ConnectionUpdateType.Disconnected:
-                return ConnectionMessageType.Disconnected
-
-
 class ConnectionMessage(CamelCaseModel):
     node_id: NodeId
-    connection_type: ConnectionMessageType
-    remote_ipv4: str
-    remote_tcp_port: int
```
JakeHillion, line 28: was this not used before?

src/exo/routing/router.py
```diff
@@ -120,7 +121,8 @@ async def register_topic[T: CamelCaseModel](self, topic: TypedTopic[T]):
             send = self.networking_receiver.clone_sender()
         router = TopicRouter[T](topic, send)
         self.topic_routers[topic.topic] = cast(TopicRouter[CamelCaseModel], router)
-        await self._networking_subscribe(str(topic.topic))
+        if self._tg is not None:
```
JakeHillion: bad rebase?

src/exo/routing/router.py
```diff
@@ -174,27 +178,33 @@ async def _networking_unsubscribe(self, topic: str):
...
+                    logger.trace(
+                        f"Received message on connection_messages with payload {message}"
+                    )
+                    if CONNECTION_MESSAGES.topic in self.topic_routers:
+                        router = self.topic_routers[CONNECTION_MESSAGES.topic]
+                        assert router.topic.model_type == ConnectionMessage
+                        router = cast(TopicRouter[ConnectionMessage], router)
+                        await router.publish(message)
+                case _:
+                    raise AssertionError("exhaustive net messages have been checked")
```
JakeHillion: I'm in two minds, but should this be a `logger.error`? It's clearly bad and is breaking an invariant, but given we can't check it at build time I think it's better to error loudly but not crash - our future testing should then pick it up, but not kill people's processes if we mess up.

rust/exo_pyo3_bindings/src/networking.rs
Evanev7: i'll clean this up in a followup & review our clippy settings then

src/exo/routing/connection_message.py
Evanev7: not for a while - i prefer our http discovery for writing events into state instead of this, and its nice to keep gossipsub independent of actual connection discovery tbh

rust/networking/src/swarm.rs
Evanev7: merged

src/exo/routing/router.py
Evanev7: resolved in person

rust/networking/examples/chatroom.rs
Evanev7: i've found it useful in a way that's been hard to automate, for demonstrating the intended use of the library outside of exo_pyo3_bindings.

rust/networking/src/swarm.rs
Evanev7: done

src/exo/routing/router.py
Evanev7: fiiiiiiiiiiiiiiiiiiiiiiiiiiiiiiiiiiiiiiiiiiiiiiiiiine

rust/networking/src/swarm.rs
JakeHillion: potentially dumb question, is it?

JakeHillion (approved):
I think we're still missing a comment, otherwise this is good to go.

PR #1553, opened 2026-02-19
Ciaran/handle disconnects

ciaranbor:
## Motivation

  - Transient network hiccups (brief Wi-Fi drops, high latency) cause immediate peer disconnects because a single ping failure triggers connection closure. This makes the cluster fragile under real-world conditions.
  - DownloadCoordinator had its own independent event forwarding path, bypassing the Worker's retry/out_for_delivery mechanism, so download progress events were permanently lost during partitions. The master would get stuck waiting for download event indices until the worker node was timed out

## Changes

  - N-strike ping tolerance (rust/networking/src/discovery.rs): Replaced instant disconnect on ping failure with a 3-consecutive-failure threshold. Relaxed ping interval to 5s and timeout to 10s. Tracks per-connection failure counts, resets on success.
  - Route DC events through Worker (src/exo/download/coordinator.py, src/exo/main.py): Removed DownloadCoordinator's own _forward_events loop and session_id/event_index_counter fields. DC now sends events via worker.event_sender.clone(), sharing the Worker's index sequence
  and retry mechanism.

## Why It Works

  - Routing DC events through the Worker's event channel means they get the same out_for_delivery retry buffer, so no events are permanently lost during partitions.

### Manual Testing

  - Deploy to a multi-node cluster and verify peers remain connected through brief network interruptions

### Automated Testing

  - Partition recovery test (src/exo/master/tests/test_partition_recovery.py): End-to-end test simulating a network partition with a proxy that drops events, verifying that both Worker heartbeats and DC download events are retried and recovered after the partition heals.

### DIFF
--- a/src/exo/master/tests/test_partition_recovery.py
+++ b/src/exo/master/tests/test_partition_recovery.py
@@ -1,0 +1,250 @@
+from collections.abc import Awaitable, Callable
+from datetime import datetime, timedelta, timezone
+from itertools import count
+from pathlib import Path
+from typing import AsyncIterator
+
+import anyio
+import pytest
+
+from exo.download.coordinator import DownloadCoordinator
+from exo.download.shard_downloader import RepoDownloadProgress, ShardDownloader
+from exo.master.main import Master
+from exo.master.tests.conftest import create_node_memory
+from exo.shared.models.model_cards import ModelCard, ModelTask
+from exo.shared.types.commands import (
+    ForwarderCommand,
+    ForwarderDownloadCommand,
+    StartDownload,
+)
+from exo.shared.types.common import ModelId, NodeId, SessionId
+from exo.shared.types.events import (
+    ForwarderEvent,
+    NodeDownloadProgress,
+    NodeGatheredInfo,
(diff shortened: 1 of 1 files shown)


AlexCheema:
Could you take a look @Evanev7 especially networking changes.
@ciaranbor says this improves stability significantly when manually testing. he also has a unit test for the download fix.

Evanev7:
why is claude so desperate to nuke the mdns ttl.... ill take a look asap, probably tomorrow though

my bad! it was the ping timing, this is fine.

JakeHillion:
I gave this a test with exo running over SSH with nohup and a terminate session, and this didn't bring my cluster to life. Was I wrong to expect it to? Was possible testing the wrong thing!

Evanev7:
the ping changes are a-ok, im not a fan of sharing senders, it breaks some local reasoning in a way i think is a bit icky. the outcome of the pr is perfect, im gonna refactor slightly to make sending events a bit more trivial

Evanev7:
closing in favour of #1536

PR #1555, opened 2026-02-19
Prevent common fatal crashes

rltakashige:
## Motivation
Occasionally, memory does not get released when we shut down. There is no reason to delay deleting the model.

Also handles can become None during shutdown, causing TypeErrors which are not handled and bringing down exo.

Similarly, we were closing the event sender in the wrong place.

Also let's not verify the SSL certificate for http connections to local peers, as this is failing sometimes and crashing. 

### Manual Testing
No more crashes as described.

### DIFF
--- a/dashboard/src/lib/stores/app.svelte.ts
+++ b/dashboard/src/lib/stores/app.svelte.ts
@@ -1652,11 +1652,12 @@ class AppStore {
       if (!reader) throw new Error("No response body");
 
       let fullContent = prefixText;
+      let streamedThinking = "";
       const collectedTokens: TokenData[] = [...tokensToKeep];
 
       interface ChatCompletionChunk {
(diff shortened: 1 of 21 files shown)


Evanev7:
nice

src/exo/utils/info_gatherer/net_profile.py
```diff
@@ -108,7 +108,7 @@ async def _probe(
                 await send.send((target_ip, expected_node_id))
 
     async with (
-        httpx.AsyncClient(timeout=timeout, limits=limits) as client,
+        httpx.AsyncClient(timeout=timeout, limits=limits, verify=False) as client,
```
JakeHillion, line 111: This is pretty suspicious, it should be a no-op for `http://` URLs which are what we should be using.

PR #1556, opened 2026-02-19
Prioritise tb for ring instances

rltakashige:
## Motivation

TB has better bandwidth and latency than ethernet. We should prioritise TB5 where possible. This drastically improves distributed image generation performance.

### Manual Testing
Saw on the dashboard that TB (169.254) addresses were prioritised.

Tested that image models scale much better.

### Automated Testing
No regression on Kimi K2.5


rltakashige:
No issues on that regard, with the fixes I merged previously for RDMA GPU locks:

image

image

image

PR #1570, opened 2026-02-20
Refactor runner into separate runners

rltakashige:
## Motivation

We're going to be refactoring the llm inference code, so we should split the runner up into parts while we can.

### Manual Testing
Works on single node, at least.

### Automated Testing
Passes CI. Will be tested by our tests today.


rltakashige:
Tested on image model!

PR #1572, opened 2026-02-20
event router

Evanev7:
replace the nack & resend logic in the worker/download coordinator with a dedicated subsystem in front of the topic router. this centralizes that logic (and the concept of system ids) to reduce replication in the codebase. each system reading or writing events now gets a clean stream of events in and can trust written events will be retried until acknowledged.


src/exo/master/api.py
```diff
@@ -1623,38 +1619,33 @@ async def run_api(self, ev: anyio.Event):
...
-                            assert not isinstance(event.chunk, ImageChunk)
-                            try:
-                                await queue.send(event.chunk)
-                            except BrokenResourceError:
-                                self._text_generation_queues.pop(event.command_id, None)
-                    if isinstance(event, TracesMerged):
-                        self._save_merged_trace(event)
+        idx = 0
+        with self.event_receiver as events:
+            async for event in events:
```
ciaranbor: event is shadowed. Perhaps `for received_event in received_events:`

src/exo/worker/main.py
```diff
@@ -59,38 +55,27 @@ class Worker:
...
-        self.session_id: SessionId = session_id
-
-        self.global_event_receiver = global_event_receiver
-        self.local_event_sender = local_event_sender
+        self.event_receiver = event_receiver
+        self.event_sender = event_sender
         self.command_sender = command_sender
         self.download_command_sender = download_command_sender
-        self.event_buffer = OrderedBuffer[Event]()
         self.out_for_delivery: dict[EventId, LocalForwarderEvent] = {}
```
ciaranbor: unused now

src/exo/master/api.py
```diff
@@ -1623,38 +1619,33 @@ async def run_api(self, ev: anyio.Event):
...
-                            event.command_id, None
-                        ):
-                            assert not isinstance(event.chunk, ImageChunk)
-                            try:
-                                await queue.send(event.chunk)
-                            except BrokenResourceError:
-                                self._text_generation_queues.pop(event.command_id, None)
-                    if isinstance(event, TracesMerged):
-                        self._save_merged_trace(event)
+        idx = 0
```
ciaranbor: Don't need this

ciaranbor (approved):
lgtm. Was thinking this would makes sense to do when looking into the download coordinator last week

PR #1574, opened 2026-02-20
worker: add EXO_MODELS_PATH for pre-downloaded model directories

JakeHillion:
Users with pre-existing model files (e.g. on shared NFS mounts or from prior downloads) had no way to point exo at those directories without going through the download coordinator. EXO_MODELS_DIR only moves the download target directory, it doesn't support read-only search paths.

Added EXO_MODELS_PATH environment variable as a colon-separated list of directories to search for models. When the worker's plan loop encounters a DownloadModel task, it checks these directories first and emits a synthetic DownloadCompleted event if found, bypassing the download coordinator entirely. The runner's build_model_path also checks these directories first so the correct path is used during model loading.

This keeps the existing event sourcing state machine unchanged — the DownloadCompleted event propagates naturally through the system, so _load_model and all downstream logic work without modification.

Test plan:
- `s1@s1s-Mac-Studio ~ % EXO_LIBP2P_NAMESPACE=jake EXO_MODELS_PATH="/Volumes/Definitely Leo's SSD" nix --extra-experimental-features 'nix-command flakes' run github:exo-explore/exo/f2babbc2f742357d97dc177619fec062ef545be4`
- Started mlx-community/Qwen3-Coder-Next-4bit - it's present on the disk and it worked.
- Renamed one safetensor of mlx-community/Qwen3-Coder-Next-4bit on the disk. It then started the download locally, as expected.

### DIFF
--- a/src/exo/download/download_utils.py
+++ b/src/exo/download/download_utils.py
@@ -20,7 +20,6 @@ from huggingface_hub import (
 )
 from loguru import logger
 from pydantic import (
-    DirectoryPath,
     TypeAdapter,
 )
 
@@ -31,7 +30,7 @@ from exo.download.huggingface_utils import (
     get_hf_endpoint,
     get_hf_token,
 )
-from exo.shared.constants import EXO_MODELS_DIR
+from exo.shared.constants import EXO_MODELS_DIR, EXO_MODELS_PATH
 from exo.shared.models.model_cards import ModelTask
 from exo.shared.types.common import ModelId
 from exo.shared.types.memory import Memory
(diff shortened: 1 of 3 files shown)


src/exo/download/download_utils.py
```diff
@@ -111,7 +110,27 @@ def map_repo_download_progress_to_download_progress_data(
...
+
+    Checks each directory for both normalized (org--model) and
+    raw (org/model) subdirectory names.  A candidate is only returned
+    if ``is_model_directory_complete`` confirms all weight files are present.
+    """
+    if EXO_MODELS_PATH is None:
+        return None
+    normalized = model_id.normalize()
+    for search_dir in EXO_MODELS_PATH:
+        for candidate in (search_dir / normalized, search_dir / model_id):
```
rltakashige: As discussed in person, let's keep this if this works for .huggingface cache; otherwise, let's keep this consistent as just normalized path for now.

JakeHillion: Yeah, it doesn't. They prefix with model-- it seems, but we can sort this out later. Went with only the normalised path to match existing behaviour.

PR #1575, opened 2026-02-20
fix: immediate cancel check after prefill completes

mlpy0:
## Problem

When a request is cancelled during prefill, the cancellation is not detected until `check_for_cancel_every` additional tokens have been generated. This is because `tokens_since_last_cancel_check` is initialized to `0`, meaning the first cancel check only happens after generating `check_for_cancel_every` tokens post-prefill.

For long prefills (which are the most likely to be cancelled), this adds unnecessary latency before the cancellation is actually honoured.

## Fix

Initialize `tokens_since_last_cancel_check` to `check_for_cancel_every` instead of `0`, so the very first token generated after prefill triggers an immediate cancel check.

```diff
- tokens_since_last_cancel_check = 0
+ tokens_since_last_cancel_check = check_for_cancel_every
```

## Impact

- Cancellations issued during prefill are detected immediately when generation begins
- No change in behaviour for non-cancelled requests (the counter resets to `0` after each check as before)
- 1 line changed

### DIFF
--- a/dashboard/src/routes/downloads/+page.svelte
+++ b/dashboard/src/routes/downloads/+page.svelte
@@ -29,7 +29,12 @@
         etaMs: number;
         modelDirectory?: string;
       }
-    | { kind: "pending"; modelDirectory?: string }
+    | {
+        kind: "pending";
+        downloaded: number;
+        total: number;
+        modelDirectory?: string;
+      }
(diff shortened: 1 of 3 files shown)


rltakashige:
Thank you! This looks reasonable.

PR #1578, opened 2026-02-20
downloads: add read_only flag to DownloadCompleted for EXO_MODELS_PATH

JakeHillion:
Models in EXO_MODELS_PATH are pre-downloaded into read-only directories
and must not be deleted. The DownloadCoordinator had no awareness of
these paths, so they never appeared as completed downloads in cluster
state, and the bench harness could attempt to delete them when freeing
disk space.

Added a `read_only: bool` field to `DownloadCompleted` (default False).
The DownloadCoordinator now checks `resolve_model_in_path` in
`_start_download`, proactively scans EXO_MODELS_PATH in
`_emit_existing_download_progress` to emit DownloadCompleted events for
all pre-downloaded models (overriding DownloadPending from the regular
scan), and refuses deletion of read-only models. The bench harness
filters out read-only models from deletion candidates.

Test plan:
- Ran with EXO_MODELS_PATH. Available models now show as downloaded in
  the UI. There isn't good UI for the fact they can't be deleted, but it
  should work with exo_bench.

### DIFF
--- a/src/exo/worker/main.py
+++ b/src/exo/worker/main.py
@@ -227,6 +227,7 @@ class Worker:
                                     shard_metadata=shard,
                                     model_directory=str(found_path),
                                     total=shard.model_card.storage_size,
+                                    read_only=True,
                                 )
                             )
                         )
--- a/bench/harness.py
+++ b/bench/harness.py
@@ -365,7 +365,7 @@ def run_planning_phase(
                 f"have {avail // (1024**3)}GB. Use --danger-delete-downloads to free space."
             )
 
-        # Delete from smallest to largest
+        # Delete from smallest to largest (skip read-only models from EXO_MODELS_PATH)
         completed = [
             (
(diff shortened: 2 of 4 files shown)


src/exo/worker/main.py
```diff
@@ -227,6 +227,7 @@ async def plan_step(self):
                                     shard_metadata=shard,
                                     model_directory=str(found_path),
                                     total=shard.model_card.storage_size,
+                                    read_only=True,
```
rltakashige, line 230: Surely this should not be true?

JakeHillion: This is the path for it found the instance in the PATH, and this is to kick us onto the next stage of the plan loop. It should be read only as it came from the PATH - a bit confusing though!

rltakashige (approved):
Looks good then! Though I think this will need to be refactored eventually.

PR #1579, opened 2026-02-20
feat: add POST /v1/cancel/{command_id} endpoint

mlpy0:
## Summary

- Adds explicit `POST /v1/cancel/{command_id}` REST endpoint to cancel in-flight text and image generation commands by ID
- Previously, cancellation only worked via HTTP disconnect (client closes SSE connection, triggering `anyio.get_cancelled_exc_class()`). Non-streaming clients and external consumers had no way to cleanly cancel active generation
- The endpoint looks up the command in active generation queues, sends `TaskCancelled` to notify workers, and closes the sender stream. Returns 404 (OpenAI error format) if the command is not found or already completed

## Changes

| File | Change |
|------|--------|
| `src/exo/shared/types/api.py` | Add `CancelCommandResponse` model |
| `src/exo/master/api.py` | Import, route registration, `cancel_command` handler |
| `src/exo/master/tests/test_cancel_command.py` | 3 test cases (404, text cancel, image cancel) |
| `docs/api.md` | Document new endpoint + update summary table |

## Design Decisions

- Uses `self._send()` (not raw `command_sender.send()`) — respects API pause state during elections
- Uses `raise HTTPException` — feeds into exo's centralized OpenAI-style error handler
- Returns typed `CancelCommandResponse` — consistent with `CreateInstanceResponse` / `DeleteInstanceResponse` patterns
- `sender.close()` is idempotent so concurrent cancel requests for the same command are harmless

## Test Plan

- [x] `test_cancel_nonexistent_command_returns_404` — verifies 404 with OpenAI error format
- [x] `test_cancel_active_text_generation` — verifies 200, `sender.close()` called, `TaskCancelled` sent with correct `cancelled_command_id`
- [x] `test_cancel_active_image_generation` — same verification for image queue
- [x] `basedpyright` — 0 new errors
- [x] `ruff check` — all checks passed
- [x] `pytest` — 3/3 new tests pass, no regressions

### DIFF
--- a/src/exo/master/api.py
+++ b/src/exo/master/api.py
@@ -571,7 +571,9 @@ class API:
 
     async def cancel_command(self, command_id: CommandId) -> CancelCommandResponse:
         """Cancel an active command by closing its stream and notifying workers."""
-        sender = self._text_generation_queues.get(command_id) or self._image_generation_queues.get(command_id)
+        sender = self._text_generation_queues.get(
+            command_id
+        ) or self._image_generation_queues.get(command_id)
         if sender is None:
             raise HTTPException(
                 status_code=404,


src/exo/master/api.py
```diff
@@ -564,6 +566,24 @@ async def delete_instance(self, instance_id: InstanceId) -> DeleteInstanceRespon
             instance_id=instance_id,
         )
 
+    async def cancel_command(self, command_id: str) -> CancelCommandResponse:
+        """Cancel an active command by closing its stream and notifying workers."""
+        cid = CommandId(command_id)
```
Evanev7:
```suggestion
    async def cancel_command(self, command_id: CommandId) -> CancelCommandResponse:
        """Cancel an active command by closing its stream and notifying workers."""
```

Evanev7: the CommandId's are just strings anyway, we don't need the extra step here

mlpy0:
What did you do @Evanev7

Evanev7:
just rebased onto latest main to try and get it merged, though we're feature freezing now before our next release

mlpy0:
> just rebased onto latest main to try and get it merged, though we're feature freezing now before our next release

Sorry for confusion

Evanev7:
no worries! it's nice to see regular contributors so let me know if i can do anything to keep things clear

mlpy0:
> no worries! it's nice to see regular contributors so let me know if i can do anything to keep things clear

Thank you 🙏🙏

PR #1587, opened 2026-02-22
Add custom prefill for pipeline

rltakashige:
## Motivation

Since we need to do distributed communications between prefill step sizes, the out-of-the-box stream_generate that we currently use prevents pipeline parallel models from doing overlapped computation. While this was technically a regression, this communication is necessary for cancellation, and we will need various distributed communications in the future (e.g. for coordinating batching).

500 lines are for one testing file, so the diffs aren't as bad as they look!

## Changes

Added a special prefill function for pipeline parallel models
Edited the model to handle 
Added a test to verify this new prefill and the original prefill produce identical results
Improved type stubs to remove some type: ignores 

## Why It Works
image

### Manual Testing
Needs more testing, but seems good so far.

### Automated Testing
Passes CI, considerable speedup seen in benchmarks (up to 1.98x) on prefill speed.

Before:
image

After:
image


.mlx_typings/mlx_lm/models/cache.pyi
```diff
@@ -16,7 +16,7 @@ class Cache(Protocol):
         self, keys: mx.array, values: mx.array
     ) -> tuple[mx.array, mx.array]: ...
     @property
-    def state(self) -> tuple[mx.array, mx.array]: ...
```
Evanev7, line 19: were these changes intentional? its inconsistent with the rest of the cache pyi.
its typings at the end of the day, just jumped out as weird

src/exo/worker/engines/mlx/generator/generate.py
```diff
@@ -55,6 +64,128 @@ class PrefillCancelled(BaseException):
     """Raised when prefill is cancelled via the progress callback."""
 
 
+def _has_pipeline_layer(model: Model):
```
Evanev7: could you rename this to _has_pipeline_communication_layer? every layer is a pipeline layer

src/exo/worker/engines/mlx/__init__.py
Evanev7: yay!

src/exo/worker/engines/mlx/generator/generate.py
```diff
@@ -55,6 +64,128 @@ class PrefillCancelled(BaseException):
...
+                    processed += chunk_size
+
+                if distributed_prompt_progress_callback is not None:
+                    distributed_prompt_progress_callback()
+
+                flush_prefill_sends()
+
+                if not is_dummy:
+                    # Only callback when something actually meaningful was computed
+                    prompt_progress_callback(processed, total)
```
Evanev7:
```suggestion
            for i in range(n_leading):
                if distributed_prompt_progress_callback is not None:
                    distributed_prompt_progress_callback()
                flush_prefill_sends()
            for i in range(n_real):
                chunk_size = real_chunk_sizes[i]
                model(
                    prompt[processed : processed + chunk_size][None],
                    cache=_prompt_cache,
                )
                quantize_cache_fn(_prompt_cache)
                processed += chunk_size
                if distributed_prompt_progress_callback is not None:
                    distributed_prompt_progress_callback()
                flush_prefill_sends()
                prompt_progress_callback(processed, total)
            for i in range(n_trailing):
                if distributed_prompt_progress_callback is not None:
                    distributed_prompt_progress_callback()
                flush_prefill_sends()

```

does this feel cleaner?

src/exo/worker/engines/mlx/generator/generate.py
```diff
@@ -94,27 +226,48 @@ def progress_callback(processed: int, total: int) -> None:
         if on_prefill_progress is not None:
             on_prefill_progress(processed, total)
 
+    def combined_progress_callback(processed: int, total: int) -> None:
+        if distributed_prompt_progress_callback is not None:
+            distributed_prompt_progress_callback()
+        progress_callback(processed, total)
+
```
Evanev7, line 236: the callbacks grow 😔

.mlx_typings/mlx_lm/models/cache.pyi
rltakashige: Changes were intentional. I realised those could and should be able to be None. I also wanted to remove some type ignores while I was at it.

src/exo/worker/engines/mlx/generator/generate.py
rltakashige: Yeah makes sense

.mlx_typings/mlx_lm/models/cache.pyi
Evanev7: oh sorry, i should be more specific, i meant removing the mx. prefix

rltakashige: oh damn, I copied it from the bottom of the file which probably didn't have mx. I'll add it back to make the diffs very slightly nicer

Evanev7 (approved):
lets get her tested

PR #1588, opened 2026-02-22
[Fix] Handle Llama 3 instruct special tokens

michaelharrigan:
## Summary
- Llama 3 instruct models emit special tokens (`<|python_tag|>`, `<|eom_id|>`, `<|eot_id|>`, `<|start_header_id|>`, etc.) that leak into chat output as visible text because exo has no awareness of the Llama 3 instruct format
- Three-layer fix:
  1. **EOS tokens**: Register `<|eom_id|>` and `<|eot_id|>` as EOS tokens so generation stops at message/turn boundaries
  2. **Tool call parsing**: New `llama3_tool_parser` that detects `<|python_tag|>` as the tool call start marker and parses the `{"name": ..., "parameters": {...}}` JSON body. Also handles EOS-as-end-marker (since `<|eom_id|>` never appears in text when it's an EOS token) by trying to parse accumulated text when `finish_reason` fires
  3. **Special token stripping**: Final-pass generator that regex-strips any remaining Llama 3 instruct tokens from text output as a safety net

## Files Changed
- `tool_parsers.py` — Added `LLAMA3_SPECIAL_TOKENS`, `llama3_tool_parser`, and `_parse_llama3_tool_calls()`
- `runner.py` — Llama 3 detection via vocab check, EOS token registration, `strip_llama3_special_tokens()` generator, improved `parse_tool_calls()` EOS handling

## Test plan
- [ ] Load a Llama 3.1/3.2/3.3 instruct model via Exo cluster
- [ ] Verify no special tokens appear in regular chat output
- [ ] Verify generation stops cleanly at end-of-turn (no runaway generation)
- [ ] Verify tool calls are parsed correctly when tools are enabled
- [ ] Verify non-Llama models (DeepSeek, GPT-OSS, etc.) are unaffected

### DIFF
--- a/src/exo/worker/runner/llm_inference/runner.py
+++ b/src/exo/worker/runner/llm_inference/runner.py
@@ -1,4 +1,5 @@
 import math
+import re
 import resource
 import time
 from collections.abc import Generator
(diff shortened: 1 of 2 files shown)


michaelharrigan:
Apologies, was supposed to be opened on my forked repo.

PR #1595, opened 2026-02-23
fix isinstance for qwen3Moe

Evanev7:
we were checking if qwen3 had a transformers Qwen3DecoderLayer rather than an mlx Qwen3MoeDecoderLayer causing an assertion error on loading qwen models - this corrects it to the actual layer type


rltakashige:
Thanks!

PR #1509, opened 2026-02-17
Fix instance type mismatch in /instance/previews endpoint

AlexCheema:
## Summary
- Fixes `instance_meta` in preview responses not matching the actual `instance` type
- `place_instance()` overrides single-node placements to `MlxRing`, but the preview was still reporting the originally requested type (e.g., `MlxJaccl`)
- Now derives `instance_meta` from the actual instance object returned (`MlxJacclInstance` vs `MlxRingInstance`) so the response is always consistent

Closes #1426

## Test plan
- [x] `basedpyright` passes with 0 errors
- [x] `ruff check` passes
- [x] `nix fmt` clean (0 changed)
- [x] `pytest` passes (260 passed, 1 skipped)
- [ ] On a single-node setup, call `GET /instance/previews?model_id=<model>` and verify that when `instance_meta` is `MlxRing`, the `instance` key wraps as `MlxRingInstance` (not `MlxJacclInstance`, and vice versa)

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #1600, opened 2026-02-23
fix: error instead of silently overriding instance_meta for single-node placements

AlexCheema:
## Motivation

`place_instance()` was silently overriding `command.instance_meta` from `MlxJaccl` to `MlxRing` when only one node was selected (placement.py:141-142). This caused two problems:

1. **Preview mismatch** (#1426 / PR #1509): The preview loop used the *requested* instance type in its response, but the actual instance created was a different type
2. **Silent substitution**: Users requesting `MlxJaccl` on a single node got `MlxRing` with no feedback — the function should not override user arguments

## Changes

Replaced the silent override with a `ValueError` when `MlxJaccl` is requested but only one node is available.

## Why It Works

- **Previews**: The preview endpoint already wraps `place_instance()` in a try/except — error previews are simply not shown in the dashboard (correct behavior)
- **Actual placements**: Users now get a clear 400 error instead of a silent substitution
- This also fixes the type mismatch reported in #1426, making PR #1509 unnecessary (supersedes it)

### Manual Testing

- On a single-node setup, `GET /instance/previews?model_id=<model>` no longer returns `MlxJaccl` previews (they error out and are filtered)
- Requesting `MlxJaccl` placement on a single node returns a 400 error

### Automated Testing
- All 11 placement tests pass (`uv run pytest src/exo/master/tests/test_placement.py -v`)
- Full test suite: 218 passed, 1 skipped (pre-existing `test_master` Keypair failure unrelated)

### DIFF
--- a/src/exo/master/placement.py
+++ b/src/exo/master/placement.py
@@ -138,8 +138,8 @@ def place_instance(
     instance_id = InstanceId()
     target_instances = dict(deepcopy(current_instances))
 
-    if len(selected_cycle) == 1:
-        command.instance_meta = InstanceMeta.MlxRing
+    if len(selected_cycle) == 1 and command.instance_meta == InstanceMeta.MlxJaccl:
+        raise ValueError("MlxJaccl requires more than one node; only one node was selected")
 
     match command.instance_meta:
         case InstanceMeta.MlxJaccl:


AlexCheema:
Closing — for n=1 all configurations are equivalent, so erroring is wrong. PR #1509 correctly fixes the actual bug (preview type mismatch).

PR #1602, opened 2026-02-23
fix: remove silent instance_meta override for single-node placements

AlexCheema:
## Motivation

`place_instance()` silently overrode `command.instance_meta` from `MlxJaccl` to `MlxRing` for single-node placements. But for n=1, all combinations of sharding (Pipeline/Tensor) and instance type (MlxRing/MlxJaccl) are equivalent and valid — there's no sharding and no inter-node communication. The override was unnecessary and caused preview responses to report the wrong instance type (#1426).

## Changes

Removed the 2-line override in `place_instance()` that mutated `command.instance_meta` for single-node cycles.

## Why It Works

For n=1, MlxJaccl creates a 1x1 device matrix and a single coordinator — functionally identical to MlxRing. By removing the override, users get exactly what they asked for, and the preview type mismatch is fixed at the root cause.

Supersedes #1509.

### Manual Testing

- On a single-node setup, `GET /instance/previews?model_id=<model>` now correctly returns MlxJaccl previews for n=1 with matching instance type

### Automated Testing
- All 11 placement tests pass (`uv run pytest src/exo/master/tests/test_placement.py -v`)
- Full test suite: 218 passed, 1 skipped (pre-existing `test_master` Keypair failure unrelated)


ciaranbor:
No changes?

AlexCheema:
Closing — keeping the N=1 → Pipeline/Ring default. PR #1509 is the correct fix for the preview type mismatch.

PR #1605, opened 2026-02-23
feat: add --trust-remote-code CLI flag

AlexCheema:
## Motivation

Some custom models (e.g. Kimi) require `trust_remote_code=True` to load their tokenizers. Currently the `TRUST_REMOTE_CODE` constant defaults to `True` globally, which is a security risk. Users need an explicit opt-in mechanism that is only available via the CLI — not the API — to prevent remote code execution attacks.

## Changes

- Added `--trust-remote-code` CLI flag to `Args` in `src/exo/main.py`
- Flag sets `EXO_TRUST_REMOTE_CODE=1` env var, inherited by runner subprocesses (same pattern as `--fast-synch`)
- `get_tokenizer()` in `utils_mlx.py` checks the env var and overrides `trust_remote_code` to `True` when set
- Changed default `TRUST_REMOTE_CODE` constant from `True` to `False` in `constants.py`, making remote code execution fully opt-in

## Why It Works

The env-var approach follows the established `--fast-synch` pattern: CLI sets the env var in the main process, runner subprocesses (spawned via `mp.Process`) inherit it. The override is applied at the single point where `trust_remote_code` flows to `load_tokenizer_for_model_id()`. The flag is intentionally CLI-only — not exposed via the API — so remote attackers cannot enable it.

### Manual Testing
- `uv run exo --help` shows the new `--trust-remote-code` flag
- `uv run exo --trust-remote-code` logs a warning about arbitrary code execution
- Without the flag, custom models added via `/models/add` load with `trust_remote_code=False`

### Automated Testing
- `uv run ruff check` — all checks passed
- `uv run pytest` — 222 passed (1 pre-existing failure in Rust bindings unrelated to this change)
- `nix fmt` — applied

### DIFF
--- a/src/exo/worker/engines/mlx/constants.py
+++ b/src/exo/worker/engines/mlx/constants.py
@@ -13,5 +13,6 @@ KV_CACHE_BITS: int | None = None
 
 DEFAULT_TOP_LOGPROBS: int = 5
 
-# Opt-in via --trust-remote-code CLI flag; default is False for security.
-TRUST_REMOTE_CODE: bool = False
+# True for built-in models with known model cards; custom models added via API default to False
+# and can be overridden with the --trust-remote-code CLI flag.
+TRUST_REMOTE_CODE: bool = True


src/exo/worker/engines/mlx/constants.py
```diff
@@ -13,5 +13,6 @@
 
 DEFAULT_TOP_LOGPROBS: int = 5
 
-# TODO: We should really make this opt-in, but Kimi requires trust_remote_code=True
+# True for built-in models with known model cards; custom models added via API default to False
+# and can be overridden with the --trust-remote-code CLI flag.
 TRUST_REMOTE_CODE: bool = True
```
Evanev7, line 18: did you miss this?

PR #1601, opened 2026-02-23
fix: improve text contrast on downloads page

AlexCheema:
## Summary
- Bumps opacity on dark-grey text/icons on the downloads page that were nearly invisible against the dark background
- Informational text (GB sizes, speeds, disk free, model IDs) → full `text-exo-light-gray`
- Interactive icons (delete, resume, retry) → `/70` at rest
- Hover-only elements (download button) → `/60`

## Test plan
- [ ] Open http://localhost:52415/downloads with models downloaded
- [ ] Verify GB downloaded amounts are clearly visible
- [ ] Verify delete trash icons are visible (not just on hover)
- [ ] Verify download speed text is readable
- [ ] Verify "paused" labels and resume buttons are visible
- [ ] Verify "X GB free" disk labels in column headers are readable

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #1609, opened 2026-02-24
fix: improve text contrast on HOME and DOWNLOADS nav links

AlexCheema:
## Motivation

Follow-up to #1601 (downloads page contrast fix). The HOME and DOWNLOADS navigation links in the top-right header use `text-exo-light-gray` (`oklch(0.6 0 0)`) which is too dim against the dark header background.

## Changes

Changed both nav links in `HeaderNav.svelte` from `text-exo-light-gray` to `text-white/70` for better visibility. Hover state (`text-exo-yellow`) is unchanged.

## Why It Works

`text-white/70` provides noticeably better contrast against `bg-exo-dark-gray` while still looking subdued relative to the yellow accent color on hover. This is consistent with the approach used in #1601.

### Manual Testing
- Verified both links are clearly readable on the home page and downloads page
- Hover state still transitions to yellow as expected

### Automated Testing
- Dashboard builds successfully

### DIFF
--- a/bench/exo_bench.py
+++ b/bench/exo_bench.py
@@ -75,7 +75,7 @@ def load_tokenizer_for_bench(model_id: str) -> Any:
         model_path = Path(
             snapshot_download(
                 model_id,
-                allow_patterns=["*.json", "*.py", "*.tiktoken"],
+                allow_patterns=["*.json", "*.py", "*.tiktoken", "*.model"],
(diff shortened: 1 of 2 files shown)


AlexCheema:
## Screenshots

### Home page
Home page nav links

### Downloads page
Downloads page nav links

PR #1615, opened 2026-02-24
fix: replace flaky internet checks with explicit offline mode

AlexCheema:
## Motivation

The socket-based internet connectivity probes (connecting to 1.1.1.1/8.8.8.8/1.0.0.1 every 10 seconds) are flaky — they produce false negatives on some networks/ISPs, causing downloads to silently fail or get stuck. Instead of dynamically detecting connectivity, this replaces it with an explicit offline mode that the user opts into.

## Changes

- **Removed internet check infrastructure**: Deleted `_test_internet_connection()` (socket probes) and `_check_internet_connection()` (10s polling loop) from `DownloadCoordinator`
- **Renamed `internet_connection` → `offline`** on `ShardDownloader` base class and all subclasses (`SingletonShardDownloader`, `CachedShardDownloader`, `ResumableShardDownloader`)
- **Removed `on_connection_lost` callbacks** from download calls — no longer needed without dynamic connectivity detection
- **Added `EXO_OFFLINE` env var support** in `constants.py` and `Args` default in `main.py`
- **Added Offline Mode toggle** to macOS app Settings → General tab, which sets `EXO_OFFLINE=true` in the process environment and restarts

## Why It Works

The download behavior (`skip_internet` conditionals in `download_utils.py`) is unchanged — we just changed what drives it. Instead of a flaky socket probe setting `internet_connection=True/False`, the user explicitly sets offline mode via:
1. `--offline` CLI flag
2. `EXO_OFFLINE=true` environment variable
3. macOS app Settings → General → Offline Mode toggle

This is more reliable and predictable. Users on flaky networks no longer get intermittent download failures.

### Manual Testing
- `uv run exo --offline` starts without internet checks, rejects downloads for unavailable models
- `EXO_OFFLINE=true uv run exo` behaves identically
- macOS app Settings toggle persists across restarts and passes env var to the exo process

### Automated Testing
- All existing tests pass (222 passed), including 8 offline-mode specific tests
- `uv run ruff check` — all checks passed
- `nix fmt` — 0 files changed

### DIFF
--- a/src/exo/download/coordinator.py
+++ b/src/exo/download/coordinator.py
@@ -1,5 +1,4 @@
 import asyncio
-import socket
 from dataclasses import dataclass, field
 from random import random
 
@@ -73,8 +72,6 @@ class DownloadCoordinator:
 
     def __post_init__(self) -> None:
         self.event_sender, self.event_receiver = channel[Event]()
-        if self.offline:
-            self.shard_downloader.set_internet_connection(False)
         self.shard_downloader.on_progress(self._download_progress_callback)
 
     def _model_dir(self, model_id: ModelId) -> str:
@@ -123,8 +120,6 @@ class DownloadCoordinator:
         logger.info(
             f"Starting DownloadCoordinator{' (offline mode)' if self.offline else ''}"
         )
-        if not self.offline:
-            self._test_internet_connection()
         try:
             async with self._tg as tg:
                 tg.start_soon(self._command_processor)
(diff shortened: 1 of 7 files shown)


src/exo/download/coordinator.py
JakeHillion, line 57: I believe this field is dead code now

AlexCheema: It's used on line 229 of `coordinator.py`:

```python3
        if self.offline:
            logger.warning(
                f"Offline mode: model {model_id} is not fully available locally, cannot download"
            )
            failed = DownloadFailed(
                shard_metadata=shard,
                node_id=self.node_id,
                error_message=f"Model files not found locally in offline mode: {model_id}",
                model_directory=self._model_dir(model_id),
            )
            self.download_status[model_id] = failed
            await self.event_sender.send(NodeDownloadProgress(download_progress=failed))
            return
```

PR #1617, opened 2026-02-25
remove python multipart

Evanev7:
we aren't using this library and it's pre 0.1.


ciaranbor:
I have a slight recollection of this being a runtime dependency of fastapi for UploadFile to work (for image editing). Let me check if editing still works without it

ciaranbor:
Ok yeah we do need it

PR #1619, opened 2026-02-25
persist node ids in .cache

Evanev7:
brings back EXO_CACHE_HOME as always ~/.cache/exo/, and store the node id in there, to prevent the node id from being persisted when copying ~/.exo


Evanev7:
well, outside of the models... but yes. this is pure-exo state

PR #1620, opened 2026-02-25
Address Mac Mini pipeline GPU timeouts

rltakashige:
## Motivation
Users were reporting GPU timeout errors on Mac Minis, which we never saw on testing with Mac Studios. It also seems to only happen with large models.

## Changes
Eval specific distributed operations.

## Why It Works

As I wrote in a Slack message:
Basically, prefill is too slow for pipeline communications. If there are both communications and GPU operations as part of an mlx graph, the communications become subject to the GPU's 5 second command buffer timeout.

For normal generation, I added evals to the communications (only during prefill, as it slows down decode) to do this, fixing GPU timeouts.

But we don't do this during warmup, as the prompt is absolutely tiny. This is still too slow on an M4 Pro on some models that it causes a GPU timeout during warmup...

----------------------
This was one of the issues. However, there is another issue:

mx.all_gather sometimes reads stale data with FAST_SYNCH enabled. I'm still investigating the root cause, but the code as it is now works on Mac Minis.

### Manual Testing
image

image

### Automated Testing
Needs a bunch on mac minis

### DIFF
--- a/tests/auto_bench.sh
+++ b/tests/auto_bench.sh
@@ -17,6 +17,13 @@ git branch -r --contains "$commit" | grep -qE '^\s*origin/' || {
   exit 1
 }
 hosts=("$@")
+
+for host; do
+  ssh -T -o BatchMode=yes -o ServerAliveInterval=30 "$host@$host" \
+    "EXO_LIBP2P_NAMESPACE=$commit /nix/var/nix/profiles/default/bin/nix build github:exo-explore/exo/$commit" &
+done
+wait
+
 cleanup() {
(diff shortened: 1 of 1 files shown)


rltakashige:
The MLX uv lock changes should be inconsequential, I think. Need to check.

rltakashige:
I think this is good to go now. It solves another issue faced by @AlexCheema where running pipeline RDMA + tensor RDMA. I'd assume that was a problem with pipeline being slowed down by tensor, causing a similar situation to the Mac Minis see naturally.

PR #1628, opened 2026-02-26
Support trace deletion in dashboard

ciaranbor:
## Motivation

Enable using the dashboard to delete traces

## Changes

  - Backend (src/exo/master/api.py): Added POST /v1/traces/delete endpoint that deletes trace files by task ID, with path traversal protection
  - API types (src/exo/shared/types/api.py): Added DeleteTracesRequest and DeleteTracesResponse models
  - Dashboard store (app.svelte.ts): Added deleteTraces() method
  - Traces page (+page.svelte): Added multi-select UI (click to select, select all/deselect all) with a bulk delete button and confirmation dialog

## Why It Works

  - Uses existing _get_trace_path helper (now hardened against path traversal) to resolve and delete trace files
  - Dashboard selection state is reactive via Svelte 5 runes; deleted traces refresh the list automatically

### Manual Testing

  - Select individual traces, select all, delete with confirmation dialog
  - Verify traces are removed from disk and list refreshes

### DIFF
--- a/src/exo/master/api.py
+++ b/src/exo/master/api.py
@@ -81,6 +81,8 @@ from exo.shared.types.api import (
     CreateInstanceResponse,
     DeleteDownloadResponse,
     DeleteInstanceResponse,
+    DeleteTracesRequest,
+    DeleteTracesResponse,
     ErrorInfo,
     ErrorResponse,
     FinishReason,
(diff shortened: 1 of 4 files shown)


src/exo/master/api.py
```diff
@@ -1708,7 +1722,10 @@ async def delete_download(
         return DeleteDownloadResponse(command_id=command.command_id)
 
     def _get_trace_path(self, task_id: str) -> Path:
-        return EXO_TRACING_CACHE_DIR / f"trace_{task_id}.json"
+        trace_path = EXO_TRACING_CACHE_DIR / f"trace_{task_id}.json"
+        if not trace_path.resolve().is_relative_to(EXO_TRACING_CACHE_DIR.resolve()):
+            raise HTTPException(status_code=400, detail=f"Invalid task ID: {task_id}")
+        return trace_path
```
Evanev7: nit: could be a static method

PR #1632, opened 2026-02-27
Refactor runner for implementing batching

rltakashige:
## Motivation

Batching will require us to send tasks concurrently and queue them up. Our current infrastructure cannot handle that all. This PR gets us closer to this by allowing multiple tasks to be sent in parallel and then queuing up tasks.

## Changes

Change Plan logic
Make runner main into a class
Add a "BatchGenerator" to which tasks can be submitted (although tasks are handled sequentially) and sent back through an MpSender.
Refactor runner to accept tasks during generation
Keep the generator threading
Separate the runner into several files for better readability

### Manual Testing
Tested manually, needs a lot more automated testing. Cancellation still works on a single device. Needs checking on multiple devices.

### DIFF
--- a/src/exo/worker/tests/unittests/test_runner/test_event_ordering.py
+++ b/src/exo/worker/tests/unittests/test_runner/test_event_ordering.py
@@ -119,6 +119,7 @@ def patch_out_mlx(monkeypatch: pytest.MonkeyPatch):
     monkeypatch.setattr(mlx_batch_generator, "warmup_inference", make_nothin(1))
     monkeypatch.setattr(mlx_batch_generator, "_check_for_debug_prompts", nothin)
     monkeypatch.setattr(mlx_batch_generator, "mx_any", make_nothin(False))
+
     def fake_all_gather(
         tasks: list[TextGeneration], group: object
     ) -> tuple[list[TextGeneration], list[TextGeneration]]:


src/exo/routing/router.py
```diff
@@ -258,6 +258,6 @@ def lock_path(path: str | bytes | PathLike[str] | PathLike[bytes]) -> Path:
 
         # if no valid credentials, create new ones and persist
         with open(path, "w+b") as f:
-            keypair = Keypair.generate_ed25519()
+            keypair = Keypair.generate()
```
JakeHillion, line 261: I'm quite confused how this worked before, can we take a closer look?

JakeHillion: It's unreachable as discussed, removed in #1633

src/exo/utils/channels.py
```diff
@@ -282,6 +283,54 @@ def __getstate__(self):
...
 
 
+class NonBlockingGenerator[T](Generator[T | None, None, None]):
+    def __init__(self, source: MpReceiver[T] | Generator[T | None, None, None]) -> None:
+        self._receiver: MpReceiver[T] | None = None
+        self._inner: Generator[T | None, None, None] | None = None
+        if isinstance(source, MpReceiver):
+            self._receiver = source
+        else:
+            self._inner = source
```
JakeHillion: I think it would make more sense to keep the constructor always setting one field that's a Union of the two (the arg type) and discriminate with `isinstance` where we use it. The current logic is a little confusing as we set `self._inner` second in the constructor but use it first. The behaviour would be clearer if we discriminate based on type when we use it.

src/exo/worker/runner/llm_inference/runner.py
```diff
@@ -63,743 +53,431 @@
...
-    with task_receiver as tasks:
-        for task in tasks:
-            if task.task_id in seen:
-                logger.warning("repeat task - potential error")
-            seen.add(task.task_id)
-            cancelled_tasks.discard(TaskId("CANCEL_CURRENT_TASK"))
-            event_sender.send(
-                TaskStatusUpdated(task_id=task.task_id, task_status=TaskStatus.Running)
+ALL_TASKS_COMPLETE_CODE = 0
+SHUTDOWN_CODE = 1
```
Evanev7: use an enum!

src/exo/worker/runner/llm_inference/batch_generator.py
Evanev7, line 1: was this intended to be included in this PR?

src/exo/worker/runner/llm_inference/runner.py
```diff
@@ -63,743 +53,431 @@
...
+            mlx_generator = parse_gpt_oss(mlx_generator)
+        elif (
+            isinstance(self.inference_model, DeepseekV32Model)
+            and "deepseek" in self.model_id.normalize().lower()
+        ):
+            mlx_generator = parse_deepseek_v32(mlx_generator)
+        elif self.tool_parser:
+            mlx_generator = parse_tool_calls(mlx_generator, self.tool_parser)
+
+        return NonBlockingGenerator(mlx_generator)
```
Evanev7: this i don't understand - why not return the top level generator object here?

src/exo/utils/channels.py
```diff
@@ -282,6 +283,54 @@ def __getstate__(self):
         return d
 
 
+class NonBlockingGenerator[T](Generator[T | None, None, None]):
```
Evanev7: ahh monads my old friend. this is (i reckon) not quite the right abstraction here. in its current iteration, id suggest using just the receiver and letting the WouldBlock exception bubble up so we don't need to do this T | None dance all the way through the pipeline.

src/exo/worker/runner/bootstrap.py
```diff
@@ -33,10 +33,15 @@ def entrypoint(
     try:
         if bound_instance.is_image_model:
             from exo.worker.runner.image_models.runner import main
+
+            main(bound_instance, event_sender, task_receiver, cancel_receiver)
```
Evanev7: a change of this scale should be reflected in the image runner.

src/exo/worker/plan.py
```diff
@@ -297,10 +297,10 @@ def _pending_tasks(
             # the task status _should_ be set to completed by the LAST runner
             # it is currently set by the first
             # this is definitely a hack
-            if task.task_id in runner.completed:
+            if task.task_id in runner.completed or task.task_id in runner.pending:
                 continue
 
-            if isinstance(runner.status, RunnerReady) and all(
+            if isinstance(runner.status, (RunnerReady, RunnerRunning)) and all(
```
Evanev7, line 303: these changes are just to allow submitting a task during batching i presume?

src/exo/worker/runner/runner_supervisor.py
```diff
@@ -172,7 +172,7 @@ async def _forward_events(self):
                     if isinstance(event, RunnerStatusUpdated):
                         self.status = event.runner_status
                     if isinstance(event, TaskAcknowledged):
-                        self.pending.pop(event.task_id).set()
+                        self.pending[event.task_id].set()
```
Evanev7: this is a very significant change to the meaning of "pending" in the supervisor, that I'm not a huge fan of. if you want a union of pending and active there should be other ways to implement it.

src/exo/worker/runner/llm_inference/batch_generator.py
Evanev7: (the batch generation code that is)

src/exo/worker/runner/llm_inference/runner.py
```diff
@@ -63,743 +53,431 @@
...
+                        tokenizer=self.tokenizer,
+                        group=self.group,
+                        kv_prefix_cache=self.kv_prefix_cache,
+                        model_id=self.model_id,
+                        device_rank=self.device_rank,
+                        cancel_receiver=self.cancel_receiver,
+                        cancelled_tasks=self.cancelled_tasks,
+                        event_sender=self.event_sender,
+                        check_for_cancel_every=self.check_for_cancel_every,
+                    )
```
Evanev7: if you are going to include a dedicated batch generator task, it should either take ownership of these values or be independent of them. really don't like this sharing.

src/exo/worker/runner/bootstrap.py
rltakashige: I think this would make the diff a bit unmanageable. I'll add another pr on top that does this.

src/exo/worker/runner/llm_inference/batch_generator.py
rltakashige: This isn't batch generation. We just queue up a batch of tasks and process sequentially. The resulting behaviour is a no-op, but this gives it the structure of batch generation that we can replace with a true implementation.

src/exo/worker/runner/bootstrap.py
Evanev7: playing devils advocate a little but;
- im not jake; i dont care about diff size
- i do not want the two runners to fall out of sync at the current moment, i want a clean break from old style main to new style class
- this is already out of scope of what a single pr could be (i.e. the batch generator interface)

src/exo/worker/runner/runner_supervisor.py
rltakashige: Will think on this

src/exo/utils/channels.py
rltakashige: Have replied similarly in a different comment, but I want to be able to use this like mlx generate that can be composed with the model output parsers with the option for None when a result isn't available.

src/exo/worker/runner/llm_inference/batch_generator.py
Evanev7: it is a batch generator, it's just a bad one.

src/exo/utils/channels.py
Evanev7: which other comment? and, i suppose if you're set on that api we should make the split explicit; have one class throw WouldBlock and have an outer wrapper that catches WouldBlock and converts it to None if you don't want to adjust the current generator mapping functions.

src/exo/worker/runner/llm_inference/batch_generator.py
rltakashige: Since we're going to have both implementations in the next PR, I've renamed this to a SequentialGenerator that implements an Inference Generator ABC. (Not sure about that naming but I couldn't come up with a better one)

src/exo/worker/runner/llm_inference/runner.py
rltakashige: Wanted the syntactic sugar of try_receive but it's not very useful except for a little clarity on what None means. Changed it

src/exo/worker/runner/llm_inference/runner.py
rltakashige: Since warmup requires a lot of these fields, I'm not sure how I should go about doing this. I have gotten rid of cancelled tasks as that can be entirely local to the generator.

src/exo/worker/runner/runner_supervisor.py
rltakashige: Added an in progress set instead

src/exo/worker/runner/bootstrap.py
rltakashige: I thought the diffs would be a lot worse but that file isn't too big... Made it a separate commit so I can extract if it ever feels necessary

src/exo/worker/runner/image_models/runner.py
Evanev7, line 1: thanks for implementing this as well

PR #1633, opened 2026-02-27
router: remove unreachable code in get_node_id_keypair

JakeHillion:
This code is unreachable, and has already rotted a bit. For example, the generate_ed25519 method is gone, and we use generate elsewhere in the live code.

Remove it to avoid confusion.

Test plan:
- It's dead.


rltakashige:
@JakeHillion Never mind, this is unnecessary after #1619

Issue #1630, opened 2026-02-27
[BUG] BrokenResourceError crash in _command_processor after event router (#1572) — not fixed by #1629


Wysie:
## Describe the bug

Load any model. Exo crashes on startup with anyio.BrokenResourceError in Master._command_processor when trying to send an InstanceCreated event. The receive end of the memory stream is already closed (open_receive_channels=0) when the sender tries to write.

## To Reproduce

1. git checkout 639243a (or any commit after #1572)
2. uv sync, npm install, etc.
3. Start Exo on a 2-node cluster
4. Crash occurs shortly after startup during instance creation

| Commit | Status |
| --- | --- |
| `db73c4f` (move messaging into rust #1549) | ✅ **Works** |
| `639243a` (event router #1572) | ❌ **Crashes** |
| `db36bd5` (Add custom prefill for pipeline #1587) | ❌ Crashes |
| `152a27e` (Fix pipeline mismatched send after 1587 #1629) | ❌ Still crashes |

## Expected behavior

Model to load successfully.

## Actual behavior

Crashes on loading models.

## Environment

- macOS Version: 26.3
- EXO Version: Commit 639243aa0956b6f2410a08c9c2001285b26fbdc5 onwards
- Hardware:
  - Device 1: Mac Studio M4 Max, 128GB RAM
  - Device 2: Mac Studio M4 Max, 128GB RAM
  - Additional devices:
- Interconnection:
  - Thunderbolt 5 Cable

## Additional context

I've reverted to db73c4f on both machines and things are working well.

Evanev7:
can't actually reproduce this currently, do you have a log of any kind? our errors tend to be quite deeply nested so it may be a secondary crash occurring here

wysie:
@Evanev7 Please see the following log. It happens if the first thing I do is to load a model that needs to use 2 nodes (mlx-community/Qwen3.5-122B-A10B-8bit on 2 x Mac Studio M4 Max 128GB). However, I just tried loading a model on just 1 node first, and then loading mlx-community/Qwen3.5-122B-A10B-8bit across 2 nodes and it seems to work. Doing more testing and will update if I figure out anything.

```
[...]

PR #1637, opened 2026-02-28
fix: replace Master event_sender after EventRouter recreation (#1630)

Wysie:
## Motivation
Fix for https://github.com/exo-explore/exo/issues/1630.
When loading a model that spans 2 nodes as the first action after startup, Exo crashes with anyio.BrokenResourceError in Master._command_processor. The receive end of the Master's event_sender channel is already closed (open_receive_channels=0 ) when the Master tries to send an InstanceCreated event.

## Changes
In _elect_loop, when is_new_master is True and the EventRouter is replaced, the Master's event_sender is now swapped in-place with a fresh sender from the new EventRouter. This mirrors the existing pattern where the Worker and DownloadCoordinator are already recreated with new channels from the new router.

## Why It Works
When is_new_master fires (which happens on every startup when a second node joins), the old EventRouter is shut down and a new one is created. Shutting down the old router cancels its _ingest coroutine, which closes the receive end of every channel created by event_router.sender(). The Worker and DownloadCoordinator are already recreated with senders from the new router, but the Master was not — its event_sender still pointed to the dead channel from the old router.
This patch replaces self.master.event_sender with a new sender from the new EventRouter, so events flow through a live channel. We swap in-place rather than recreating the Master to preserve its DiskEventLog, internal state, and running tasks.
The reason loading a single-node model first appeared to work around the issue is a timing race: if the model loads fast enough, the Master sends its events before the election fires and kills the old channel. With a 2-node model, placement and RDMA setup take longer, so the election completes first and the channel is already dead by the time the Master tries to send.

### Manual Testing

2 x Mac Studio M4 Max 128GB, RDMA

Loaded a model that requires 2 nodes and it works properly now.

### DIFF
--- a/src/exo/main.py
+++ b/src/exo/main.py
@@ -176,6 +176,17 @@ class Node:
                 # - Shutdown and re-create the worker
                 # - Shut down and re-create the API
 
+                if result.is_new_master:
+                    await anyio.sleep(0)
+                    self.event_router.shutdown()
+                    self.event_router = EventRouter(
+                        result.session_id,
+                        self.router.sender(topics.COMMANDS),
+                        self.router.receiver(topics.GLOBAL_EVENTS),
+                        self.router.sender(topics.LOCAL_EVENTS),
+                    )
+                    self._tg.start_soon(self.event_router.run)
+
                 if (
                     result.session_id.master_node_id == self.node_id
                     and self.master is not None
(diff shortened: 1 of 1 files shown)


Evanev7:
oh i see whats happened here and thanks so much for pursuing it, and frankly im not sure why this didn't come up when i was trying to recreate the issue - when a node is elected, it creates its senders for the master before the new event router is created. could i ask you to move the code that recreates the event router before the code that creates the master? it should be first in this block instead of after the master. (rather than updating the master afterwards)

wysie:
@Evanev7 Done, thanks for the suggestion!

PR #1642, opened 2026-03-02
Implement continuous batching

### LINKED ISSUE #1020: [FEATURE] Concurrent inference / continuous batching.

rltakashige:
## Motivation

Following the changes made in #1632 !
Closes #1020

### DIFF
--- a/src/exo/worker/runner/llm_inference/model_output_parsers.py
+++ b/src/exo/worker/runner/llm_inference/model_output_parsers.py
@@ -351,7 +351,7 @@ def parse_tool_calls(
         if response.text.endswith(tool_parser.end_parsing):
             # parse the actual tool calls from the tool call text
             combined = "".join(tool_call_text_parts)
-            parsed = tool_parser.parse(combined.strip(), tools=None)
+            parsed = tool_parser.parse(combined.strip(), tools=tools)
             logger.info(f"parsed {tool_call_text_parts=} into {parsed=}")
             in_tool_call = False
             tool_call_text_parts = []


Evanev7:
ready for review? or not quite yet

src/exo/worker/engines/mlx/generator/batch_generate.py
```diff
@@ -0,0 +1,361 @@
...
+                    f"KV cache hit: {prefix_hit_length}/{len(all_prompt_tokens)} tokens "
+                    f"cached ({100 * prefix_hit_length / len(all_prompt_tokens):.1f}%)"
+                )
+                prompt_tokens = remaining_tokens
+            else:
+                cache = make_kv_cache(self.model)
+        else:
+            cache = make_kv_cache(self.model)
+
+        seed = task_params.seed or 42
```
Evanev7: minor: should use explicit None check - this will override 0 with 42.

src/exo/worker/engines/mlx/generator/batch_generate.py
```diff
@@ -0,0 +1,361 @@
...
+                all_prompt_tokens,
+                list(cache),
+                cache_snapshots,
+                prefix_hit_length,
+                matched_index,
+            )
+
+        last_tokens = prompt_tokens[-2:]
+
+        logits_processors: list[Any] = []
```
Evanev7: i believe we have some repetition penalty logits processors? assuming that merged, this should presumably duplicate that logic (or maybe a single make_logits_processors idk)

src/exo/worker/engines/mlx/generator/batch_generate.py
```diff
@@ -0,0 +1,361 @@
...
+        if is_bench:
+            eos_ids = eos_ids_from_tokenizer(self.tokenizer)
+            logits_processors = [ban_token_ids(eos_ids)]
+
+        max_tokens = task_params.max_output_tokens or MAX_TOKENS
+
+        uids = self._mlx_gen.insert(
```
Evanev7: can this ever return multiple uids? we should guard that case

Evanev7: reading further this seems to be for multiple insertion - assuming we have no interest in multiple insertion, we should just assert it's a single uid.

src/exo/worker/engines/mlx/generator/batch_generate.py
```diff
@@ -0,0 +1,361 @@
...
+        )
+
+        return uids[0]
+
+    def step(self) -> list[tuple[int, GenerationResponse]]:
+        if not self.has_work:
+            return []
+
+        responses = self._mlx_gen.next()
+        mx.clear_cache()
```
Evanev7: this clear_cache could use a comment imo

src/exo/worker/engines/mlx/generator/batch_generate.py
```diff
@@ -0,0 +1,361 @@
...
+
+        results: list[tuple[int, GenerationResponse]] = []
+
+        for response in responses:
+            if response.uid not in self._active_tasks:
```
Evanev7, line 223: feels like an error we should report

src/exo/worker/engines/mlx/generator/batch_generate.py
```diff
@@ -0,0 +1,361 @@
...
+                    prompt_tokens=total_prompt_tokens,
+                    completion_tokens=state.completion_tokens,
+                    total_tokens=total_prompt_tokens + state.completion_tokens,
+                    prompt_tokens_details=PromptTokensDetails(
+                        cached_tokens=state.prefix_hit_length
+                    ),
+                    completion_tokens_details=CompletionTokensDetails(
+                        reasoning_tokens=state.reasoning_tokens
+                    ),
+                )
```
Evanev7, line 313: how realistic is it to extract this as a function of state & mlx_stats? seems like that's all this requires.

src/exo/worker/engines/mlx/generator/batch_generate.py
```diff
@@ -0,0 +1,361 @@
...
+                        stats=stats,
+                        usage=usage,
+                    ),
+                )
+            )
+
+            if is_done:
+                del self._active_tasks[response.uid]
+            elif max_stop_len > 0 and len(state.accumulated_text) > max_stop_len:
+                state.accumulated_text = state.accumulated_text[-max_stop_len:]
```
Evanev7: if this is the only purpose of accumulated_text (scanning for stop sequences) could we rename it to something like `potential_stop_sequence_text`?

src/exo/worker/engines/mlx/utils_mlx.py
```diff
@@ -750,3 +751,55 @@ def _parse_single_tool(text: str) -> dict[str, Any]:
...
+        return [
+            int.from_bytes(utf8_task_id[i : i + 1]) for i in range(len(utf8_task_id))
+        ]
+
+    def decode_task_id(encoded_task_id: list[int]) -> TaskId:
+        return TaskId(
+            bytes.decode(b"".join((x).to_bytes(length=1) for x in encoded_task_id))
+        )
+
+    uuid_byte_length = 36
```
Evanev7, line 771: this is a little spooky imo - can we assert the taskid is of appropriate length?

src/exo/worker/engines/mlx/utils_mlx.py
```diff
@@ -750,3 +751,55 @@ def _parse_single_tool(text: str) -> dict[str, Any]:
...
+        .tolist(),
+    )
+    all_task_ids: list[list[TaskId]] = [
+        [decode_task_id(encoded_task_id) for encoded_task_id in rank_tasks[:count]]
+        for rank_tasks, count in zip(gathered, all_counts, strict=True)
+    ]
+
+    agreed_ids: set[TaskId] = set(all_task_ids[0])
+    for rank_tasks in all_task_ids[1:]:
+        agreed_ids &= set(rank_tasks)
```
Evanev7: do we prefer

agreed_ids = set[TaskId].intersection(*(set(tids) for tids in all_task_ids)?

src/exo/worker/runner/llm_inference/batch_generator.py
```diff
@@ -294,53 +279,212 @@ def close(self) -> None:
...
+        agreed, different = mx_all_gather_tasks(self._maybe_queue, self.group)
+        self._queue.extend(task for task in self._maybe_queue if task in agreed)
+        self._maybe_queue = [task for task in self._maybe_queue if task in different]
+
+    def step(
+        self,
+    ) -> Iterable[
+        tuple[TaskId, GenerationResponse | ToolCallResponse | Cancelled | Finished]
+    ]:
+        self.agree_on_tasks()
```
Evanev7: potential todo - should we merge submit() and step()? if we only ever actually submit() on step(), it could make sense to shrink the interface to step(new_tasks) -> list[Current Results] - though the connection between input and output is not obvious... i've talked myself out of it. potential future refactor could be to make submit(task) -> Return channel though.

src/exo/worker/runner/llm_inference/batch_generator.py
```diff
@@ -294,53 +279,212 @@ def close(self) -> None:
...
+        int,
+        tuple[
+            TextGeneration,
+            GeneratorQueue[GenerationResponse],
+            Generator[GenerationResponse | ToolCallResponse | None],
+        ],
+    ] = field(default_factory=dict, init=False)
+
+    def __post_init__(self) -> None:
+        self._mlx_gen = ExoBatchGenerator(
```
Evanev7, line 334: _mlx_gen -> _exo_gen

src/exo/worker/runner/llm_inference/batch_generator.py
```diff
@@ -294,53 +279,212 @@ def close(self) -> None:
...
+    ) -> Iterable[
+        tuple[TaskId, GenerationResponse | ToolCallResponse | Cancelled | Finished]
+    ]:
+        self.agree_on_tasks()
+
+        # Submit any queued tasks to the engine
+        while self._queue and len(self._active_tasks) < EXO_MAX_CONCURRENT_REQUESTS:
+            task = self._queue.popleft()
+            try:
+                uid = self._build_generator(task)
```
Evanev7: it is not clear that both _build_generator and _mlx_gen.submit **run prefill immediately** - i think we should keep behaviour as is but change the interface slightly

src/exo/worker/runner/llm_inference/batch_generator.py
```diff
@@ -294,53 +279,212 @@ def close(self) -> None:
...
+                    self.tool_parser,
+                    self.tokenizer,
+                    type(self.model),
+                    self.model_id,
+                    task.task_params.tools,
+                )
+            self._active_tasks[uid] = (task, queue, output_generator)
+
+        if not self._mlx_gen.has_work:
+            return self._drain_cancellations()
```
Evanev7: similarly "drain cancellations" imo implies removing a cancellation rather than removing it's corresponding task

src/exo/worker/runner/llm_inference/batch_generator.py
```diff
@@ -294,53 +279,212 @@ def close(self) -> None:
...
+
+        results = self._mlx_gen.step()
+
+        output: list[
+            tuple[TaskId, GenerationResponse | ToolCallResponse | Cancelled | Finished]
+        ] = []
+        for uid, response in results:
+            if uid not in self._active_tasks:
+                # should we error here?
```
Evanev7, line 424: ref comment and review comment above, i think at least a log is due here

src/exo/worker/runner/llm_inference/runner.py
```diff
@@ -392,25 +377,57 @@ class Builder:
...
     inference_model: Model | None = None
     tokenizer: TokenizerWrapper | None = None
-    tool_parser: ToolParser | None = None
     group: mx.distributed.Group | None = None
-    kv_prefix_cache: KVPrefixCache | None = None
 
     def build(
         self,
     ) -> InferenceGenerator:
+        import os
```
Evanev7: no reason to import late here

Evanev7 (approved):
exciting!!! these are mostly stylistic changes with one or two minor correctness things we were probably doing wrong before anyway.

!!!!! continuous batching !!!!!

src/exo/worker/engines/mlx/generator/batch_generate.py
rltakashige: I make assumptions that uids[0] is the real uid anyway, so will assert that.

src/exo/worker/engines/mlx/generator/batch_generate.py
rltakashige: If we were to do something like this, I'd want to make both mlx_generate and this batch generate to be more similar in structure. But I think we can address this in in the future.

PR #1643, opened 2026-03-02
fix: download coordinator and pipeline parallel fixes

icantrainingteam-ctrl:
## Summary
- **Skip auto-resume for already-complete downloads**: `_emit_existing_download_progress` unconditionally called `_start_download_task` for any download with bytes on disk, including already-complete ones. This overwrote `DownloadCompleted` with `DownloadOngoing` every 60s, preventing the worker planner from issuing `LoadModel` (which requires `DownloadCompleted` on all nodes).
- **Drain fire-and-forget progress tasks before computing final download status**: Ensures all in-flight progress callbacks are awaited so `file_progress` reflects the true "complete" state before the final summary is computed.
- **Auto-resume stalled downloads**: Allow `DownloadFailed` models to be retried instead of permanently skipping them.
- **Handle ArraysCache in pipeline/tensor parallel**: SSM layers in hybrid models (e.g. Qwen3-Next) use `ArraysCache` which lacks a `.keys` attribute. Added `hasattr` guards with fallback to `cache[0]`.
- **Set pipeline prefill mode during warmup**: 3-node pipeline parallel deadlocks during warmup because `all_gather` requires all ranks simultaneously during sequential prefill. Now sets prefill mode so `send/recv` is used instead.
- **Track completed-then-cancelled tasks in runner supervisor**: Late cancel signals for already-completed tasks were silently dropped. Now tracks them so downstream logic stays consistent.

## Test plan
- [x] Verified on a 3-node (32GB + 64GB Mac Studio) pipeline cluster running `mlx-community/Llama-3.3-70B-Instruct-4bit`
- [x] Before fix: instance stuck at `RunnerConnected` indefinitely, download showing `DownloadOngoing` at 100%
- [x] After fix: all nodes report `DownloadCompleted`, model progresses through `CreateRunner` → `ConnectToGroup` → `LoadModel` → `StartWarmup` successfully


src/exo/download/coordinator.py
```diff
@@ -140,9 +209,9 @@ async def _start_download(self, shard: ShardMetadata) -> None:
         # Check if already downloading, complete, or recently failed
```
Evanev7, line 209: can you update this comment?

src/exo/download/download_utils.py
Evanev7, line 1: nice

src/exo/worker/engines/mlx/auto_parallel.py
Evanev7, line 1: @rltakashige can you take a look at these? looks ok to me but it's your code

src/exo/worker/runner/llm_inference/runner.py
```diff
@@ -274,28 +273,31 @@ def on_layer_loaded(layers_loaded: int, total: int) -> None:
...
+                            logger.warning(f"[PREFILL DEBUG] cancel_receiver.collect() returned: {new_cancels}")
+                        cancelled_tasks.update(new_cancels)
                         want_to_cancel = (_task_id in cancelled_tasks) or (
                             TaskId("CANCEL_CURRENT_TASK") in cancelled_tasks
                         )
-                        if mx_any(want_to_cancel, _group):
+                        logger.info(f"[PREFILL DEBUG] rank={device_rank} task={_task_id[:8]} want_to_cancel={want_to_cancel} cancelled_tasks={cancelled_tasks} processed={processed}/{total}")
+                        any_cancel = mx_any(want_to_cancel, _group)
+                        logger.info(f"[PREFILL DEBUG] rank={device_rank} mx_any result={any_cancel}")
+                        if any_cancel:
```
Evanev7, line 300: switch these to logger.debug (exo -v to enable) and drop the [PREFILL DEBUG] prefix if they are valuable

Evanev7:
closing as stale but i've copied your fix for pipeline into #1748

PR #1645, opened 2026-03-02
fix(worker): emit error chunks when a runner dies mid-command

### LINKED ISSUE #1586: API streams hang forever when runner process crashes during warmup/loading
## Bug Description

Streaming API responses (`/v1/chat/completions` with `stream: true`) hang forever when a runner process crashes during **model loading** or **warmup** phases. The stream never receives an error or terminates — it blocks indefinitely.

This happens because:
1. Runner crashes outside `TextGeneration` don't produce an `ErrorChunk`
2. The API's `_apply_state()` has no handler for `RunnerFailed` events
3. `_token_chunk_stream()` has no timeout or health-check mechanism

## Steps to Reproduce

1. Start a multi-node cluster with the master + at least one runner
2. Create an inference instance (e.g. via `/v1/models`)
3. Trigger a runner crash during model loading or warmup (e.g. OOM, corrupt model files, or kill the runner process)
4. Send a streaming chat completion request to the master's API
5. **Observe**: the SSE stream stays open forever — no error event, no timeout, no termination

## Expected Behavior

The stream should return an error response within a reasonable time (e.g. a few seconds) and close the connection.

## Actual Behavior

The stream hangs indefinitely. The client must manually abort the connection. No error is ever sent.

## Root Cause Analysis

Traced against latest `main` at `1780e4ad`. Three gaps combine to cause the hang:

### Gap 1: Runner — No ErrorChunk for non-TextGeneration crashes

**File**: `src/exo/worker/runner/llm_inference/runner.py`

- The `TextGeneration` case (~L235) has a try/except that sends `ChunkGenerated(ErrorChunk(finish_reason="error"))` before re-raising (~L384-395) — this correctly signals the API.
- The `LoadModel` case (~L150) calls `load_mlx_items()` at ~L176 with **no try/except**.
- The `StartWarmup` case (~L194) calls `warmup_inference()` at ~L209 with **no try/except**.

When these cases crash, the exception propagates to `bootstrap.py` (~L49-58) which sends `RunnerStatusUpdated(RunnerFailed)` only — **no ErrorChunk is ever generated**.

### Gap 2: API — `_apply_state()` ignores RunnerFailed entirely

**File**: `src/exo/master/api.py`

- `_apply_state()` (~L1622) handles exactly two event types:
  - `ChunkGenerated` → routes chunks to stream queues (~L1634-1652)
  - `TracesMerged` → saves traces (~L1653-1654)
- There are **zero references** to `RunnerStatusUpdated`, `RunnerFailed`, or any failure propagation mechanism in the entire API file.

This means even if the master's state correctly records a runner as failed, the API layer never learns about it and never notifies waiting streams.

### Gap 3: No timeout on `_token_chunk_stream()`

**File**: `src/exo/master/api.py`

- `_token_chunk_stream()` (~L567): the `async for chunk in token_chunks` loop at ~L582 blocks indefinitely.
- The loop only terminates on: `finish_reason is not None`, task cancellation, or async channel close.
- There is no timeout, no heartbeat, and no periodic health check on the runner/instance status.

### Additional: Master dispatches to dead runners

**File**: `src/exo/master/main.py`

- `_command_processor()` (~L133-174) selects an instance by lowest task count with **no runner health check**, so new tasks can be dispatched to a failed runner.
- `_event_processor()` (~L385-410) applies `RunnerFailed` to state but generates **no error events** for pending tasks on that runner.

**File**: `src/exo/shared/apply.py`

- `apply_runner_status_updated()` (~L193-198) only updates `state.runners` — **no cleanup of pending tasks** associated with the failed runner.

## Suggested Fix

A multi-layered fix:

1. **`_apply_state()` in `api.py`**: Handle `RunnerStatusUpdated(RunnerFailed)` events — find all stream queues associated with tasks on the failed runner and push an `ErrorChunk(finish_reason="error")` to each, so streams terminate.

2. **Runner `runner.py`**: Wrap `LoadModel` and `StartWarmup` task handlers in try/except blocks that send `ChunkGenerated(ErrorChunk(finish_reason="error"))` before re-raising, matching the existing `TextGeneration` pattern.

3. **Timeout safety net in `_token_chunk_stream()`**: Add an `asyncio.wait_for` or periodic health check so streams can't hang forever even if the above signals are somehow missed.

## Environment

- **Commit**: `1780e4ad` on `main`
- **OS**: macOS (Apple Silicon)
- **Python**: 3.12+

pandego:
Closes #1586

## Summary
- track in-flight tasks in `RunnerSupervisor` (not only unacknowledged pending tasks)
- when `_check_runner()` detects a crashed runner, emit `ChunkGenerated(ErrorChunk)` for each in-flight command task (`TextGeneration`, `ImageGeneration`, `ImageEdits`)
- keep existing `RunnerStatusUpdated(RunnerFailed)` emission so planner/state still transition correctly
- add a unit test for supervisor crash path to ensure an error chunk is emitted before failed runner status

## Why
`#1586` reports streams that can hang forever when runners crash during warmup/loading. This keeps failure signaling at the runner-supervisor layer, matching maintainer guidance in the issue thread.

## Validation
- attempted: `uv run pytest src/exo/worker/tests/unittests/test_runner/test_runner_supervisor.py`
- blocked locally by environment disk exhaustion while uv tried to materialize heavy CUDA wheels (`No space left on device` during `nvidia-cudnn-cu13` extraction)

I kept the change scoped and added a targeted unit test for the failure path.

### DIFF
--- a/src/exo/worker/runner/runner_supervisor.py
+++ b/src/exo/worker/runner/runner_supervisor.py
@@ -262,7 +262,7 @@ class RunnerSupervisor:
                             chunk=ErrorChunk(
                                 model=self.shard_metadata.model_card.model_id,
                                 error_message=(
-                                    "Runner crashed before completing command "
+                                    "Runner shutdown before completing command "
                                     f"({cause})"
                                 ),
                             ),
--- a/src/exo/worker/tests/unittests/test_runner/test_runner_supervisor.py
+++ b/src/exo/worker/tests/unittests/test_runner/test_runner_supervisor.py
@@ -83,7 +83,7 @@ async def test_check_runner_emits_error_chunk_for_inflight_text_generation() -> 
     assert isinstance(got_chunk, ChunkGenerated)
     assert got_chunk.command_id == command_id
     assert isinstance(got_chunk.chunk, ErrorChunk)
-    assert "Runner crashed before completing command" in got_chunk.chunk.error_message
+    assert "Runner shutdown before completing command" in got_chunk.chunk.error_message
 
     assert isinstance(got_status, RunnerStatusUpdated)
     assert isinstance(got_status.runner_status, RunnerFailed)


Evanev7:
looks good, waiting on #1642 before this i reckon

pandego:
Quick transparency update on this push:

- I fixed the concrete `nix flake check` typecheck errors reported in CI for the new regression test (`test_runner_supervisor.py`) and rebased on latest `main`.
- I ran targeted local validation, but full local env parity is limited on this runner (native build toolchain/maturin path), so I could not reproduce the full CI stack end-to-end before push.
- This update is intended to unblock CI verification of the typed fix directly in the project’s canonical checks.

If anything else fails in CI, I’ll follow up with a minimal patch quickly.

PR #1644, opened 2026-03-02
Add support for Qwen3.5

0xDaizz:
## Motivation

Qwen3.5 MoE models (e.g., `Qwen3.5-397B-A17B-6bit`) are now supported by `mlx-lm` via `qwen3_5_moe` model type, but exo lacks tensor parallel sharding support for this architecture. This prevents running large Qwen3.5 models across multiple nodes.

Qwen3.5 uses a GatedDeltaNet hybrid attention mechanism similar to Qwen3-Next, but with a different projection layout — separate `in_proj_qkv`, `in_proj_z`, `in_proj_b`, `in_proj_a` instead of Qwen3-Next's combined `in_proj_qkvz` and `in_proj_ba`. This requires architecture-aware sharding logic.

### `auto_parallel.py`
- Add imports for `Qwen3_5TextModel`, `Qwen3_5MoeModel`, `Qwen3_5DecoderLayer`, and `Qwen3_5SparseMoeBlock` from `mlx_lm.models.qwen3_5` / `qwen3_5_moe`
- Add Qwen3.5 models to the `QwenShardingStrategy` dispatch table
- Handle Qwen3.5's separate GatedDeltaNet projections via `hasattr(linear_attn, "in_proj_qkvz")` conditional
- Use section-aware sharding for `in_proj_qkv` with `segments=[key_dim, key_dim + key_dim]` to correctly split q/k/v sections across devices
- Add `Qwen3_5SparseMoeBlock` to MoE and shared expert sharding

### `model_cards.py`
- Add `Qwen3_5MoeForConditionalGeneration` to the `supports_tensor` whitelist

### `utils_mlx.py`
- Add Qwen3.5 EOS token IDs (`248046`, `248044`) to `get_eos_token_ids_for_model()`, following the existing pattern for GLM, Kimi, and GPT-OSS models

## Why It Works

Qwen3.5's GatedDeltaNet has an `in_proj_qkv` linear layer with three concatenated sections: `[q(key_dim), k(key_dim), v(value_dim)]`. A naive contiguous split (`segments=1`) would slice across section boundaries, corrupting q/k/v values and producing garbled output.

By passing `segments=[key_dim, key_dim + key_dim]` to `shard_linear()`, each section is split independently before distributing across devices. This ensures every rank receives correctly aligned q, k, and v components.

[...]

PR #1647, opened 2026-03-02
fix: update fa_idx/ssm_idx for Qwen3.5 after pipeline layer slicing

### LINKED ISSUE #1646: [BUG] mlx-community/Qwen3.5-122B-A10B-8bit Not Working Suddenly on >=2 Nodes
## Describe the bug

When loading mlx-community/Qwen3.5-122B-A10B-8bit across 2 nodes, one of them will throw this error:

`TypeError: ArraysCache.make_mask() got an unexpected keyword argument 'return_array'`

## To Reproduce

1. Download mlx-community/Qwen3.5-122B-A10B-8bit 
2. Select 2 nodes (Pipeline) + RDMA or TCP/IP
3. Launch

## Expected behavior

Model should load.

## Actual behavior

One node throws the error.

## Environment

- macOS Version: 26.3
- EXO Version: I've tried 1.0.68 all the way to the latest commit
- Hardware:
  - Device 1: MacBook Studio M4 Max, 128GB RAM
  - Device 2: MacBook Studio M4 Max, 128GB RAM
  - Additional devices:
- Interconnection:
  - Thunderbolt 5 cable, RDMA
  - 10GbE Ethernet between all devices

## Additional context
It was working fine previously, I can't figure out what is causing this. Nothing changed in my environment other than updating Xcode to 26.3 but I highly doubt this is the cause.

My initial findings are suggesting that mlx-lm 0.30.7 is causing the issue but this is really weird as it was working fine previously on this same version too.

Wysie:
## Motivation

When running a Qwen3.5 model (e.g. `mlx-community/Qwen3.5-122B-A10B-8bit`) across 2 nodes in pipeline mode, inference crashes with one of two errors depending on which node receives which layer slice:

- `TypeError: ArraysCache.make_mask() got an unexpected keyword argument 'return_array'`
- `AttributeError: 'ArraysCache' object has no attribute 'keys'`

Fixes #1646

## Root Cause

Qwen3.5 is a hybrid SSM/attention model. Its layers alternate between linear attention layers (GatedDeltaNet, using `ArraysCache`) and full-attention layers (using `KVCache`) every `full_attention_interval` (default: 4) layers.

`Qwen3_5TextModel.__init__` sets `fa_idx = full_attention_interval - 1` and `ssm_idx = 0` as **global** layer indices. When Exo splits the model across nodes, each node only receives a slice of the layers and a corresponding local cache list. However, `fa_idx` and `ssm_idx` were never updated to reflect the local slice.

On a node whose slice starts mid-model (e.g. layers 48–95), `cache[fa_idx]` (e.g. `cache[3]`) lands on an `ArraysCache` entry instead of the expected `KVCache`. This causes:

1. `create_attention_mask` to call `ArraysCache.make_mask` with kwargs it does not accept
2. The pipeline sync code to access `.keys` which `ArraysCache` does not have

## Changes

1. **`pipeline_auto_parallel`**: After slicing layers for each node, update `fa_idx` and `ssm_idx` on `Qwen3_5TextModel` to point to the first full-attention and first linear (SSM) layer in the local slice respectively. This follows the same pattern already used for `GptOssMoeModel` (`swa_idx`/`ga_idx`) and `Step35InnerModel` (`_swa_idx`/`_full_idx`).

2. **`PipelineLastLayer.__call__`** and **`patch_pipeline_model`**: Guard `.keys` access with `hasattr(_cache, "keys")` to be safe against cache types that do not have this attribute (such as `ArraysCache`). This is a defensive fix that prevents similar crashes for other future hybrid SSM/attention models.

## Manual Testing

Verified with `mlx-community/Qwen3.5-122B-A10B-8bit` split across 2x Mac Studio M4 Max 128GB nodes in pipeline mode. Both nodes complete warmup and inference successfully after this fix.

## Additional
The Qwen3.5-122B actually worked for me on v1.0.68 when I first tried but it no longer does. I'm not really sure why it worked initially, but this is what I think could've happened (I'm not really sure, need someone to verify):

The first time I ran it on v1.0.68, my two nodes happened to have a memory ratio that produced a 24/24 split (or any split divisible by 4). With a 24/24 split:

- Node 0: layers 0–23, cache[3] = KVCache (layer 3 is full attention)
- Node 1: layers 24–47, cache[3] = KVCache (layer 27 is full attention)

Both nodes worked perfectly by coincidence.

But when I ran it again later — even on the same v1.0.68 code — the available RAM on my nodes was slightly different (maybe a browser was open, or macOS allocated memory differently, or Spotlight was indexing). This caused a split like 23/25 instead of 24/24:

- Node 0: layers 0–22 → cache[3] = KVCache (layer 3) — still works
- Node 1: layers 23–47 → cache[3] = ArraysCache (layer 26) — crashes

This is exactly what my traceback shows: cache[0] = KVCache (the slice starts at layer 23, which is full attention), but cache[3] = ArraysCache (layer 26, which is linear).

### DIFF
--- a/src/exo/worker/engines/mlx/auto_parallel.py
+++ b/src/exo/worker/engines/mlx/auto_parallel.py
@@ -38,6 +38,7 @@ from mlx_lm.models.qwen3_next import Qwen3NextDecoderLayer, Qwen3NextSparseMoeBl
 from mlx_lm.models.step3p5 import Model as Step35Model
 from mlx_lm.models.step3p5 import Step3p5MLP as Step35MLP
 from mlx_lm.models.step3p5 import Step3p5Model as Step35InnerModel
+from mlx_lm.models.qwen3_5 import Qwen3_5TextModel
 
 from exo.shared.logging import logger
 from exo.shared.types.worker.shards import PipelineShardMetadata
@@ -190,10 +191,13 @@ class PipelineLastLayer(CustomMlxLayer):
             if cache is not None:
                 # CacheList (used by MLA models like DeepSeekV32, GLM MoE DSA)
                 # doesn't have .keys directly; access via first sub-cache.
+                # ArraysCache (used by hybrid SSM/attention models like Qwen3.5)
+                # has no .keys attribute at all — guard before accessing.
                 _cache = cache[0] if hasattr(cache, "caches") else cache  # type: ignore
-                _cache.keys = mx.depends(_cache.keys, output)  # type: ignore
+                if hasattr(_cache, "keys"):
+                    _cache.keys = mx.depends(_cache.keys, output)  # type: ignore
             mx.eval(output)
-            if cache is not None:
+            if cache is not None and hasattr(_cache, "keys"):  # type: ignore
                 mx.eval(_cache.keys)  # type: ignore
 
         if not self.is_prefill:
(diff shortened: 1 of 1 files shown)


rltakashige:
Hi there, thank you for opening this PR! I've taken the main parts from here and implemented it, including a fix for the behaviour you mentioned:

> Node 0: layers 0–22 → cache[3] = KVCache (layer 3) — still works
> Node 1: layers 23–47 → cache[3] = ArraysCache (layer 26) — crashes

So will be closing this in favour of #1644 . Thanks again!

wysie:
Thanks!

PR #1649, opened 2026-03-02
Batch cleanup

Evanev7:
what da ya think!


src/exo/worker/runner/llm_inference/batch_generator.py
```diff
@@ -68,92 +106,120 @@ class SequentialGenerator(InferenceGenerator):
...
-    )
+    _maybe_queue: list[TextGeneration] = field(default_factory=list, init=False)
+    _queue: deque[TextGeneration] = field(default_factory=deque, init=False)
     _active: (
         tuple[
             TextGeneration,
-            MpSender[GenerationResponse],
             Generator[GenerationResponse],
+            SendableGenerator[GenerationResponse],
+            Generator[GenerationResponse | ToolCallResponse | None],
```
Evanev7, line 127: commenting ahead of time; the first generator is for the raw GenerationResponses from mlx_generate - for compat with batching - the second is to send them into the processing pipeline, the third is for processed responses.

PR #1651, opened 2026-03-03
fix: coerce tool-call argument types from tool schema

sblOWPCKCR:
Apply schema-aware coercion to parsed tool-call arguments so Hermes-style toolcalls can still return typed JSON (e.g. integer ids).

 - pass request tools into parse_tool_calls
 - coerce parsed argument values by function parameters schema
 - add unit tests for coercion and unknown-tool passthrough

## Motivation

Models that use Hermes-based toolcall syntax (Qwen3.5) can't reliably call tools with non-string parameters
Example tool:
```json
    {
      "type": "function",
      "function": {
        "name": "process",
        "description": "Manage background processes",
        "parameters": {
          "type": "object",
          "properties": {
            "action": {
              "type": "string",
              "enum": ["spawn", "output", "kill", "list"]
            },
            "id": {
              "type": "integer",
              "description": "Process id"
            },
            "command": {
              "type": "string",
              "description": "Command to run for spawn"
            }
          },
          "required": ["action"],
          "additionalProperties": false
        }
      }
    }
```
Model transcript:
```
<tool_call>
<function=process>
<parameter=action>
output
</parameter>
<parameter=id>
0
</parameter>
</function>
</tool_call>
```
And the API returns:
`{"id":"a8f11689-d840-4ca5-ab1d-ead3678a11a9","name":"process","arguments":"{\"action\": \"output\", \"id\": \"0\"}"}}`

Tool definition declared `id` as `integer`, the model output is type-agnostic, and the translation layer treats everything as a string. 

The same Qwen3.5-27B on OpenRouter and GPT-4.1-mini on openai obey the function signature and emit correct call:
```
{"name":"process","arguments":"{\"action\": \"output\", \"id\": 0}"}}
```

Steps to reproduce:
```
❯ curl -sS -v http://localhost:52415/v1/chat/completions \
  -H 'Content-Type: application/json' \  -d @- <<'JSON'
{
  "model": "mlx-community/Qwen3.5-27B-4bit",
  "stream": false,
  "temperature": 0,
  "messages": [
    {
      "role": "user",
      "content": "Call the process tool with action=output and id=0. Do not explain anything. Just make the tool call."
    }
  ],
  "tools": [
    {
      "type": "function",
      "function": {
        "name": "process",
        "description": "Manage background processes",
        "parameters": {
          "type": "object",
          "properties": {
            "action": {
              "type": "string",
              "enum": ["spawn", "output", "kill", "list"]
            },
            "id": {
              "type": "integer",
              "description": "Process id"
            },
            "command": {
              "type": "string",
              "description": "Command to run for spawn"
            }
          },
          "required": ["action"],
          "additionalProperties": false
        }
      }
    }
  ],
  "tool_choice": {
    "type": "function",
    "function": {
      "name": "process"
    }
  }
}
JSON
```
Look for type of `id` function call

## Changes

Function call parameters are now converted to the types that the function declaration has

## Why It Works

We now explicitly convert types where we know it (and skip if we don't)

### Manual Testing
Create Qwen3.5-<ANY> instance, send the curl command above. Check that 'id' is now serialized as a number

### Automated Testing
Unit tests to cover basic type conversions

### DIFF
--- a/src/exo/worker/runner/llm_inference/model_output_parsers.py
+++ b/src/exo/worker/runner/llm_inference/model_output_parsers.py
@@ -22,7 +22,7 @@ from exo.worker.engines.mlx.utils_mlx import (
 )
 from exo.worker.runner.bootstrap import logger
 
-from .tool_parsers import ToolParser, execute_tool_parser
+from .tool_parsers import ToolParser
 
 
 @cache
@@ -38,7 +38,7 @@ def apply_all_parsers(
     tokenizer: TokenizerWrapper,
     model_type: type[Model],
     model_id: ModelId,
-    tools: list[dict[str, Any]] | None = None,
+    tools: list[dict[str, Any]] | None,
 ) -> Generator[GenerationResponse | ToolCallResponse | None]:
     mlx_generator = receiver
 
@@ -329,7 +329,7 @@ def parse_thinking_models(
 def parse_tool_calls(
     responses: Generator[GenerationResponse | None],
     tool_parser: ToolParser,
-    tools: list[dict[str, Any]] | None = None,
+    tools: list[dict[str, Any]] | None,
 ) -> Generator[GenerationResponse | ToolCallResponse | None]:
     in_tool_call = False
     tool_call_text_parts: list[str] = []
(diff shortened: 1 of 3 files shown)


Evanev7:
i believe mlx's parsers have some upstream support for this, though i actually prefer this final-pass approach. can you move the code into the execution of the tool_parser, rather than in runner.py?

Evanev7:
also as always, thanks for the pr!

sblOWPCKCR:
@Evanev7 how about now?

Evanev7:
nice. might be worth integrating vllms tool parsers rather than mlxs in future, but this is an easy win for now

PR #1656, opened 2026-03-04
remove custom discovery

Evanev7:
simplesimplesimple


ciaranbor:
Any reason we can't use commit hash for version mismatch?

PR #1658, opened 2026-03-04
Ciaran/re download bug

ciaranbor:
## Motivation

After deleting a model and re-downloading it, the CachedShardDownloader returns the stale cached path, so ensure_shard short-circuits and no download actually happens.

## Changes

- Added invalidate(model_id) method to the ShardDownloader ABC and all implementations
- CachedShardDownloader.invalidate evicts cache entries matching the model ID and delegates down
- DownloadCoordinator.delete_model calls invalidate after cancelling active downloads, before deleting files
- Added end-to-end test that downloads, deletes, and re-downloads a model through the coordinator

## Why It Works

The cache is cleared when a model is deleted, so the next ensure_shard call performs a fresh download instead of returning the stale path.

## Automated Testing

New test_re_download_after_delete_completes exercises the full download → delete → re-download flow through DownloadCoordinator with CachedShardDownloader + SingletonShardDownloader wrappers matching production.

### DIFF
--- a/src/exo/download/impl_shard_downloader.py
+++ b/src/exo/download/impl_shard_downloader.py
@@ -19,9 +19,7 @@ def exo_shard_downloader(
     max_parallel_downloads: int = 8, offline: bool = False
 ) -> ShardDownloader:
     return SingletonShardDownloader(
-        CachedShardDownloader(
-            ResumableShardDownloader(max_parallel_downloads, offline=offline)
-        )
+        ResumableShardDownloader(max_parallel_downloads, offline=offline)
     )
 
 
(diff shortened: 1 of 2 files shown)


AlexCheema (dismissed):
This is good, and I would merge it.
Before we do, only thing I'm thinking is if we should also now cover the edge case that a user deletes it from disk?

It might be worth instead removing `CachedShardDownloader` completely - the question would be how often do we call `ensure_shard`?

ciaranbor:
I'm almost sure it only gets called from a user-initiated download, will remove it

ciaranbor:
Also, the download logic already uses a local cache to avoid hitting HF unnecessarily. There's no real benefit to it

PR #1659, opened 2026-03-04
Fix copy code button not working in dashboard

### LINKED ISSUE #1657: [BUG] Copy Button in Code Blocks on Dashboard (Web UI) Doesn't Work
## Describe the bug
When using the web UI and the response contains a code block, there's a copy button that allows us to copy the code. However, clicking on it doesn't actually copy the code.

## To Reproduce

1. Load any model
2. Prompt model to build a simple app in HTML, e.g. a todo app in HTML
3. The response will contain code block, click the copy button

## Expected behavior

The code should be copied.

## Actual behavior

Nothing happens.

## Additional context

Tested on different devices (Mac, Windows, etc.)  in Chrome, opening the dashboard.

Wysie:
Fixes #1657

## Motivation

The copy button on code blocks in the dashboard does nothing when clicked. This affects all code blocks in assistant responses.

## Root Cause

In `MarkdownContent.svelte`, click event listeners are bound to `.copy-code-btn` elements via `setupCopyButtons()` inside a Svelte 5 `$effect`. However, the effect fires before the DOM has been updated with the new HTML, so `querySelectorAll(".copy-code-btn")` finds zero buttons.

Additionally, during streaming, the `content` prop updates on every token, causing the entire `{@html processedHtml}` to be re-rendered. This destroys all previously bound event listeners, even if they were successfully attached.

## Changes

Replaced the per-button `addEventListener` approach with **event delegation** — a single click listener on the container element that catches clicks bubbling up from any `.copy-code-btn` or `.copy-math-btn`. This:

- Eliminates the timing issue (the listener exists before the buttons are rendered)
- Survives HTML re-renders during streaming (no need to re-bind)
- Removes the need for `setupCopyButtons()` and the `data-listenerBound` tracking

## Testing

1. Load any model
2. Prompt it to generate a code block (e.g. "write a hello world in Python")
3. Click the copy button on the code block
4. Paste — the code is copied correctly
5. Verified the button also works during streaming (before generation completes)

### DIFF
--- a/dashboard/src/lib/components/MarkdownContent.svelte
+++ b/dashboard/src/lib/components/MarkdownContent.svelte
@@ -510,21 +510,25 @@
     if (!containerRef || !browser) return;
 
     function handleDelegatedClick(event: MouseEvent) {
-      const codeBtn = (event.target as HTMLElement).closest('.copy-code-btn') as HTMLButtonElement | null;
+      const codeBtn = (event.target as HTMLElement).closest(
+        ".copy-code-btn",
+      ) as HTMLButtonElement | null;
       if (codeBtn) {
         handleCopyClick({ currentTarget: codeBtn } as unknown as Event);
         return;
       }
-      const mathBtn = (event.target as HTMLElement).closest('.copy-math-btn') as HTMLButtonElement | null;
+      const mathBtn = (event.target as HTMLElement).closest(
+        ".copy-math-btn",
+      ) as HTMLButtonElement | null;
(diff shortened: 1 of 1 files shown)


AlexCheema (approved):
looks good, tested and works.

wysie:
How can I fix the checks failing?

Evanev7:
just run the formatter, `nix fmt`

wysie:
Thanks @Evanev7

Issue #934, opened 2025-12-20
Vulkan Support


TOTO-sys28:
There is people have a old gpu's and they want to use them 
this is why we need a vulkan support to use them

Evanev7:
Agreed! see #904 for some of our other discussion so far

mahald:
Robust Vulkan support—particularly on Linux—is essential for cross-vendor compatibility across AMD, NVIDIA, and Intel GPUs. I am waiting for this implementation to fully leverage Ryzen AI+ 395 systems equipped with Thunderbolt 5.

Atomic-Germ:
> Robust Vulkan support—particularly on Linux—is essential for cross-vendor compatibility across AMD, NVIDIA, and Intel GPUs. I am waiting for this implementation to fully leverage Ryzen AI+ 395 systems equipped with Thunderbolt 5.

That chip is just going to cook itself if you try it.

FlorianHeigl:
vulkan also saves some nervous breakdowns with mi50/radeon pro vii/radeon vii, especially as a fall-back or fall-through choice. when i tried last year (was it?) with hip and clang both were non-fun options. vulkan is much more a "safe" choice as in it usually works when you try to use it.
I do have my mi50 patched to use the gpu output, it has a working rocm install and i have a vllm docker container that's optimized for it. what i can never have is the certainty that all that pile of stones is still standing the next day or if i tried to update something to the next version. even if rocm were the preferable/faster (sometimes is, sometimes ain't) it would be ok to experiment if there was a safe second path...
also, PSA, i never again wanna try solving such problems with zluda.

Ryzen AI  395 sounds like a nice problem to have :>

Issue #913, opened 2025-12-18
[TRACKING] General Linux CUDA Support


Evanev7:

pcgeek86:
It's probably a separate feature request, but it would be awesome to build a heterogeneous cluster with both Linux and Windows systems, all with NVIDIA GPUs.

For example, I have a Windows 11 system with an NVIDIA GeForce RTX 4070 Ti SUPER 16 GB, and a Linux server with an RTX 3060 12 GB. If I could cluster these together, along with additional nodes, that would be sweet!

AndreiCravtov:
Man, Windows is like a whole different beast. Atleast macOS is in the Unix family. But it would be cool for sure

Evanev7:
We expect we'll need a full on compat matrix at some point for features like RDMA - I don't expect we'll ever have windows RDMA for example - but never say never. Pipeline on the other hand is 100% doable and 'proper' windows support should generalise to heterogeneous pipeline.

paulie-g:
I would add Linux RDMA and RoCE support as a todo. Both are incredibly affordable right now with older adapters/NICs (eg Mellanox ConnectX-{3,4,5}) and switches. People who know what they're doing can easily match RDMA-over-Thunderbolt performance (sufficient throughput at ~same latency) for three figures (with access to the US secondary market). Such support would open the gates for affordable clusters with heavily discounted secondary market gear (older server platforms with enough pcie to run multiple older GPUs and an infiniband/RoCE capable card). While they won't win any tok/s performance prizes, they would enable loading and running large models in the four figures price range.

Evanev7:
Out of curiosity, what do you consider a large model? I wasn't aware you could really get to the TB range for <10k

paulie-g:
> Out of curiosity, what do you consider a large model? I wasn't aware you could really get to the TB range for <10k

In *theory* either via *high N* cards that are sub-$5/GB (eg V340L) or CPU/DDR3. But you're right, in current-year 1TB+ is probably what one would call 'large' and I haven't actually attempted this. My point was rather that RDMA/RoCE can be had on old, inexpensive platforms that would make this sort of experiment possible.

Evanev7:
absolutely, I consider this issue solved with pipeline CUDA but RDMA should follow (it will probably be smoother to implement than macos RDMA as well haha)

paulie-g:
[...]

Issue #904, opened 2025-12-18
Framework Desktop Support


Evanev7:
EXO Does not currently support linux unified memory machines like the Framework Desktop out of the box. We should add this ASAP

AndreiCravtov:
True

Gulianrdgd:
Hey, I created this pull request for Ollama to allow distributed inference on the framework desktop. https://github.com/ollama/ollama/pull/13087.

However I would like to run EXO, since then we could ideally connect more devices than just our framework desktops.

In my pull request I use the ROCm driver instead of CUDA or Vulkan for the framework desktop. I see that in your planned Tier 2 platform support there is not mention of ROCm? Why is that exactly? (Here is some info on ROCm for AMD Strix Halo https://rocm.docs.amd.com/projects/radeon-ryzen/en/latest/docs/install/installryz/native_linux/install-ryzen.html)

Maybe I can help with some of the challenges?

AndreiCravtov:
I think given that AMD's equivalent to CUDA is infact ROCm and the unified memory chip on framework desktop is from ADM, it should be ROCm on that list.

But let's see what others say.

Evanev7:
We generally haven't investigated ROCm vs Vulkan support enough to commit to what kind of support we're going to provide - Vulkan seemed the more general option so it's what we added yesterday - I'll give those docs a read and we can re-evaluate.

geerlingguy:
Vulkan would work across Intel, AMD, and Nvidia (if for some reason you don't want to use CUDA), and currently has great support across even arm64 and riscv64 architectures too, which ROCm supports... almost not at all :(

nbritton:
Chasing ROCm is a boondoggle because AMD drops support for their hardware before anyone even has time to develop a software stack around their ecosystem. Compounding on that, [Vulkan is already faster than ROCm anyways](https://github.com/ggml-org/llama.cpp/discussions/10879). So focus your efforts on Vulkan and RDMA ibverbs with ConnectX-4/5/6/7/8 cards (mlx5) with PeerDirect / GPUDirect.

[...]

PR #1660, opened 2026-03-04
feat(Linux): Tinygrad runner for LLM inference on Linux

triko88:
## Motivation

The initial version suffered from dysfunctional Linux code. The re-write uses MLX CUDA for Linux, however the performance as been reported to be sub-optimal. This limits the users into two categories:
- Apple users who use MLX
- Linux users with Nvidia graphic cards, who can use MLX CUDA

While the older version ([ex-exo](https://github.com/exo-explore/ex-exo)) used tinygrad, the overall codebase wasn't optimized to handle the architectural differences between Apple Silicon and the contemporary PC architecture. This led to a broken experience on Linux, which me and other Linux users faced. As evident by issues #904, #910, #913 and #934. 

One user concluded (for the archived version) after using it on Nvidia RTX 3060 12GB VRAM:
> Either I need some other hardware, OS or libraries or this Exo thing does not work at all… 
> Will give a try later.
>
> *Michał Sobczak* (https://michalasobczak.pl/ai-ml/2025/03/exo-the-gpu-cluster-tinygrad-mlx/)

The motivation here was clear. **Build a usable, if not a performant tinygrad runner for Linux that can run heterogeneously with Apple systems in the future.**

## Changes

This change introduces a tinygrad based runner in this project that can load MLX safetensor weights and infer them. Because of its fundamental nature, this is a huge change. Done in 10 phases to build a foundational and correct inference engine that can do the following, while relying purely on tensor ops:
- Deseralize Huggingface MLX safetensors file. The idea is that the Apple and non-Apple systems must share the weights, and infer accordingly. 
- Implement an architecture spec registry to unify weight names and functions for their transformer architecture. 
- Build quantized linear and embedding layers that are interoperable with tinygrad counterparts.
- Run MLX dequantization strategy on tinygrad during inference.
- Reduce kernel dispatch and reuse kernels, to minimize latency and maximize throughput.

## Why It Works

The original exo treated tinygrad as a drop-in runtime equivalent to MLX. It isn't, MLX is a runtime. You call an op, it executes. Tinygrad is a compiler, it builds a computation graph, generates GPU source code, compiles it into a kernel, and only then dispatches. First compilation costs 50–1600ms per kernel shape; cached re-invocations cost ~3ms.

The old architecture ran tinygrad on a background thread via `run_in_executor` inside the main process. This meant:

1. **Kernel compilation during inference.** No warmup phase, so every unique shape compiled while the user waited. Producing 26s time-to-first-token.
2. **Auto-tuner disabled.** Tinygrad's BEAM search uses Python signals (main thread only). Running in a thread pool crashed with `BEAM=1`, so users defaulted to `BEAM=0`, no optimization.
3. **State leaking between requests.** Lazy computation graphs accumulated across calls with no clean boundary, causing context contamination and infinite generation loops.

This change exploits the new architecture's process-isolated Runner model. The tinygrad Runner is a separate child process: main thread available, environment variables (`DEV`, `JIT`, `TC`, `BEAM`) inherited naturally, memory space isolated. The Worker's plan tree sequences the lifecycle correctly: `DownloadModel → LoadModel → StartWarmup → Ready`. `StartWarmup` pre-compiles every kernel before accepting requests.

**Result:** TTFT 26s → 745ms, throughput 9 → 64.5 tok/s, clean generation termination, and context isolation by process boundary. Later commits extend this pattern for other models and GPU backends.

### Manual Testing

**Hardware:** HP Omen 16-n0079AX
**Specs:** 
- CPU: AMD Ryzen 7 6800H
- GPU: AMD Radeon RX 6650M
- OS: Omarchy Linux with Cachy OS kernel (Linux v6.19)

**E2E test steps**
- Run exo using `uv run exo`. 
- To test cache coherence, run `DEBUG=1 uv run exo`.
- Generally, tinygrad selects the best inference backend for the machine. However, in case of MMU failure like with RDNA 2, you can run exo with another inference backend through the environment variable.
- To test with another inference backend, `DEV=<backend> uv run exo` to run the following backends:
  - `DEV=HIP` to run Heterogeneous-computing Interface for Portability. Use it for RDNA 2 GPUs or older.
  - `DEV=CL` to run OpenCL.
- Select tensor in the dashboard to run tensor cores in your GPU.

### Automated Testing

Unit tests are written in `src/exo/worker/tests/unittests/test_tinygrad`. Run the tests using `pytest`:
```sh
uv run basedpyright && uv run ruff check && uv run pytest
```

Specifically for the changed files, the most relevant test files are:

- `uv run pytest src/exo/worker/tests/unittests/test_tinygrad/test_cache.py`. Exercises KVCache (renamed keys/values fields)
- `uv run pytest src/exo/worker/tests/unittests/test_tinygrad/test_sampling.py`. Exercises `sample_token` (strict=True zip fix)
- `uv run pytest src/exo/worker/tests/unittests/test_tinygrad/test_generate.py`. Exercises the generator (TinyJit import, cache field refs)
- `uv run pytest src/exo/worker/tests/unittests/test_tinygrad/test_layers.py`. Exercises attention layer
- `uv run pytest src/exo/worker/tests/unittests/test_tinygrad/test_weight_loader.py`. Exercises weight loading
- `uv run pytest src/exo/shared/tests/test_tokenizer_shared.py`. The fixed test_unknown_model_returns_none test

## Expected Results
- `basedpyright`: 8 pre-existing errors (all in connection_message.py, test_master.py, test_node_id_persistence.py. Not from this branch)
- `ruff`: 0 errors
- `pytest`: All pass. 5 MLX collection errors on Linux are pre-existing (macOS-only mlx.core dependency). 1 test_master failure is pre-existing (Keypair.to_node_id attribute error).

screenshot-2026-03-03_21-44-47

https://github.com/user-attachments/assets/c4516e81-283f-4f5d-aec1-98636f7f666a

### DIFF
--- a/src/exo/routing/router.py
+++ b/src/exo/routing/router.py
@@ -241,6 +241,6 @@ def get_node_id_keypair(
 
         # if no valid credentials, create new ones and persist
         with open(path, "w+b") as f:
-            keypair = Keypair.generate_ed25519()
+            keypair = Keypair.generate()
             f.write(keypair.to_bytes())
             return keypair
--- a/src/exo/worker/runner/bootstrap.py
+++ b/src/exo/worker/runner/bootstrap.py
@@ -76,15 +76,7 @@ def entrypoint(
         os.environ.setdefault("JIT", "1")
         os.environ.setdefault("BEAM", "2")
 
-        # Set TC based on sharding strategy:
-        # Tensor parallelism → TC=1, Pipeline → TC=0
-        from exo.shared.types.worker.shards import TensorShardMetadata
-
-        if not os.environ.get("TC"):
-            if isinstance(bound_instance.bound_shard, TensorShardMetadata):
-                os.environ["TC"] = "1"
-            else:
-                os.environ["TC"] = "0"
+        os.environ.setdefault("TC", "1")
 
     try:
         if is_tinygrad:


Evanev7:
huge! i appreciate our batching changes will have made some work for you to merge this in, but before i dig into it can you move exo/architecture/master back to exo/master?

triko88:
Sure, I'll move directory `exo/architecture/master` back to `exo/master`.

triko88:
I've pushed the change, please review it and let me know whether the directory has been correctly merged or not.

src/exo/routing/router.py
```diff
@@ -238,7 +221,7 @@ def get_node_id_keypair(
     Obtain the :class:`PeerId` by from it.
     """
     # TODO(evan): bring back node id persistence once we figure out how to deal with duplicates
-    return Keypair.generate()
+    return Keypair.generate_ed25519()
```
Evanev7, line 224: minor merge issue

src/exo/main.py
```diff
@@ -40,12 +38,12 @@ class Node:
...
     node_id: NodeId
     offline: bool
-    _tg: TaskGroup = field(init=False, default_factory=TaskGroup)
+    _tg: TaskGroup = field(init=False, default_factory=anyio.create_task_group)
 
     @classmethod
     async def create(cls, args: "Args") -> Self:
         keypair = get_node_id_keypair()
-        node_id = NodeId(keypair.to_node_id())
+        node_id = NodeId(keypair.to_peer_id())
```
Evanev7, line 46: this as well

Evanev7 (commented):
just the merging for now. i might take a stab at merging this in with the recent refactor myself if that's ok? the new runner **should** be more backend agnostic, but im not sure if that's actually the case

src/exo/routing/router.py
triko88: This change allowed me to start exo on Linux during E2E test.

Edit: After reverting this change, it worked. The error comes from main function. This one is safe to revert.

src/exo/main.py
triko88: Couldn't load exo on Linux without this change for E2E tests

Edit: Reverting the change on Linux lead me to this:
```
exo feature/linux-support  ? ❯ DEV=HIP DEBUG=1 uv run exo
[ 11:36:30.8739PM | INFO    ] Starting EXO
[ 11:36:30.8743PM | INFO    ] EXO_LIBP2P_NAMESPACE: None
  File "/home/apan/git-projects/exo/.venv/bin/exo", line 10, in <module>
    sys.exit(main())
             ~~~~^^
  File "/home/apan/git-projects/exo/src/exo/main.py", line 272, in main
    node = anyio.run(Node.create, args)
  File "/home/apan/git-projects/exo/.venv/lib/python3.13/site-packages/anyio/_core/_eventloop.py", line 74, in run
    return async_backend.run(func, args, {}, backend_options)
           ~~~~~~~~~~~~~~~~~^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/apan/git-projects/exo/.venv/lib/python3.13/site-packages/anyio/_backends/_asyncio.py", line 2325, in run
    return runner.run(wrapper())
           ~~~~~~~~~~^^^^^^^^^^^
  File "/home/apan/.local/share/uv/python/cpython-3.13.12-linux-x86_64-gnu/lib/python3.13/asyncio/runners.py", line 118, in run
    return self._loop.run_until_complete(task)
           ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~^^^^^^
  File "/home/apan/.local/share/uv/python/cpython-3.13.12-linux-x86_64-gnu/lib/python3.13/asyncio/base_events.py", line 725, in run_until_complete
    return future.result()
           ~~~~~~~~~~~~~^^
  File "/home/apan/git-projects/exo/.venv/lib/python3.13/site-packages/anyio/_backends/_asyncio.py", line 2313, in wrapper
    return await func(*args)
           ^^^^^^^^^^^^^^^^^
  File "/home/apan/git-projects/exo/src/exo/main.py", line 46, in create
    node_id = NodeId(keypair.to_node_id())
                     ^^^^^^^^^^^^^^^^^^
AttributeError: 'builtins.Keypair' object has no attribute 'to_node_id'
```

triko88: I have no issues with merging my changes with your recent refractor. My runner is as backend agnostic as it can be, since the changes are tinygrad tensor operations.

Evanev7: yup - have you rebuilt the rust bindings inbetween? uv should catch it but often you need to `uv sync --upgrade-package exo_pyo3_bindings`

triko88: I haven't rebuilt the rust bindings. I rebuilt it to see it is causing any issues, running  `uv sync --upgrade-package exo_pyo3_bindings` post rebuild didn't cause any issue.

Evanev7: hm - i suppose your base branch is far enough back that these changes haven't landed yet

triko88: yep - there were cases where my initial iterations broke while catching up with the base branch. So I decided to slow down my update frequency with the branch.

Evanev7: reasonable!

compscidr:
Tested this on my rtx5080 on ubuntu, seems to be working!, here's some instructions in case anyone wants to try to reproduce: https://gist.github.com/compscidr/3e071ab6c2dce60339ca953eb0a98787

Working with Llama 3.2 1B 4bit
However, Llama 3.1 8B 8bit fails with shape broadcast error in rms_norm:
```
  ValueError: cannot broadcast (1, 64, 1024) to new_shape=(1, 64, 4096)
  Likely a hidden_size chunking bug in the tinygrad forward pass
```

Actually was able to get it working with the other model - there are more detailed notes about how to do so as well. There is one other issue that came up with fork / spawn with notes in there too.

PR #1665, opened 2026-03-05
feat(mlx): add repetition_penalty and repetition_context_size to chat completions

mlpy0:
## Problem

Models running through exo can get stuck in repetition loops — generating the same text over and over until hitting the token limit. It happens more with quantized models where probability distributions can become degenerate.

`mlx_lm` has `make_logits_processors(repetition_penalty, repetition_context_size)` that handles this, but it is never called anywhere in the pipeline.

## Solution

Wire `repetition_penalty` and `repetition_context_size` through the stack:

- `ChatCompletionRequest` → accept the params from the client
- `TextGenerationTaskParams` → carry them through the pipeline
- `chat_completions.py` adapter → map to internal params
- `generate.py` → call `make_logits_processors()` and merge into the `logits_processors` list

When `repetition_penalty` is `None` (default), `make_logits_processors` returns `[]` so existing behavior is unchanged.

## Usage

```json
{
  "model": "mlx-community/...",
  "messages": [...],
  "repetition_penalty": 1.15,
  "repetition_context_size": 64
}
```

### DIFF
--- a/src/exo/worker/engines/mlx/generator/generate.py
+++ b/src/exo/worker/engines/mlx/generator/generate.py
@@ -10,7 +10,7 @@ from mlx_lm.generate import (
     stream_generate,
 )
 from mlx_lm.models.cache import ArraysCache, RotatingKVCache
-from mlx_lm.sample_utils import make_sampler
+from mlx_lm.sample_utils import make_logits_processors, make_sampler
 from mlx_lm.tokenizer_utils import TokenizerWrapper
 
 from exo.shared.types.api import (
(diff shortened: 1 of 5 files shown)


src/exo/worker/engines/mlx/generator/generate.py
```diff
@@ -476,6 +476,11 @@ def mlx_generate(
         eos_ids = eos_ids_from_tokenizer(tokenizer)
         logits_processors = [ban_token_ids(eos_ids)]
 
+    logits_processors = logits_processors + make_logits_processors(
+        repetition_penalty=task.repetition_penalty,
+        repetition_context_size=task.repetition_context_size,
+    )
+
```
Evanev7: nit: just set the default value of logits_processors to make_logist_processors()

mlpy0: done

PR #1670, opened 2026-03-06
Add backpressure guard for chat text generations

### LINKED ISSUE #1664: [BUG] Cline sub-agents and parallel tool calling crashes cluster
## Describe the bug

Using Cline with sub-agents enabled will crash the entire cluster or sometimes the machine with more resources.

## To Reproduce

1. Setup TCP pipeline sharding strategy cluster with M3 max 64 gb and M4 max 128gb
2. Load Qwen3-coder-next 8bit on the cluster
3. Start a plan with Sub-agents and parallel tool calling feature enabled in cline

## Expected behavior

Prompt evaluation from all sub-agents succeeds

## Actual behavior

Main machine crashes and disappears from the cluster.

## Environment

- macOS Version: 26.3
- EXO Version: 1.0.68
- Hardware:
  - Device 1: M4 max 128gb, 1tb storage
  - Device 2: M3 max 64gb, 1tb storage
  - Additional devices:
- Interconnection:
  - TCP connection Wifi 6 between both devices.

pandego:
## Summary
Add backpressure protection for chat text generations in the master API.

## Problem
When too many text generations are already in flight, new requests can overload the master process and degrade responsiveness.

## Root cause
There was no explicit in-flight guard for text generation requests at the API entry point.

## Fix
- Add a configurable backpressure guard for in-flight text generation requests.
- Return HTTP 429 when the in-flight count is at or above the configured limit.
- Add focused tests for both allow and reject behavior.

## Validation
- uv run pytest src/exo/master/tests/test_text_generation_backpressure.py src/exo/master/tests/test_cancel_command.py

Closes #1664.

### DIFF
--- a/src/exo/api/main.py
+++ b/src/exo/api/main.py
@@ -2,6 +2,7 @@ import base64
 import contextlib
 import hashlib
 import json
+import os
 import random
 import time
 from collections.abc import AsyncGenerator, Awaitable, Callable, Iterable
@@ -864,6 +865,25 @@ class API:
             "TODO: we should send a notification to the user to download the model"
         )
 
+    def _enforce_text_generation_backpressure(self) -> None:
+        """Protect workers from overload by limiting concurrent text streams.
+
+        Sub-agent workloads can issue many concurrent requests (for example, parallel
+        tool-calling clients). A bounded number of in-flight streams prevents runaway
+        queue growth and OOM cascades on smaller clusters.
+        """
+        max_in_flight = int(os.getenv("EXO_MAX_IN_FLIGHT_TEXT_GENERATIONS", "2"))
+        in_flight = len(self._text_generation_queues)
+        if in_flight >= max_in_flight:
+            raise HTTPException(
+                status_code=HTTPStatus.TOO_MANY_REQUESTS,
+                detail=(
+                    "Server is busy processing other generations. "
+                    f"in_flight={in_flight}, limit={max_in_flight}. "
+                    "Retry shortly or reduce client-side parallelism."
+                ),
+            )
+
     async def _send_text_generation_with_images(
         self, task_params: TextGenerationTaskParams
     ) -> TextGeneration:
@@ -917,6 +937,7 @@ class API:
         self, payload: ChatCompletionRequest
     ) -> ChatCompletionResponse | StreamingResponse:
         """OpenAI Chat Completions API - adapter."""
+        self._enforce_text_generation_backpressure()
         task_params = await chat_request_to_text_generation(payload)
         validated_model = await self._validate_model_has_instance(task_params.model)
         task_params = task_params.model_copy(update={"model": validated_model})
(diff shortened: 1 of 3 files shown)


Evanev7 (commented):
i like the direction - can we pass this through the construction of the API & a cli flag rather than an env var? i'm working towards implementing live configuration and it'd be good to start plumbing these things instead of using env.

pandego:
Great suggestion - done in `acef4c92`.

I replaced the env-based limit with explicit API/CLI plumbing:
- added CLI flag `--max-in-flight-text-generations` (default `2`)
- wired `Args.max_in_flight_text_generations` through `Node.create` into `API(...)`
- updated `API` to use `self.max_in_flight_text_generations` in `_enforce_text_generation_backpressure`
- updated backpressure tests to assert configured limit behavior without env patching

Validation run:
- `uv run pytest src/exo/master/tests/test_text_generation_backpressure.py -q` (2 passed)

pandego:
Refreshed this branch onto current `main` and force-pushed the updated PR branch.

Validation blocker from this environment:
- `uv run pytest src/exo/master/tests/test_text_generation_backpressure.py src/exo/master/tests/test_cancel_command.py -q`
- blocked while building the `mlx` dependency because the local toolchain cannot find `metal` (`xcrun: error: unable to find utility "metal"`)

The PR diff is still the intended focused scope against current `main`:
- `src/exo/main.py`
- `src/exo/master/api.py`
- `src/exo/master/tests/test_text_generation_backpressure.py`

pandego:
I rebased this branch onto the current `main` so the dirty merge state is cleared.

Validation note: I re-ran the focused local check path, but this environment still hits the same `mlx` Metal Toolchain blocker before the target path can complete, so I could not produce a fresh green local run here.

If you want, I can keep digging on the `mlx` blocker next, but the branch itself is refreshed and ready for another look.

pandego:
I rebased this branch onto the current `main` so the dirty merge state is cleared.

Validation note: I re-ran the focused local check path, but this environment still hits the same `mlx` Metal Toolchain blocker before the target path can complete, so I could not produce a fresh green local run here.

If you want, I can keep digging on the `mlx` blocker next, but the branch itself is refreshed and ready for another look.

pandego:
Rebased this onto current main and kept the backpressure guard aligned with the newer async/image-aware text generation flow.

Validation:
- `uvx ruff check src/exo/api/main.py src/exo/main.py src/exo/master/tests/test_text_generation_backpressure.py`
- `uv run --no-sync python -m compileall -q src/exo/api/main.py src/exo/main.py src/exo/master/tests/test_text_generation_backpressure.py`

I also tried the focused pytest path, but `uv run pytest src/exo/master/tests/test_text_generation_backpressure.py -q` stalled during dependency sync before pytest started in this worktree.

pandego:
Rebased this branch onto the current `main` and force-pushed the refreshed branch.

Validation:
- `uv run pytest src/exo/master/tests/test_text_generation_backpressure.py -q`

The PR should be back in a clean review state.

pandego:
Rebased this branch onto the current `main` and force-pushed the refreshed branch.

Validation:
- `uv run pytest src/exo/master/tests/test_text_generation_backpressure.py -q`

The PR should be back in a clean review state.

PR #1673, opened 2026-03-06
Ciaran/gpt oss tool call finish reason

ciaranbor:
## Motivation

When gpt-oss truncates a response mid-tool-call (e.g. hitting max tokens), the finish_reason was swallowed because the parser was in tool-accumulation mode and never yielded it back to the caller. This caused the API to hang or fail to signal completion.

## Changes

In parse_gpt_oss, when accumulating tool call arguments and a finish_reason arrives, yield the response (with empty text) so the finish reason propagates

## Why It Works

The parser now checks for finish_reason on every chunk while inside a tool call. When the model stops generating (truncation), the finish signal is forwarded immediately rather than being silently consumed.

### Automated Testing

Added tests covering truncated tool calls (finish_reason="length") and plain text truncation

### DIFF
--- a/src/exo/worker/runner/llm_inference/model_output_parsers.py
+++ b/src/exo/worker/runner/llm_inference/model_output_parsers.py
@@ -117,7 +117,7 @@ def parse_gpt_oss(
             if delta:
                 tool_arg_parts.append(delta)
             if response.finish_reason is not None:
-                yield response.model_copy(update={"text": ""})
+                yield response.model_copy(update={"text": "".join(tool_arg_parts)})
                 tool_arg_parts = []
             continue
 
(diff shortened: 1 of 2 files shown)


src/exo/worker/runner/llm_inference/model_output_parsers.py
```diff
@@ -116,6 +116,9 @@ def parse_gpt_oss(
         if current_tool_name is not None:
             if delta:
                 tool_arg_parts.append(delta)
+            if response.finish_reason is not None:
+                yield response.model_copy(update={"text": ""})
```
Evanev7: for consistency with the other tool parser, should we yield the partial tool call content? or vice versa, should the other parser yield nothing on a partial tool call?

ciaranbor: I figured it's best to not return invalid JSON. Not sure if that's actually a problem.

ciaranbor: So I would lean towards updating the other parser in line with this

ciaranbor: After discussing we've done a 180. Returning partial results

Issue #950, opened 2025-12-21
mDNS Peer Discovery Not Working on macOS with libp2p


lucasajackson:
# mDNS Peer Discovery Not Working on macOS with libp2p

## Environment
- **OS**: macOS Tahoe 26.2
- **EXO Version**: 0.3.0 (installed from source, latest main branch as of Dec 21, 2025)
- **Python**: 3.13.11
- **Hardware**: Two Mac Studios connected via Thunderbolt 5
- **Network**: Both machines on same local subnet, firewall disabled
- **Installation Method**: `git clone` + `uv pip install -e .`

## Description
Peer discovery via libp2p mDNS is not functioning on macOS. Two EXO instances running on the same local network cannot discover each other, despite both libp2p swarms initializing successfully and listening on their respective network interfaces.

## Steps to Reproduce
1. Install EXO from source on two macOS Tahoe 26.2 systems
2. Connect systems via Thunderbolt 5 (or same local network)
3. Start EXO on both systems: `python -m exo --no-api -v`
4. Observe that both nodes start successfully and listen on their IPs
5. Wait for mDNS discovery (even with modified 5-second query interval)
6. Result: No peer discovery occurs

## Expected Behavior
- EXO instances should discover each other via libp2p mDNS
- Peers should appear in each other's peer lists
- RDMA connections should establish between discovered peers

## Actual Behavior
- Both EXO instances start successfully
- libp2p swarm initializes and listens on multiple interfaces
- No mDNS service advertisements detected (`dns-sd -B _p2p._udp local.` returns no results)
- No peer discovery events in logs
- No connections established between nodes

### Network Initialization (from logs)
```
RUST: networking task started
RUST: other event NewListenAddr { listener_id: ListenerId(1), address: /ip4/127.0.0.1/tcp/XXXXX }
RUST: other event NewListenAddr { listener_id: ListenerId(1), address: /ip4/169.254.X.X/tcp/XXXXX }
RUST: other event NewListenAddr { listener_id: ListenerId(1), address: /ip4/192.168.X.X/tcp/XXXXX }
Starting node 12D3KooW... (libp2p peer ID)
```

### UDP Ports in Use
Both instances have UDP sockets on port 5353 (mDNS):
```
python3.1  PID  username   35u  IPv4 0xXXXXXXXXXXXXXXXX      0t0  UDP *:5353
python3.1  PID  username   37u  IPv4 0xXXXXXXXXXXXXXXXX      0t0  UDP *:5353
```

[...]

PR #1681, opened 2026-03-07
Warn when libp2p mdns multicast send fails

sblOWPCKCR:
## Motivation

Peer discovery consistently fails on my boxes from time to time. Once it gets into this this state, nothing fixes it short of reboot. I never had Firewall on.

I traced it down to multicast not working. Here's a minimal example to check if you're in this state (should fail with and without --pin):
```rust
use socket2::{Domain, Protocol, Socket, Type};
use std::env;
use std::net::{Ipv4Addr, SocketAddrV4, UdpSocket};

fn main() -> std::io::Result<()> {
    let ip: Ipv4Addr = env::args().nth(1).expect("usage: mdns-repro <src-ip> [--pin]").parse().unwrap();
    let pin = env::args().any(|a| a == "--pin");
    let sock = Socket::new(Domain::IPV4, Type::DGRAM, Some(Protocol::UDP))?;
    sock.bind(&SocketAddrV4::new(ip, 0).into())?;
    if pin {
        sock.set_multicast_if_v4(&ip)?;
    }
    let sock = UdpSocket::from(sock);
    let dst = SocketAddrV4::new(Ipv4Addr::new(224, 0, 0, 251), 5353);
    println!("bound={}, pin_multicast_if={pin}", sock.local_addr()?);
    match sock.send_to(b"x", dst) {
        Ok(n) => println!("sent {n} bytes to {dst}"),
        Err(e) => println!("send_to({dst}) -> {e:?}"),
    }
    Ok(())
}
```

Interestingly, equivalent code works with `/usr/bin/python3`!

Even more interestingly, the above code (and exo) consistenly failed for me in my old-running `tmux` session, but both worked just fine if I created a new `tmux` (or even raw `ssh`) session. 

My hypothesis is that macOS assigns different network policies depending on **something**. Somebody more knowledgable can pick the investigation up.

This PR only makes mDNS errors **visible** - you would know that discovery is non-functional immediately:
```
[ 12:58:25.6102PM | WARNING ] libp2p mDNS multicast send failed. address=192.168.1.10. Peer auto-discovery may not work in this process context. If peers do not form a cluster, relaunch exo from a fresh shell/session (for example outside an existing tmux server).
```

## Changes

Set up tracing for mDNS

## Why It Works

It doesn't solve the problem, but makes it visible

### Manual Testing
I now see this warning when I'm in 'bad' state:
```
[ 12:58:25.6102PM | WARNING ] libp2p mDNS multicast send failed. address=192.168.1.10. Peer auto-discovery may not work in this process context. If peers do not form a cluster, relaunch exo from a fresh shell/session (for example outside an existing tmux server).
```
### Automated Testing
N/A

Potentially related:  #950


AlexCheema:
Sounds potentially similar to issues @ciaranbor was running into with tmux?

ciaranbor:
Yes, this is exactly the same issue. When the login shell that the tmux session was created in closes, discovery breaks and will never work again for that session. Perhaps the daemon stuff @Evanev7 is working on will fix this?

PR #1684, opened 2026-03-09
Fix stale state.runners

ciaranbor:
## Motivation

When a runner shuts down, there's no mechanism to remove it from state.runners, causing stale state to accumulate.

## Changes

- apply.py: apply_runner_status_updated now handles RunnerShutdown status by removing the runner from state instead of adding/updating it. Removed the separate apply_runner_deleted function entirely.
- events.py: Removed the RunnerDeleted event type — runner cleanup is now handled via RunnerStatusUpdated with a RunnerShutdown status.
- Tests: New test_apply_runner_deleted.py with 2 tests: verifying RunnerShutdown removes the runner, and that a normal status update adds it.

## Why It Works

Instead of a separate RunnerDeleted event and coordinating its emission from master/placement, runner cleanup is handled naturally through the existing RunnerStatusUpdated path — a RunnerShutdown status signals removal. This is simpler and requires no changes to master or placement logic.

### Automated Testing

2 new unit tests covering RunnerShutdown removal and normal runner status addition

### DIFF
--- a/src/exo/master/tests/test_placement.py
+++ b/src/exo/master/tests/test_placement.py
@@ -14,7 +14,11 @@ from exo.shared.models.model_cards import ModelCard, ModelId, ModelTask
 from exo.shared.topology import Topology
 from exo.shared.types.commands import PlaceInstance
 from exo.shared.types.common import CommandId, NodeId
-from exo.shared.types.events import InstanceCreated, InstanceDeleted, TaskStatusUpdated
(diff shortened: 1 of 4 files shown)


Evanev7:
like the direction but i trust our event log - could we send RunnerDeleted from the worker end? i'm also not found of replacing assertions with silent ignores

Evanev7 (approved):

PR #1697, opened 2026-03-10
feat: expose max_concurrent_requests in /state endpoint

mlpy0:
## Summary

Exposes the `EXO_MAX_CONCURRENT_REQUESTS` concurrency limit per-node in the `/state` endpoint, making it visible to dashboards and orchestrators without needing to check environment variables directly.

### Changes

- **`NodeIdentity`** (`profiling.py`): Added `max_concurrent_requests: int` field (defaults to `EXO_MAX_CONCURRENT_REQUESTS`)
- **`StaticNodeInformation`** (`info_gatherer.py`): Added `max_concurrent_requests: int` field, populated in `gather()`
- **`apply.py`**: Propagates `max_concurrent_requests` from `StaticNodeInformation` into `NodeIdentity` during state updates

### Motivation

The `/state` endpoint already exposes per-node runtime info (memory, disk, system, network, thunderbolt) but not the concurrency limit. This follows the existing per-node pattern and makes the configured value accessible without SSH or env var inspection.

### Verification

- `uv run ruff check` — 0 errors
- `uv run pytest` — 249 passed

### DIFF
--- a/src/exo/shared/apply.py
+++ b/src/exo/shared/apply.py
@@ -306,6 +306,7 @@ def apply_node_gathered_info(event: NodeGatheredInfo, state: State) -> State:
                     "chip_id": info.chip,
                     "os_version": info.os_version,
                     "os_build_version": info.os_build_version,
+                    "max_concurrent_requests": info.max_concurrent_requests,
                 }
             )
             update["node_identities"] = {
--- a/src/exo/shared/types/profiling.py
+++ b/src/exo/shared/types/profiling.py
@@ -5,6 +5,7 @@ from typing import Literal, Self
 
 import psutil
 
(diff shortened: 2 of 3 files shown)


Evanev7:
while i agree we should expose this information, currently this is a per node setting that requires synchronisation, and until we have that synchronisation it should remain private state.

mlpy0:
> while i agree we should expose this information, currently this is a per node setting that requires synchronisation, and until we have that synchronisation it should remain private state.

Are you planning to disallow different concurrency configurations from working together, similar to versioning?

Evanev7:
pretty much - i'm planning out a configuration system at the moment so it should be per-instance information that can be atomically updated. if you submit two different concurrencies currently things will break horribly!

Issue #1606, opened 2026-02-24
Way to download models to specific location that can be distributed to nodes.


4226181:
Would be incredibly useful if a node with plenty of storage could store the model and allow other nodes on the network to draw from the storage from that node. This would replace the need for each node to have sufficient storage to store the model locally, as many users prioritize performance and processing power over storage. This would also allow for only 1 node to download the model instead of waiting for every node to download it, saving time and data.

AlexCheema:
WIP here: https://github.com/exo-explore/exo/pull/1463

Issue #721, opened 2025-02-20
Feature Request - Add Rsync


Selbyl:
Any thoughts on adding RSync so that if there's already a node in the cluster with the required model you can download the model over the local network to save time?

FlorianHeigl:
technically a torrenting file download would be best to have scaling and increasing efficiency / restartability. 
on windows, that would in fact be available in the OS using BITS, which is normally only used for updates, remote sites ("branch cache") and exploitation. but technically, it could be used for good something good instead.
maybe the  easiest way would be to just allow users to hook into the download commands more easily and allow to have different types of downlaods ("master", "first starter", "proxied", ...). 

i'd immediately just drop in a NFS share, problem over.
rsync would need to be used with --inplace and other ugly flags, it might be a bit horrific unless a model is split into multiple smaller files. cysnc2 would be better for this purpose since it  is more aware there could be a ring of systems that should get a mildly-atomic update. but back when i used that I never tried it with very large files.

heck, even a client-sw shared drive from some NAS would be better, but, all of the above have the problem that exo needs to propagate 'ok this one guy is now downloading, and everyone else just wait and then use this other command to sync up'
Thus, torrenting would be a ton more efficient - the node discocery is already there so it would mean there's a parallel download "from the internets" and a back-hand sharing of what's already been fetched.

Then come along the people that use software commercially and would get fired if they run a torrent client and then we are back to just using rsync.
So yeah, but maybe we can start from the 'have download commands that can be adjusted' and 'have storage methods that can be adfjusted' and 'have status exchange that can be adjusted' and 'have fetch a model from a discovered peer in a way that can be adjusted'.

I think if an implementation starts at that last step it would be best since it then can, by default, fall through to unchanged behaviour and suitable implementations for each step can be added and act the same way.
So we don't make things more fragile, but instead  can have a fail-operational model download.

JoaGamo:
> Then come along the people that use software commercially and would get fired if they run a torrent client

This is straight up a lie. Torrent itself is a p2p file sharing protocol just like using sftp.

Evanev7:
Coming soon:tm:

PR #1699, opened 2026-03-11
feat: peer-to-peer model downloads over LAN

ecohash-co:
## Summary

When multiple nodes need the same model, **every node currently downloads the full model independently from HuggingFace**. This PR adds peer-to-peer model transfer so only one node downloads from HF while others fetch it over the LAN — eliminating redundant internet downloads and cutting cluster startup time roughly in half.

- **PeerFileServer**: Lightweight aiohttp HTTP server on each node (port 52416) that serves model files from the local cache with Range request support
- **PeerAwareShardDownloader**: Wraps `ResumableShardDownloader` in a decorator pattern — checks if any peer already has the model before hitting HuggingFace, falls back to HF on failure
- **Streaming relay**: Followers can download from a peer *while it's still downloading from HF*, via `.partial.meta` companion files that track flushed byte boundaries
- **Zero config**: Enabled by default, disable with `--no-peer-download` or `EXO_PEER_DOWNLOAD_PORT` env var

## How it works

```
Current:   HuggingFace → Node A (full download)
           HuggingFace → Node B (full download)     ← 2× internet bandwidth

Proposed:  HuggingFace → Node A → serves over LAN → Node B
                              ↓                         ↓
                        saves locally              saves locally
```

1. **No leader election needed** — first node to start downloading becomes the de facto seed
2. **No new gossipsub messages** — reuses existing `NodeDownloadProgress` events and topology for peer discovery + IP resolution
3. **Graceful fallback** — if peer transfer fails or is unreachable, the node falls back to HuggingFace with `.partial` resume support
4. **Backend-agnostic** — works with MLX, tinygrad, PyTorch (disk-to-disk transfer, any engine can load)
5. **Network-agnostic** — works over any LAN (Ethernet, WiFi, Thunderbolt)

### Streaming relay detail

The key innovation: while Node A is downloading `model.safetensors` from HF, it writes a companion `.partial.meta` file after each 8MB chunk flush:

```json
{"safe_bytes": 83886080, "total": 4294967296, "etag": "abc123"}
```

Node B's `PeerFileServer` reads this metadata and serves *only the safe byte range* — never unflushed data. Node B polls with Range requests as Node A progresses, receiving chunks as they become available. If Node A dies mid-download, Node B's retry loop times out and falls back to HF, resuming from its own `.partial` file.

## Relationship to PR #1463

PR #1463 implements MLX-specific memory-to-memory transfer via `all_sum` over Thunderbolt. This PR is **complementary**:

| | This PR | PR #1463 |
|---|---|---|
| **Layer** | Disk-to-disk (download time) | Memory-to-memory (load time) |
| **Transport** | HTTP over any LAN | MLX distributed over Thunderbolt |
| **Backends** | All (MLX, tinygrad, PyTorch) | MLX only |
| **Networks** | Any (Ethernet, WiFi, TB) | Thunderbolt only |
| **Code overlap** | None | None |

### New files (4 modules + 1 test)
- `src/exo/download/peer_file_server.py` — HTTP file server with Range + streaming relay support
- `src/exo/download/peer_download.py` — HTTP client for downloading from peers
- `src/exo/download/peer_state.py` — Reads `State.downloads` + `State.topology` to find peers with models
- `src/exo/download/peer_shard_downloader.py` — `ShardDownloader` decorator: try peer, fall back to HF
- `src/exo/download/tests/test_peer_download.py` — 11 tests covering server, client, and edge cases

### Modified files (4 files, minimal changes)
- `src/exo/download/download_utils.py` — Write `.partial.meta` after each chunk flush (+25 lines)
- `src/exo/download/impl_shard_downloader.py` — Factory accepts optional `PeerStateProvider` (+7 lines)
- `src/exo/main.py` — Wire up `PeerFileServer` + `PeerStateProvider`, add `--no-peer-download` flag (+55 lines)
- `src/exo/shared/constants.py` — Add `EXO_PEER_DOWNLOAD_PORT` constant (+3 lines)

## Test plan

- [x] All 11 new peer download tests pass
- [x] All 20 existing download tests still pass (no regressions)
- [x] All 79 worker tests still pass
- [ ] Manual test: 2-node cluster, verify second node downloads from first over LAN
- [ ] Manual test: kill the seed node mid-download, verify follower falls back to HF
- [ ] Manual test: `--no-peer-download` flag disables peer download

Addresses #1257, #721, #1606

### DIFF
--- a/src/exo/download/peer_shard_downloader.py
+++ b/src/exo/download/peer_shard_downloader.py
@@ -1,0 +1,277 @@
+"""Peer-aware shard downloader that tries LAN peers before HuggingFace.
+
+Wraps an existing ShardDownloader and adds a peer-download step: before
+hitting HuggingFace, check if any peer on the LAN already has the model
+(or is downloading it) and fetch from them instead. Falls back to the
+inner downloader (HF) if peer download fails.
+"""
+
+import asyncio
+import time
+from collections.abc import Awaitable
+from datetime import timedelta
+from pathlib import Path
+from typing import AsyncIterator, Callable
+
+from loguru import logger
+
+from exo.download.download_utils import (
+    RepoDownloadProgress,
+    calculate_repo_progress,
+    ensure_models_dir,
+    fetch_file_list_with_cache,
+    is_image_model,
+    resolve_allow_patterns,
+)
+from exo.download.huggingface_utils import filter_repo_objects
+from exo.download.peer_download import (
+    download_file_from_peer,
+    get_peer_file_status,
+)
+from exo.download.peer_state import PeerStateProvider
+from exo.download.shard_downloader import ShardDownloader
+from exo.shared.types.memory import Memory
+from exo.shared.types.worker.downloads import RepoFileDownloadProgress
+from exo.shared.types.worker.shards import ShardMetadata
+
+
+class PeerAwareShardDownloader(ShardDownloader):
+    """ShardDownloader that tries peer download before HuggingFace.
+
+    Decorates an inner ShardDownloader (typically ResumableShardDownloader).
+    On ensure_shard(), checks if any cluster peer already has the model
+    and downloads from them over the LAN. Falls back to the inner
+    downloader if no peer has it or the peer transfer fails.
+    """
+
+    def __init__(
+        self,
+        inner: ShardDownloader,
+        peer_state_provider: PeerStateProvider,
+    ) -> None:
+        self._inner = inner
+        self._peer_state = peer_state_provider
+        self._progress_callbacks: list[
+            Callable[[ShardMetadata, RepoDownloadProgress], Awaitable[None]]
+        ] = []
+
+    def on_progress(
+        self,
+        callback: Callable[[ShardMetadata, RepoDownloadProgress], Awaitable[None]],
+    ) -> None:
+        self._inner.on_progress(callback)
+        self._progress_callbacks.append(callback)
+
+    async def ensure_shard(
+        self, shard: ShardMetadata, config_only: bool = False
+    ) -> Path:
+        if config_only:
+            return await self._inner.ensure_shard(shard, config_only=True)
+
+        model_id = shard.model_card.model_id
+        normalized = model_id.normalize()
+
+        # Check if any peer has this model
+        peers = self._peer_state.get_peers_for_model(normalized)
+        if not peers:
+            logger.debug(f"No peers have {model_id}, downloading from HuggingFace")
+            return await self._inner.ensure_shard(shard, config_only=False)
+
+        # Try each peer (completed peers first)
+        for peer in peers:
+            logger.info(
+                f"Attempting peer download of {model_id} from "
+                f"{peer.ip} (status: {peer.status})"
+            )
+            result = await self._try_peer_download(
+                shard, peer.ip, self._peer_state.peer_download_port, normalized
+            )
+            if result is not None:
+                logger.info(
+                    f"Successfully downloaded {model_id} from peer {peer.ip}"
+                )
+                return result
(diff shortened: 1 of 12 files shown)


src/exo/download/peer_state.py
```diff
@@ -0,0 +1,93 @@
...
+                    peers.append(PeerInfo(node_id=peer_node_id, ip=ip, status=status))
+
+        # Sort: completed peers first
+        peers.sort(key=lambda p: 0 if p.status == "complete" else 1)
+        return peers
+
+    def _resolve_peer_ip(self, peer_node_id: NodeId, state: State) -> str | None:
+        """Resolve a peer's IP address from the topology graph."""
+        try:
+            for conn in state.topology.out_edges(self.node_id):
```
rltakashige: There likely needs to be some sort of link prioritisation, as we wouldn't want a super slow connection for this, etc.

src/exo/download/peer_shard_downloader.py
```diff
@@ -0,0 +1,277 @@
...
+    downloader if no peer has it or the peer transfer fails.
+    """
+
+    def __init__(
+        self,
+        inner: ShardDownloader,
+        peer_state_provider: PeerStateProvider,
+    ) -> None:
+        self._inner = inner
+        self._peer_state = peer_state_provider
```
rltakashige: I'm not sure how I feel about the worker's state being coupled with the download coordinator. Imo, since the download coordinator isn't owned by worker, we could have the worker request for a file from another node.

rltakashige (commented):
Quick pass, not a thorough review. I'm interested to see how this performs.

src/exo/download/peer_state.py
ecohash-co: Agreed — addressed in 921a6bf. Peers are now sorted by:
1. Connection type: RDMA/Thunderbolt connections rank first (lower latency, higher bandwidth)
2. Download status: completed downloads before ongoing ones

The `PeerEndpoint` type now includes a `connection_type` field (`"rdma"` or `"socket"`) populated from the topology graph.

src/exo/download/peer_shard_downloader.py
ecohash-co: Good call — fully refactored in 921a6bf. The architecture is now:

1. **Worker discovers peers** (it owns the State) via a pure function `discover_peers_for_model()`
2. **Worker embeds peers** in the `StartDownload` command as `available_peers: list[PeerEndpoint]`
3. **DownloadCoordinator receives peers** in the command and passes them to the shard downloader
4. **No coupling** between download coordinator and worker state — the command is self-contained

The `PeerStateProvider` class has been replaced with a stateless pure function. `main.py` no longer wires any state accessor lambda.

ecohash-co:
Rebased on current `main` (through 7b4161) — clears the DIRTY state and resolves conflicts from the 79 commits of drift since March.

Grounding this in something fresh: spent this afternoon watching a 2-node Mac Studio cluster redownload a 227 GB MiniMax-M2.7 checkpoint independently on both nodes over a shared 1 Gbps uplink. With this PR's peer transfer, the second node pulls from the first over TB5 (~5 min at 650 MB/s SSH-capped; would be higher with the cleartext HTTP transport this PR actually uses) instead of a second round-trip to HF.

The streaming relay design (`.partial.meta` published after each 8 MB flush) is exactly what makes this safe when both nodes start at the same time — Node B races Node A to HF but converges on Node A as soon as any bytes are flushed.

Happy to address any review feedback. @rltakashige — back in March you mentioned interest in perf numbers; I can run atlas↔epi TB5 benchmarks and post them here if that'd help a decision. Also glad to squash the two commits or keep them split, whichever reads better.

PR #1702, opened 2026-03-11
add script section to delete APFS Volume and Container

### LINKED ISSUE #1386: EXO macOS Uninstall Leaves APFS Container Behind
Summary

After uninstalling EXO on macOS using the official uninstaller script, a dedicated APFS container and volume named EXO remain on disk. The software components are fully removed, but the disk structure created by the application is not cleaned up.

What Was Removed Successfully
	•	Background network daemon
	•	LaunchDaemon service
	•	Network configuration changes
	•	Log files
	•	EXO network location

No EXO processes, services, or network extensions remain active.

What Remains

An APFS container (~890 MB) with a single volume:
	•	Container: disk6
	•	Volume Name: EXO
	•	Role: None
	•	Mount Point: Not mounted
	•	Not part of system volume group

This container is separate from the macOS system container and does not affect system operation, but it occupies disk space and cannot be removed from normal macOS mode.

Disk Structure Evidence
`diskutil apfs list`

Shows:
	•	Main system container (≈1TB) with standard macOS volumes
	•	Separate small container (~890MB) containing only the EXO volume

Issue

The official uninstaller does not remove the APFS container created by EXO.
Users are left with a persistent disk container that requires Recovery Mode or advanced disk operations to remove.

nightguarder:
## Motivation

Fixes #1386.
After uninstalling EXO on macOS using the official uninstaller script, a dedicated APFS container and volume named `EXO` (~890 MB) remains on the disk. This occupies disk space and leaves behind a tangling container that isn't cleaned up automatically.

## Changes

Minimal update to the`@app/EXO/uninstall-exo.sh` script to include an APFS volume and container cleanup step.

- Added logic using `diskutil` to search for an APFS volume named `EXO`, find its parent container, and securely delete the container.

## Why It Works

By dynamically querying `diskutil apfs list` for the `EXO` volume and executing `diskutil apfs deleteContainer` (or `deleteVolume` as a fallback), we guarantee that the isolated disk structure created during the application's installation is fully removed alongside the other system modifications.

## Test Plan

1. Run the updated `sudo ./app/EXO/uninstall-exo.sh` script on a macOS machine with an existing EXO installation.

2. Verify via `diskutil apfs list` and `diskutil list` that the `EXO` volume and its parent APFS container have been completely removed from the disk structure.
3. Verify the script continues gracefully if the volume does not exist.
4. 
### Automated Testing
<!-- Describe changes to automated tests, or how existing tests cover this change —>
<!-- - —>

### DIFF
--- a/app/EXO/uninstall-exo.sh
+++ b/app/EXO/uninstall-exo.sh
@@ -136,6 +136,27 @@ done
 
 echo ""
 echo "========================================"
+
+# Remove EXO APFS Volume and Container
+if diskutil apfs list | grep -q "Volume Name:.*EXO"; then
+  echo_info "Locating EXO APFS volume..."
+  EXO_DISK=$(diskutil list | grep " EXO " | awk '{print $NF}')
+  if [[ -n "$EXO_DISK" ]]; then
+    EXO_CONTAINER=$(diskutil info "$EXO_DISK" | grep "Part of Whole:" | awk '{print $4}')
+    if [[ -n "$EXO_CONTAINER" ]]; then
+       echo_info "Deleting EXO APFS container $EXO_CONTAINER..."
+       diskutil apfs deleteContainer "$EXO_CONTAINER" 2>/dev/null || true
+    else
+       echo_info "Deleting EXO APFS volume $EXO_DISK..."
+       diskutil apfs deleteVolume "$EXO_DISK" 2>/dev/null || true
+    fi
+  fi
+else
+  echo_warn "No EXO APFS volume found. You may need to delete it manually."
+fi
+
+echo ""
+echo "========================================"
 echo_info "EXO uninstall complete!"
 echo "========================================"
 echo ""


app/EXO/uninstall-exo.sh
```diff
@@ -134,6 +134,27 @@ for app_path in "/Applications/EXO.app" "$HOME/Applications/EXO.app"; do
...
+# Remove EXO APFS Volume and Container
+if diskutil apfs list | grep -q "Volume Name:.*EXO"; then
+  echo_info "Locating EXO APFS volume..."
+  EXO_DISK=$(diskutil list | grep " EXO " | awk '{print $NF}')
+  if [[ -n "$EXO_DISK" ]]; then
+    EXO_CONTAINER=$(diskutil info "$EXO_DISK" | grep "Part of Whole:" | awk '{print $4}')
+    if [[ -n "$EXO_CONTAINER" ]]; then
+       echo_info "Deleting EXO APFS container $EXO_CONTAINER..."
+       diskutil apfs deleteContainer "$EXO_CONTAINER" 2>/dev/null || true
+    else
```
Evanev7, line 149: this is a pretty bad idea IMO - if it finds a disk which is part of say.. disk1.. it will attempt to delete your whole disk.

nightguarder:
I think the issue was likely because a corrupt exo failure or Sparkle-Updater. Not as the original user mentioned with uninstalling. 
The DMG is used either for the initial installation or downloaded in the background by the Sparkle Auto-Updater.

nightguarder:
@Evanev7 Can you close the issue? #1386 
> The DMG is used either for the initial installation or downloaded in the background by the Sparkle Auto-Updater.

The likely cause for the issue was that the Auto-Updater didn’t unmount the “~890 MB” container. 
Safe to say this is not an issue with EXO  as it has never ran any partitioning command.

Evanev7:
thanks for looking into it!

PR #1703, opened 2026-03-11
Reject duplicate instance placements

mlpy0:
## Summary

Calling `POST /place_instance` or `POST /instance` multiple times with the same model, nodes, sharding mode, and instance meta currently creates duplicate instances. Each one consumes GPU memory and compute for no benefit.

This adds a duplicate check in `placement.py` where both code paths converge:
- `place_instance()` — checked after shard assignments are computed (so single-node overrides are already applied)
- `add_instance_to_placements()` — replaces the existing `# TODO: validate against topology`

Two instances are considered duplicates when they share all four of: model ID, node set, sharding (Pipeline vs Tensor), and instance meta (MlxRing vs MlxJaccl). `ValueError` is the standard validation error in this file — the command processor and preview endpoints already handle it.

## Test plan

- [x] `test_place_instance_rejects_duplicate_placement` — same params twice → `ValueError`
- [x] `test_add_instance_rejects_duplicate_placement` — different instance IDs, same placement → `ValueError`
- [x] `test_place_instance_allows_different_model_same_nodes` — different models on same topology → succeeds
- [x] `test_place_instance_allows_same_model_different_sharding` — Pipeline vs Tensor on same model → succeeds
- [x] All 19 placement tests pass, basedpyright 0 errors, ruff clean

### DIFF
--- a/src/exo/master/placement.py
+++ b/src/exo/master/placement.py
@@ -1,7 +1,7 @@
 import random
 from collections.abc import Mapping
 from copy import deepcopy
-from typing import Sequence
+from typing import Sequence, TypeAlias
 
 from exo.master.placement_utils import (
     Cycle,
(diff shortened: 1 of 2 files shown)


Evanev7:
there are legitimate reasons to create two identical instances on the same device. i don't see what issue this solves - we should (and i believe already do) reject duplicate instanceids (which is clearly an error) however.

mlpy0:
Then my bad. I was adding a recovery mechanism to my system in case of rdma issues etc. and i thought adding conflict safety is nice to have.

Evanev7:
no worries! it's a reasonable suggestion and implementation, im just not confident it fixes a real problem

PR #1714, opened 2026-03-12
fix: shield macmon cleanup from cancellation to prevent orphaned process

ecohash-co:
## Summary

Fixes a bug where macmon metrics (GPU usage, temperature, power, RAM) freeze permanently in the dashboard while non-macmon metrics (disk usage) continue updating normally.

**Root cause: asyncio subprocess pipe transport flow control stall**

Observed on a 2-node Mac Studio M3 Ultra cluster (Python 3.13.12, anyio 4.11.0, macOS with kqueue) running EXO.app for ~36 hours. Diagnostics confirmed:

1. **Only macmon monitoring is stuck** — disk metrics from the same InfoGatherer continue updating, proving the Worker, EventRouter, and API pipelines are healthy
2. **macmon IS producing data** — its stdout pipe is full at exactly 65536 bytes (64KB OS buffer), and macmon is blocked on write at 0% CPU
3. **The pipe read-end FD is still open** — the exo process holds it, but asyncio isn't reading from it
4. **The stale GPU value (0.21) is wrong** — should be ~1.0 (matching the other node under identical load)

This is consistent with asyncio's subprocess pipe transport getting stuck in flow control: `pause_reading()` is called when the internal buffer exceeds the high-water mark, but `resume_reading()` is never called, permanently deregistering the FD from kqueue. The `receive()` coroutine waits forever for data asyncio will never deliver.

```
$ lsof -p <macmon_pid>
macmon  74691  FD=1  PIPE  SIZE=65536  ->0x5ae78ecf376ccd0e  # full pipe

$ lsof -p <exo_pid> | grep <pipe_id>
exo     74689  FD=47  PIPE  SIZE=65536  ->0xd129da474c1340f7  # read end held open, not consumed
```

Note: anyio 4.11's `Process.aclose()` already uses `CancelScope(shield=True)` for cleanup during cancellation — this is NOT an election cleanup issue (confirmed by @ciaranbor's testing).

**Fix:** Replace the `async for` iteration with an explicit `receive()` inside `fail_after()`. If macmon produces no output for 10× its configured interval (minimum 30s), `TimeoutError` fires, `open_process` cleanup kills macmon and tears down the stale transport, and the loop restarts with a fresh subprocess and fresh asyncio transport.

## Test plan

- [x] All 246 existing tests pass
- [ ] Verify macmon restarts after simulated pipe stall (e.g., `SIGSTOP` macmon, wait for timeout, confirm restart and metrics resume)
- [ ] Long-running soak test on multi-node cluster to confirm the fix prevents recurrence

### DIFF
--- a/src/exo/utils/info_gatherer/info_gatherer.py
+++ b/src/exo/utils/info_gatherer/info_gatherer.py
@@ -529,6 +529,9 @@ class InfoGatherer:
         if self.macmon_interval is None:
             return
         # macmon pipe --interval [interval in ms]
+        # Timeout: if macmon produces no output for this many seconds, restart it.
+        # macmon writes every macmon_interval seconds, so 10x that is generous.
+        read_timeout = max(self.macmon_interval * 10, 30)
         while True:
             try:
                 async with await open_process(
@@ -542,10 +545,15 @@ class InfoGatherer:
                     if not p.stdout:
                         logger.critical("MacMon closed stdout")
                         return
-                    async for text in TextReceiveStream(
-                        BufferedByteReceiveStream(p.stdout)
-                    ):
+                    stream = TextReceiveStream(BufferedByteReceiveStream(p.stdout))
+                    while True:
+                        with fail_after(read_timeout):
+                            text = await stream.receive()
                         await self.info_sender.send(MacmonMetrics.from_raw_json(text))
+            except TimeoutError:
+                logger.warning(
+                    f"MacMon produced no output for {read_timeout}s, restarting"
+                )
             except CalledProcessError as e:
                 stderr_msg = "no stderr"
                 stderr_output = cast(bytes | str | None, e.stderr)


ciaranbor:
I was not able to reproduce this. I tried spamming master node restarts but the processes were cleaned up and restarted on re-election every time

ecohash-co:
@ciaranbor Thanks for testing. You're right — I dug into anyio 4.11's `Process.aclose()` and it already uses `CancelScope(shield=True)` for pipe cleanup and `kill()+wait()` on cancellation ([`_asyncio.py:1072-1088`](https://github.com/agronholm/anyio/blob/4.11.0/src/anyio/_backends/_asyncio.py#L1072-L1088)). So election-triggered task cancellation shouldn't orphan macmon. My original theory about `__aexit__` under cancellation was wrong.

I've revised the fix to address the actual observable failure: **no read timeout on macmon output**. The diagnostics from our cluster are definitive:

```
# macmon stdout pipe completely full (64KB OS buffer), blocked on write
$ lsof -p <macmon_pid>
macmon  74691  FD=1  PIPE  SIZE=65536  ->0x5ae78ecf376ccd0e

# exo process still holds the read end, not consuming
$ lsof -p <exo_pid> | grep <pipe_id>
exo     74689  FD=47  PIPE  SIZE=65536  ->0xd129da474c1340f7

# macmon at 0% CPU (blocked on write), running for 2+ days
$ ps -p 74691 -o pid,stat,etime,%cpu
74691   S   02-02:03:21   0.0
```

Something caused the `async for text in TextReceiveStream(...)` to stop reading. The exact trigger is unclear — possibly a macOS kernel pipe edge case, a transient anyio scheduling issue, or something else entirely. But the current code has **no recovery path**: once reading stalls, metrics freeze permanently.

The revised fix replaces the `async for` with an explicit `receive()` inside `fail_after(read_timeout)`. If macmon produces no output for 10x its configured interval (minimum 30s), the process is killed and restarted. Simple, defensive, and addresses the symptom regardless of root cause.

ecohash-co:
Updated with a corrected analysis — the original theory about election cleanup was wrong (thanks for catching that @ciaranbor). Here's what's actually happening:

**The problem is NOT elections.** anyio 4.11's `Process.aclose()` already shields cleanup properly. The issue is an asyncio subprocess pipe transport stall.

I verified that non-macmon metrics (disk usage) from the **same InfoGatherer** continue updating on the affected node — proving the InfoGatherer, Worker._forward_info, EventRouter, and API pipelines are all healthy. Only the `_monitor_macmon` task is stuck.

macmon is actively producing data (its stdout pipe is full at 64KB), but asyncio's transport isn't delivering it to the `receive()` coroutine. This is consistent with the transport's flow control getting stuck: `pause_reading()` fired but `resume_reading()` never did, deregistering the FD from kqueue permanently. The coroutine sits at `receive()` waiting for data that the OS has but asyncio won't serve.

The fix wraps `receive()` in `fail_after(30)`. When the timeout fires, it interrupts the stuck coroutine, `open_process.__aexit__` tears down the stale subprocess/transport, and the loop restarts with a fresh one.

You can verify the fix works by sending `SIGSTOP` to the macmon process to simulate a stall — the timeout should fire and restart it.

Evanev7:
i literally made this exact change before and it didn't work for mysterious reasons! i'm going to test this now...

Evanev7:
literally identical to #1618 but what the hell glad it's back

PR #1719, opened 2026-03-13
feat: add --memory-fraction CLI arg to cap available memory for placement

mlpy0:
## Summary

- Adds `--memory-fraction` CLI argument (0.1–1.0, default 1.0) that scales reported `ram_available` at the MemoryUsage creation layer
- Both memory code paths are covered: `from_psutil()` (non-macOS / macmon fallback) and macmon's `from_raw()` (macOS)
- Also configurable via `EXO_MEMORY_FRACTION` env var
- Placement logic sees reduced available memory → assigns fewer layers per node → leaves headroom for KV cache, activations, and intermediate tensors

## Motivation

During inference, KV cache, activations, and intermediate tensors grow beyond what the placement system accounts for. By treating available memory as a fraction of reality (e.g., 90%), nodes get fewer layers and retain a safety buffer. This is a common pattern (TensorFlow's `gpu_memory_fraction`, PyTorch's `max_split_size_mb`, etc.).

## Changes

| File | Change |
|------|--------|
| `src/exo/main.py` | Add `memory_fraction` field + Pydantic validator + argparse arg + env var propagation |
| `src/exo/shared/types/profiling.py` | Add `memory_fraction` param to `from_psutil()`, scale `ram_available` |
| `src/exo/utils/info_gatherer/info_gatherer.py` | Read `EXO_MEMORY_FRACTION` env var, pass to `from_psutil()` |
| `src/exo/utils/info_gatherer/macmon.py` | Read `EXO_MEMORY_FRACTION` env var, scale `ram_available` in `from_raw()` |

## Usage

```bash
# Use 90% of available memory for placement decisions
exo --memory-fraction 0.9

# Or via environment variable
EXO_MEMORY_FRACTION=0.85 exo
```

## Test plan

- [x] `uv run ruff check` — all checks passed
- [x] `uv run pytest` — 249 passed
- [ ] Manual: run `exo --memory-fraction 0.9 -v` and check `/state` endpoint to confirm reported memory is ~90% of actual

### DIFF
--- a/src/exo/utils/info_gatherer/info_gatherer.py
+++ b/src/exo/utils/info_gatherer/info_gatherer.py
@@ -466,7 +466,9 @@ class InfoGatherer:
             if override_memory_env
             else None
         )
-        memory_fraction = min(1.0, max(0.1, float(os.getenv("EXO_MEMORY_FRACTION", "1.0"))))
+        memory_fraction = float(os.getenv("EXO_MEMORY_FRACTION", "1.0"))
+        if not 0.1 <= memory_fraction <= 1.0:
+            raise ValueError(f"EXO_MEMORY_FRACTION must be between 0.1 and 1.0, got {memory_fraction}")
         if self.memory_poll_rate is None:
             return
         while True:
--- a/src/exo/utils/info_gatherer/macmon.py
+++ b/src/exo/utils/info_gatherer/macmon.py
@@ -50,7 +50,9 @@ class MacmonMetrics(TaggedModel):
 
     @classmethod
     def from_raw(cls, raw: RawMacmonMetrics) -> Self:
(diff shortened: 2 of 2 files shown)


Evanev7:
i think OVERRIDE_MEMORY_MB handles this currently, i don't see the benefit of a second debug override.

mlpy0:
OVERRIDE_MEMORY_MB replaces ram_available with a static value, so it stops tracking actual usage. If I set it to 10 GB and the model consumes 8 GB, the system still reports 10 GB free and could over-allocate.

It also only applies to the psutil path. On macOS with macmon installed, it's ignored.

A fraction-based approach scales the real vm.available each poll, so it tracks consumption while keeping a buffer.

PR #1721, opened 2026-03-13
Fix download coordinator race condition

ciaranbor:
## Motivation

Nodes crash with BrokenResourceError during elections when in-flight downloads try to send on a closed event channel.

## Changes

- Catch BrokenResourceError in all event-sending paths, treating it as a graceful shutdown
- Cancel active download tasks eagerly in shutdown() to shrink the race window

## Why It Works

We expect receives on the stale event channel, not a problem


Evanev7:
are the try: excepts actually necessary here? cancelling active downloads should be sufficient

PR #1725, opened 2026-03-14
feat: suppress EOS tokens during tool call generation

dale-lakes:
## Summary

- Prevents models from emitting EOS mid-tool-call, which causes partial tool call tokens to be dumped as raw text to the client
- Adds a stateful logits processor (`ToolCallEosSuppressor`) that masks EOS token logits to `-1e9` while inside a `<|tool_calls_section_begin|>` ... `<|tool_calls_section_end|>` block
- Wired into both `mlx_generate` (single request) and `ExoBatchGenerator` (batch), following the same pattern as existing thinking state tracking

## Problem

When a thinking model (e.g. Kimi K2.5) generates a tool call with a very long argument (such as writing a large file), the model can choose to emit EOS before the closing `<|tool_calls_section_end|>` token. When this happens, the tool call parser in `model_output_parsers.py` hits the `finish_reason is not None` check at line 369 and falls back to yielding the partial tool call as raw text:

```
tool call parsing interrupted, yield partial tool call as text
```

The client then receives raw special tokens (`<|tool_calls_section_begin|><|tool_call_begin|>functions.write:87<|tool_call_argument_begin|>...`) instead of a properly parsed tool call response.

## Approach

Inspired by [vLLM's guided decoding](https://docs.vllm.ai/en/latest/features/tool_calling/) which uses logit masking to enforce structured outputs. Our implementation is simpler — just a boolean flag and EOS masking, no full grammar engine needed:

1. `ToolCallEosSuppressor` — a logits processor that sets all EOS token logits to `-1e9` when `in_tool_call` is `True`
2. The generation loop toggles `in_tool_call` when it decodes `tool_call_start` / `tool_call_end` tokens (same text-matching approach as existing `in_thinking` tracking)
3. Only activated for models that have `tool_call_start` set on the tokenizer (currently Kimi models)

## Test plan

- [ ] Test with Kimi K2.5 generating a very long tool call argument (e.g. writing a large markdown file via a `write` tool)
- [ ] Verify normal tool calls still parse correctly
- [ ] Verify non-tool-call generation (pure text, thinking) is unaffected
- [ ] Verify models without `tool_call_start` (non-Kimi) are unaffected

### DIFF
--- a/src/exo/worker/engines/mlx/generator/batch_generate.py
+++ b/src/exo/worker/engines/mlx/generator/batch_generate.py
@@ -30,6 +30,7 @@ from exo.worker.engines.mlx.cache import (
 )
 from exo.worker.engines.mlx.constants import DEFAULT_TOP_LOGPROBS, MAX_TOKENS
 from exo.worker.engines.mlx.generator.generate import (
+    ToolCallEosSuppressor,
     ban_token_ids,
     eos_ids_from_tokenizer,
     extract_top_logprobs,
@@ -59,6 +60,7 @@ class _EngineTask:
     cache_snapshots: list[CacheSnapshot] | None
     detokenizer: StreamingDetokenizer
     on_generation_token: Callable[[], None] | None = None
+    eos_suppressor: ToolCallEosSuppressor | None = None
     generated_text_parts: list[str] = field(default_factory=list)
     potential_stop_sequence_text: str = ""
     completion_tokens: int = 0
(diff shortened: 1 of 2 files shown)


Evanev7:
did you hit token limits or another eos? if the model is generating incorrectly then this is a worthwhile fix but if you hit max_tokens your tool call failing is correct behaviour & should be the caller / harness' responsibility imo

dale-lakes:
You're right — after further investigation, the tool call interruptions we were seeing were caused by insufficient `max_tokens` (the client was sending 4096, which wasn't enough for long tool call arguments like writing large files). Bumping the output token budget on the client side resolved it.

We never confirmed a case where the model genuinely emitted EOS mid-tool-call while still having token budget remaining. Closing this since the fix belongs on the caller side, not in the inference engine.

If someone later confirms a genuine EOS-mid-tool-call issue (model choosing to stop while still having budget), this approach could be revisited.

PR #1733, opened 2026-03-16
Delete bound runners when deleting an instance

### LINKED ISSUE #1732: Deleting an instance leaves stale runner state in master state
# Issue: Deleting an instance leaves stale runner state in master state

## Summary

Deleting an instance can remove it from authoritative placement state while leaving the instance's runners behind in authoritative runner state.

That leaves EXO in a split lifecycle condition where:

- the instance is gone
- the runners are still present
- health and readiness can stay degraded because of deleted-instance runner state

## Why This Matters

Instance deletion is used for cleanup and rollback. If deletion does not also remove the runner state owned by that instance, EXO can remain unhealthy after cleanup appears to have succeeded.

This makes operator recovery unreliable and breaks the expectation that instance deletion is a full lifecycle teardown.

## Expected Behavior

Deleting an instance should also delete all runner state bound to that instance.

After deletion:

- `state.instances` should no longer contain the instance
- `state.runners` should no longer contain runners owned by the deleted instance
- health and readiness should not remain red because of deleted-instance runners

## Actual Behavior

The instance can disappear from placement state while its runners remain in authoritative runner state.

## Scope

This is not cluster-specific. Any EXO deployment that uses instance deletion for cleanup or rollback can hit it.

## Suggested Fix

When generating deletion transition events, emit `RunnerDeleted` for every runner bound to the instance before emitting `InstanceDeleted`.

## Candidate Files

- `tools/exo/src/exo/master/placement.py`
- `tools/exo/src/exo/shared/apply.py`
- `tools/exo/src/exo/master/main.py`

## Test Coverage That Should Exist

- deleting an instance emits `RunnerDeleted` for all bound runners
- applying those events removes the runners from authoritative state
- health no longer reports deleted-instance runners

## Reference Implementation

Genesis fork commit:

- `e7664679d218da6db42577190b2198f125dae4a2`

funnyjas03:
Closes #1732

Merge note: suggested order `1/6` in the lifecycle hardening series.

## Summary

This PR fixes a lifecycle asymmetry in instance teardown.

Before this change, EXO could delete an instance from placement state while leaving the instance's runners in authoritative runner state. That could keep health and readiness degraded even though cleanup appeared to have succeeded.

This change makes instance deletion emit `RunnerDeleted` for all runners bound to the instance before `InstanceDeleted`.

## Problem

Instance lifecycle and runner lifecycle were not being torn down together. In practice this meant:

- the instance could disappear from `state.instances`
- its runners could remain in `state.runners`
- operator cleanup and rollback could leave stale failed state behind

## Change

- emit `RunnerDeleted` for each runner bound to a deleted instance
- preserve the existing deletion path instead of creating a separate cleanup mechanism

## Why This Is Safe

The patch narrows teardown behavior to match EXO's existing ownership model. It does not add new policy. It only makes deletion clean up the child state the deleted instance already owns.

## Tests

- deleting an instance emits runner cleanup events for all bound runners
- applying those events removes runner state from authoritative state

## Validation

Validated as part of a larger lifecycle patchset on a live 4-node Apple Silicon cluster. The relevant cluster-level outcome was that failed placement cleanup no longer left stale failed runner state behind after teardown.

### DIFF
--- a/src/exo/master/placement.py
+++ b/src/exo/master/placement.py
@@ -26,6 +26,7 @@ from exo.shared.types.events import (
     Event,
     InstanceCreated,
     InstanceDeleted,
+    RunnerDeleted,
     TaskStatusUpdated,
 )
 from exo.shared.types.memory import Memory
@@ -233,6 +234,7 @@ def get_transition_events(
     # find instances to delete
     for instance_id in current_instances:
         if instance_id not in target_instances:
+            instance = current_instances[instance_id]
             for task in tasks.values():
                 if task.instance_id == instance_id and task.task_status in [
                     TaskStatus.Pending,
@@ -244,6 +246,13 @@ def get_transition_events(
                             task_id=task.task_id,
                         )
                     )
+
+            for runner_id in instance.shard_assignments.runner_to_shard:
+                events.append(
+                    RunnerDeleted(
+                        runner_id=runner_id,
+                    )
+                )
 
             events.append(
                 InstanceDeleted(
--- a/src/exo/master/tests/test_placement.py
+++ b/src/exo/master/tests/test_placement.py
@@ -17,6 +17,7 @@ from exo.shared.types.common import CommandId, NodeId
 from exo.shared.types.events import (
     InstanceCreated,
     InstanceDeleted,
+    RunnerDeleted,
     TaskStatusUpdated,
 )
 from exo.shared.types.memory import Memory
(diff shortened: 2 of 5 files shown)


funnyjas03:
Series merge guide:\n\n1. #1733 teardown symmetry\n2. #1734 startup stalls become terminal failure\n3. #1737 phase-aware readiness reporting\n4. #1738 durable failure propagation across closed channels\n5. #1741 atomic rollback after shard failure\n6. #1742 orphan-runner reconciliation\n\nOnly #1742 is intentionally stacked: it depends on #1733 because upstream  does not yet have . The others are intended to be independently reviewable in the order above.

funnyjas03:
Series merge guide:

1. #1733 teardown symmetry
2. #1734 startup stalls become terminal failure
3. #1737 phase-aware readiness reporting
4. #1738 durable failure propagation across closed channels
5. #1741 atomic rollback after shard failure
6. #1742 orphan-runner reconciliation

Only #1742 is intentionally stacked: it depends on #1733 because upstream main does not yet have RunnerDeleted. The others are intended to be independently reviewable in the order above.

PR #1742, opened 2026-03-16
Reconcile orphan runner state in master state

### LINKED ISSUE #1739: Master state retains orphan runner statuses after instance teardown
# Issue: Master state retains orphan runner statuses after instance teardown

## Summary

EXO can retain runner state that no longer belongs to any live instance. Even after deletion and rollback are improved, those orphan runner records can keep health and readiness degraded indefinitely.

## Why This Matters

Child state without a live parent needs a bounded reconciliation path. Without one, EXO can carry stale runner status forever after unusual teardown paths.

## Expected Behavior

If a runner is not referenced by any live instance for a bounded interval, EXO should delete that orphan runner state.

If the runner becomes referenced again before the grace period expires, cleanup should be cancelled.

## Actual Behavior

Unreferenced runner state can survive indefinitely.

## Scope

This is a general state-convergence issue, not a Genesis-specific operational quirk.

## Suggested Fix

- track orphan runners in master state
- apply a bounded TTL before deleting them
- cancel cleanup if the runner becomes referenced again
- make `RunnerDeleted` application idempotent so reconciliation can overlap with explicit teardown safely

## Candidate Files

- `tools/exo/src/exo/master/main.py`
- `tools/exo/src/exo/shared/apply.py`

## Test Coverage That Should Exist

- orphan runners are deleted after the grace period
- re-referenced runners are preserved
- `RunnerDeleted` application is idempotent

## Reference Implementation

Genesis fork commit:

- `b8ae577fd2953c89b9b9a990be8eb3eea7a24fdc`

funnyjas03:
Closes #1739

Merge note: suggested order `6/6` in the lifecycle hardening series.

Dependency note: this draft currently depends on #1733 because upstream `main` does not yet have `RunnerDeleted`. Once #1733 lands, this PR can be rebased to a narrow orphan-reconciliation-only diff.

## Summary

This PR adds a bounded reconciliation path for orphan runner state.

Before this change, EXO could retain runner records that no longer belonged to any live instance. Those orphan runners could continue degrading health and readiness even after the original instance had been deleted or rolled back.

This change tracks orphan runners in master state and emits `RunnerDeleted` after a grace period if they remain unreferenced.

## Problem

Even with better deletion and rollback, unusual teardown paths can still leave unowned child state behind. EXO needs a bounded way to converge that stale runner state away.

## Change

- track orphan runners locally in master state
- delete them after a bounded TTL if they remain unreferenced
- cancel deletion if the runner becomes referenced again
- make `apply_runner_deleted()` idempotent so reconciliation is safe alongside explicit cleanup

## Why This Is Safe

This is conservative reconciliation, not aggressive garbage collection. It only applies to runner state that has no live owning instance for the full grace period.

## Tests

- orphan runners are deleted after the grace period
- runners that regain a live owner before the TTL are preserved
- `apply_runner_deleted()` remains idempotent

## Validation

Validated as part of the full lifecycle patchset on a live cluster. After a real `397B` startup failure, the cluster converged back to `ready=true`, `api_verdict=PASS`, and `failed_runners=[]`.

### DIFF
--- a/src/exo/master/main.py
+++ b/src/exo/master/main.py
@@ -1,3 +1,4 @@
+import os
 from datetime import datetime, timedelta, timezone
 
 import anyio
@@ -38,6 +39,7 @@ from exo.shared.types.events import (
     LocalForwarderEvent,
     NodeGatheredInfo,
     NodeTimedOut,
+    RunnerDeleted,
     TaskCreated,
     TaskDeleted,
     TaskStatusUpdated,
@@ -60,9 +62,14 @@ from exo.shared.types.tasks import (
     TextGeneration as TextGenerationTask,
 )
 from exo.shared.types.worker.instances import InstanceId
+from exo.shared.types.worker.runners import RunnerId
 from exo.utils.channels import Receiver, Sender
 from exo.utils.event_buffer import MultiSourceBuffer
 from exo.utils.task_group import TaskGroup
+
+ORPHAN_RUNNER_TTL_SECONDS = float(
+    os.environ.get("EXO_ORPHAN_RUNNER_TTL", "15")
+)
 
 
 class Master:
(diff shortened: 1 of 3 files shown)


funnyjas03:
Reviewer note: this branch is intentionally stacked on #1733. Once #1733 lands, I can rebase this to a narrow orphan-reconciliation-only diff immediately.

PR #1757, opened 2026-03-19
fix: return error message when tool call parsing fails

dale-lakes:
## Summary

- When tool call parsing fails (malformed syntax from the model), yield a descriptive error message instead of raw special-token markup
- Clients can surface this error back to the model, allowing it to self-correct instead of repeating the same malformed call

## Problem

When a model produces a tool call with malformed syntax (e.g., wrong separator character, missing argument delimiter), `parse_tool_calls` yields the raw markup (`<|tool_calls_section_begin|><|tool_call_begin|>functions.shell!166...`) as plain assistant text. Clients like Goose don't recognize this as a failed tool call, the model never receives error feedback, and it repeats the same malformed call indefinitely.

## Fix

Replace the raw markup yield with a clear error message:
```
[tool call error] A tool call was attempted but could not be parsed due to malformed syntax. Please retry using the correct tool call format.
```

The raw markup is still logged at WARNING level for debugging. The error text flows through as a normal `TokenChunk`, so it's compatible with all existing OpenAI-compatible clients.

## Test plan

- [x] Failed parse yields error message, not raw markup
- [x] Valid tool calls still parse and yield `ToolCallResponse` normally
- [x] Non-tool-call text passes through unchanged
- [x] All 42 existing runner tests pass


Evanev7:
perhaps we should update the finish reason to "error" instead? i'm not keen on replacing the text content with something the model didn't produce.

dale-lakes:
Good call — updated to use `finish_reason="error"` and preserve the original text. This sends an `ErrorChunk` to the client, which matches the same error format used by vLLM and SGLang.

Evanev7:
closing in favour of #1785

PR #1758, opened 2026-03-19
fix: format ErrorChunk as valid OpenAI streaming chunk

dale-lakes:
## Summary

- Format `ErrorChunk` SSE events as valid OpenAI `ChatCompletionResponse` chunks instead of a non-standard `{"error": {...}}` shape
- Fix the non-streaming path to return the error as content instead of raising a `ValueError`

## Problem

When Exo encounters an error during inference (OOM, internal failure, etc.), the streaming path sends:

```
data: {"error": {"message": "...", "type": "InternalServerError", "code": 500}}
```

This doesn't match the OpenAI streaming chunk schema (`{"choices": [...], "object": "chat.completion.chunk", ...}`). Any OpenAI-compatible client (Goose, LiteLLM, etc.) attempting to deserialize this as a `StreamingChunk` will fail with a "Stream decode error", crashing the client session.

The non-streaming path has a similar issue — it raises a `ValueError` instead of returning a parseable response.

## Fix

**Streaming**: Format the error as a standard `ChatCompletionResponse` with the error message as assistant content and `finish_reason: "stop"`. The client receives a valid chunk it can parse and surface to the user/model.

**Non-streaming**: Append the error as content text and set `finish_reason: "stop"` instead of raising, so the response is still a valid `ChatCompletionResponse`.

## Test plan

- [x] All 42 runner tests pass
- [x] All 84 master tests pass
- [x] Type checking passes (basedpyright)
- [x] Linting passes (ruff)

### DIFF
--- a/src/exo/master/adapters/chat_completions.py
+++ b/src/exo/master/adapters/chat_completions.py
@@ -10,8 +10,6 @@ from exo.shared.types.api import (
     ChatCompletionMessageText,
     ChatCompletionRequest,
     ChatCompletionResponse,
-    ErrorInfo,
-    ErrorResponse,
     FinishReason,
     Logprobs,
     LogprobsContentItem,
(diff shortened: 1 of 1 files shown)


Evanev7:
i'd argue we should throw an http error here instead of faking a response chunk

dale-lakes:
You're right. Looked into how vLLM, SGLang, and the OpenAI Python client handle this — sending `{"error": {...}}` as an SSE data event is the de facto standard. The OpenAI client explicitly checks for an `error` key in parsed SSE data and raises `APIError`. The real bug is that our downstream client (Goose) doesn't handle this format. Closing this and filing a fix on the Goose side instead.

PR #1761, opened 2026-03-20
fix: snapshot dicts before iteration in master _plan to prevent RuntimeError

mpuodziukas-labs:
## Problem

`Master._plan` iterates `self.state.instances.items()` and `self.state.last_seen.items()` across `await` points. Any concurrent coroutine that modifies these dicts (node arrival, departure, heartbeat) causes:

```
RuntimeError: dictionary changed size during iteration
```

This crashes the placement loop entirely, leaving the cluster in a degraded state until the next planning cycle.

## Fix

Wrap both loops with `list()` to snapshot before iteration:

```python
# kill broken instances — snapshot to avoid mutation during iteration
for instance_id, instance in list(self.state.instances.items()):

# time out dead nodes — snapshot to avoid mutation during iteration
for node_id, time in list(self.state.last_seen.items()):
```

## Tests

### DIFF
--- a/src/exo/master/main.py
+++ b/src/exo/master/main.py
@@ -374,9 +374,9 @@ class Master:
     # These plan loops are the cracks showing in our event sourcing architecture - more things could be commands
     async def _plan(self) -> None:
         while True:
-            # kill broken instances
+            # kill broken instances — snapshot to avoid mutation during iteration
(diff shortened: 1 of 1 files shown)


Evanev7:
nice catch

Evanev7:
on second thought, state is immutable! this error is not real.

PR #1763, opened 2026-03-20
Prevent failed instance retries

ciaranbor:
## Motivation

Currently, when a runner fails, the master retries the instance. Most of the time, this causes a loop over failure. Retries need backoff and a cap.

## Changes

- src/exo/worker/main.py: Before creating a runner, check an exponential backoff timer per instance. After EXO_MAX_INSTANCE_RETRIES failures, send DeleteInstance to permanently remove the instance. Record attempts on Shutdown; reset on InstanceDeleted.
- src/exo/utils/keyed_backoff.py: Add attempts() method to query retry count
- src/exo/shared/constants.py: Add EXO_MAX_INSTANCE_RETRIES = 3.

## Why It Works

The worker gates CreateRunner tasks behind a KeyedBackoff, adding exponential delay (2s base, 30s cap) between retries. After 3 failures the worker sends DeleteInstance, stopping retries entirely. The backoff resets when the instance is deleted, so a fresh placement starts clean.

### DIFF
--- a/src/exo/utils/keyed_backoff.py
+++ b/src/exo/utils/keyed_backoff.py
@@ -1,11 +1,9 @@
 import time
-from typing import Generic, TypeVar, final
-
-K = TypeVar("K")
+from typing import final
 
 
 @final
-class KeyedBackoff(Generic[K]):
+class KeyedBackoff[K]:
     """Tracks exponential backoff state per key."""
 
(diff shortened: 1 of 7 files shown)


AlexCheema:
I agree with limiting the number of retries with exponential backoff, but we should not remove retries altogether. Often I have run into situations where the instance fails the first time then succeeds after a retry (particularly common with RDMA).

PR #1768, opened 2026-03-22
Sync custom model cards across nodes

ciaranbor:
## Motivation

Custom model cards were only saved locally on the node that handled the API request. 

## Changes

- Added AddCustomModelCard and DeleteCustomModelCard commands
- Added CustomModelCardAdded and CustomModelCardDeleted events
- Added custom_model_cards field to cluster State
- Master handles new commands by emitting corresponding events
- Workers persist model cards to disk and update the in-memory cache on event receipt
- Separated fetch_from_hf (pure fetch) from disk persistence (now handled by event-sourcing layer)
- Exposed add_to_card_cache() helper for the worker to update the cache

## Why It Works

- Follows the existing event-sourcing pattern: API → Command → Master → Event → all Workers
- Every node applies the same events, so custom model cards are consistent across the cluster

### Manual Testing

Add/delete a custom model via the API on one node, verify it appears/disappears on all nodes

### Automated Testing

Existing tests cover apply() logic; new event types follow the same discriminated-union pattern

### DIFF
--- a/src/exo/api/main.py
+++ b/src/exo/api/main.py
@@ -129,8 +129,8 @@ from exo.shared.logging import InterceptLogger
 from exo.shared.models.model_cards import (
     ModelCard,
     ModelId,
+    get_card,
     get_model_cards,
-    is_custom_card,
 )
 from exo.shared.tracing import TraceEvent, compute_stats, export_trace, load_trace_file
 from exo.shared.types.chunks import (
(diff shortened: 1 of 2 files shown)


src/exo/shared/types/state.py
```diff
@@ -61,6 +62,9 @@ class State(CamelCaseModel):
     # Detected cycles where all nodes have Thunderbolt bridge enabled (>2 nodes)
     thunderbolt_bridge_cycles: Sequence[Sequence[NodeId]] = []
 
+    # Custom model cards synced across the cluster
+    custom_model_cards: Mapping[ModelId, ModelCard] = {}
```
Evanev7: is it a bad idea to outright replace the card cache with this state field?

Issue #1456, opened 2026-02-12
[BUG] DeepSeek V3.2 4bit Always thinking


ShireCiel:
## Describe the bug

Becuase of this commit.
Deepseek v3.2 model always enable think mode.
Deepseek v3.2 offical support thinking and no thinking mode.
Maybe can follow vllm's format "https://docs.vllm.ai/projects/recipes/en/latest/DeepSeek/DeepSeek-V3_2.html#tool-calling-example” to add a parameter to choose add <think> or not add it?

## To Reproduce

1. loading mlx/Deepseek V3.2 4bit
2. Do any curl request

## Expected behavior

Deepseek v3.2 offical support thinking and no thinking mode. but exo's output always have <think>.

## Actual behavior

When loading <https://huggingface.co/mlx-community/DeepSeek-V3.2-4bit> exo's output always have <think>.
But if i load model using "python -m mlx.server" it's always no <think>

## Environment

- macOS Version:  26.2
- EXO Version:  git commit (2fbdb27bb13a7915d75bc3599f638817d4042ce8)
- Hardware: 
  - Device 1: Mac Studio(M3 Ultra 512G)

Issue #1371, opened 2026-02-04
Replace model ID string matching with architecture-based checks in tokenizer loading


AlexCheema:
## Context

In `src/exo/worker/engines/mlx/utils_mlx.py`, the tokenizer loading code uses string matching on model IDs to apply model-specific patches (EOS tokens, custom tokenizer loading). This is fragile — it breaks when model IDs don't follow the expected naming convention.

Raised in https://github.com/exo-explore/exo/pull/1369#discussion_r2032835281.

## Current string matches

- [`"gemma-3" in model_id_lower`](https://github.com/exo-explore/exo/blob/2430a88b490c4cbd717c69e349c07ff6079ded6a/src/exo/worker/engines/mlx/utils_mlx.py#L387) — patches EOS token IDs to include `<end_of_turn>` (token 106)
- [`"kimi-k2" in model_id_lower`](https://github.com/exo-explore/exo/blob/2430a88b490c4cbd717c69e349c07ff6079ded6a/src/exo/worker/engines/mlx/utils_mlx.py#L339) — loads custom TikToken tokenizer and sets EOS token IDs
- [`"kimi-k2"` EOS check](https://github.com/exo-explore/exo/blob/2430a88b490c4cbd717c69e349c07ff6079ded6a/src/exo/worker/engines/mlx/utils_mlx.py#L309) / [`"glm-4.7-flash"`](https://github.com/exo-explore/exo/blob/2430a88b490c4cbd717c69e349c07ff6079ded6a/src/exo/worker/engines/mlx/utils_mlx.py#L311) / [`"glm"`](https://github.com/exo-explore/exo/blob/2430a88b490c4cbd717c69e349c07ff6079ded6a/src/exo/worker/engines/mlx/utils_mlx.py#L314) — sets EOS token IDs in [`get_eos_token_ids_for_model()`](https://github.com/exo-explore/exo/blob/2430a88b490c4cbd717c69e349c07ff6079ded6a/src/exo/worker/engines/mlx/utils_mlx.py#L295-L316)

## Possible approaches

- **`config.json` `model_type` field**: The canonical architecture identifier in mlx_lm. Available at tokenizer load time without needing the model object. Covers Gemma (`gemma3`, `gemma3n`) and GLM variants.
[...]

PR #1769, opened 2026-03-22
fix: DeepSeek V3.2 warmup crash and tool calling + add catalog cards

vskiwi:
## Summary

DeepSeek V3.2 (`DeepseekV32ForCausalLM`) is already supported by exo's inference engine (architecture whitelisted in `model_cards.py`, DSML encoding added in #1548), but **doesn't work out of the box** due to two bugs:

### Bug 1: `warmup_inference` passes empty model ID

`warmup_inference()` in `generate.py` accepts `model_id: ModelId` as a parameter but creates `TextGenerationTaskParams(model=ModelId(""), ...)` instead of using it. Since `_needs_dsml_encoding()` checks `"deepseek-v3.2" in task_params.model.lower()`, the empty string never matches → falls back to `tokenizer.apply_chat_template()` → **ValueError** because V3.2 has no Jinja chat template.

**Fix:** `model=ModelId("")` → `model=model_id` (one line).

### Bug 2: `_needs_dsml_encoding` limited to tool calling

`_needs_dsml_encoding()` returns `True` only when `task_params.tools` is present or tool messages exist in `chat_template_messages`. For warmup and regular chat requests without tools → `return False` → Jinja fallback → **ValueError**.

Unlike V3.1 (which has a `.jinja` chat template file that transformers picks up automatically), V3.2 **has no Jinja template at all** — it uses Python-based DSML encoding for all message types.

**Fix:** For V3.2, always return `True` — DSML encoding handles all message types.

### Catalog cards

Added inference model cards for:
- `mlx-community/DeepSeek-V3.2-8bit`
- `mlx-community/DeepSeek-V3.2-4bit`

Parameters taken from model `config.json` on HuggingFace, storage sizes from HF API. Capabilities include `thinking_toggle` (related: #1456).

## Notes

- The model ID string matching approach (`"deepseek-v3.2" in model.lower()`) is acknowledged tech debt — see #1371 for the planned architecture-based approach.

## Test plan

- [ ] Start exo with DeepSeek V3.2 model → warmup should complete without crash
- [ ] Send a regular chat message (no tools) → should get a response
- [ ] Send a chat message with tools → should work as before
- [ ] V3.2 cards should appear in the dashboard model catalog

### DIFF
--- a/rust/exo_pyo3_bindings/exo_pyo3_bindings.pyi
+++ b/rust/exo_pyo3_bindings/exo_pyo3_bindings.pyi
@@ -42,7 +42,7 @@ class MessageTooLargeError(builtins.Exception):
 
 @typing.final
 class NetworkingHandle:
-    def __new__(cls, identity: Keypair, bootstrap_peers: list[builtins.str], listen_port: builtins.int) -> NetworkingHandle: ...
+    def __new__(cls, identity: Keypair, bootstrap_peers: typing.Sequence[builtins.str], listen_port: builtins.int) -> NetworkingHandle: ...
     async def gossipsub_subscribe(self, topic: builtins.str) -> builtins.bool:
         r"""
         Subscribe to a `GossipSub` topic.
--- a/src/exo/worker/engines/mlx/auto_parallel.py
+++ b/src/exo/worker/engines/mlx/auto_parallel.py
@@ -57,8 +57,8 @@ from mlx_lm.models.step3p5 import Model as Step35Model
 from mlx_lm.models.step3p5 import Step3p5MLP as Step35MLP
 from mlx_lm.models.step3p5 import Step3p5Model as Step35InnerModel
 
-from exo.shared.logging import logger
 from exo.shared.types.worker.shards import PipelineShardMetadata
+from exo.worker.runner.bootstrap import logger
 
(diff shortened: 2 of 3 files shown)


rltakashige:
Hi there! Thanks for the contributions.

I've updated this branch to try to fix a bunch of other failure cases I was seeing with DSV32.

vskiwi:
@rltakashige Thanks for picking this up and expanding it — the parser-level `finish_reason` handling is much more comprehensive now (covers `in_tool_call` state, single-token DSML blocks, etc.).

One thing #1775 addresses that isn't in this PR: the `BatchGenerator.step()` drain loop. Currently only the `SequentialGenerator` path drains remaining parser outputs when a generation finishes. In `BatchGenerator`, if `StopIteration` fires while the output generator still has buffered items (e.g. flushed pending buffer + finish token), they'll be lost. The fix is a small `while` drain in two places — happy to rebase #1775 to only include that part once this lands, or it could be folded in here.

rltakashige:
Hey vskiwi, I think we have addressed this issue here already! We drain the parsed output generator first before checking for the finish reason on the original response.

vskiwi:
Thanks for the thorough expansion and the drain fix — looking forward to contributing more!

PR #1770, opened 2026-03-23
feat(cluster): KV cache remote tier + crash-loop fix + security gate

mpuodziukas-labs:
## Summary
- **Fix worldSize=1 cascade failure**: `placement_utils.py` now falls back to `node_network` interfaces when topology graph has no `SocketConnection` edges yet (race between peer-reconnect and the 10s reachability poll). Prevents `ValueError: MLX ring backend requires connectivity between neighbouring nodes` → crash loop → Metal OOM → kernel panic chain.
- **Fix `[Errno 48]` runner port conflict**: `worker/main.py` shuts down any existing runner with the same ID before creating a new one — stops `Address already in use` when `CreateRunner` arrives while the previous process still holds the port.
- **P3.4 Sentinel Replay-Attack Dedup** security gate check: flags `apply`/`handle_event`/`process_event` functions missing dedup early-return guards against replayed events.

## Test plan
- [ ] All 352 existing tests pass (`uv run pytest`)
- [ ] 0 type errors (`uv run basedpyright`)
- [ ] Ruff clean (`uv run ruff check`)
- [ ] worldSize=2 maintained after Mini reconnect (no more `ValueError` on placement)
- [ ] No `[Errno 48]` on runner restart after crash

### DIFF
--- a/src/exo/master/placement_utils.py
+++ b/src/exo/master/placement_utils.py
@@ -346,10 +346,39 @@ def _find_ip_prioritised(
     """Find an IP address between nodes with prioritization.
 
     Priority: ethernet > wifi > unknown > thunderbolt
+    Falls back to node_network interfaces when topology graph has no edges yet
+    (e.g. before the reachability poll has completed after a peer reconnect).
     """
     ips = list(_find_connection_ip(node_id, other_node_id, cycle_digraph))
     if not ips:
-        return None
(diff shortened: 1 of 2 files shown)


Evanev7:
heya! please open an issue if you want some of these features in exo, we can discuss some of the design

PR #1771, opened 2026-03-23
fix(placement): fall back to node_network when topology edges absent

mpuodziukas-labs:
Two fixes for ring placement stability on reconnect.

**`placement_utils.py`**: `_find_ip_prioritised` returned `None` when topology had no `SocketConnection` edges — this happened every time a peer reconnected because the reachability poll runs every 10s. The `None` propagated to `build_ring_assignments` as a missing IP, raising `ValueError` and leaving `worldSize=1`. Fix: when `ips` is empty, fall back to `node_network.interfaces` ranked by the same priority dict (thunderbolt → ethernet → wifi).

**`worker/main.py`**: `_create_supervisor` silently overwrote `self.runners[runner_id]` without shutting down the old process. The old runner kept its port open, causing `[Errno 48] Address already in use` on the new runner's bind. Fix: pop and shut down before replacing.

### DIFF
--- a/src/exo/master/placement_utils.py
+++ b/src/exo/master/placement_utils.py
@@ -364,7 +364,6 @@ def _find_connection_ip(
     node_j: NodeId,
     cycle_digraph: Topology,
 ) -> Generator[str, None, None]:
-    """Find all IP addresses that connect node i to node j."""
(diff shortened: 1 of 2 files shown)


Evanev7:
this is obviously not just a small fix, please open an issue or re-submit with more information!

PR #1787, opened 2026-03-24
fix: cancel generation when SSE stream closes

dale-lakes:
## Motivation

When the SSE stream closes (e.g. after sending a ToolCallChunk), `_token_chunk_stream` only sends `TaskFinished`, which deletes the task from state. The runner never receives a cancel signal, so the MLX generation continues until `max_tokens`.

## Changes

1. **`api/main.py`**: Send `TaskCancelled` in the `finally` block of `_token_chunk_stream` before `TaskFinished`. This ensures the cancel pipeline is triggered on every stream exit.

2. **`worker/main.py`**: When `TaskDeleted` arrives, cancel the runner directly in the event applier before removing the task from state. This bypasses the planner's 100ms polling loop, which can miss the `Cancelled` status if `TaskDeleted` removes the task first.

## Why It Works

The cancel pipeline already exists (`TaskCancelled` → `TaskStatusUpdated(Cancelled)` → `CancelTask` → runner cancel). The issue was that it was never triggered on normal stream close. The worker-level fix adds defense-in-depth by reacting to `TaskDeleted` synchronously, avoiding the race between the planner's poll interval and task deletion.

### Manual Testing
Hardware: 2x Mac Studio M2 Ultra 512GB, Thunderbolt 5

- Ran Goose sessions with Kimi K2.5 tool calling through exo
- Verified generations stop promptly after each tool call turn
- Confirmed subsequent turns work correctly after cancellation

### Automated Testing
Existing test suite passes (249 passed, 1 skipped)


Evanev7:
i'm concerned this is fixing the wrong issue - if kimi is making multiple tool calls, we're probably closing the stream too early and should support that. i haven't seen this behaviour so i appreciate you raising it!

rltakashige:
Hi as well - I think the issue listed here is genuine, and I think I've fixed in a different (and slightly nicer) way in #1769. Please check if it works for you, as I don't think we'll have the time to test your particular setup for a while.

rltakashige:
Going to close this as stale. Hopefully #1769 is sufficient - please reopen this if you do not believe this to be the case!

PR #1790, opened 2026-03-25
Rename Nemotron to NVIDIA in model picker with logo

AlexCheema:
## Motivation

The Nemotron model family in the model picker sidebar was displaying as "Nemotron" with a generic checkmark icon. Since these models are NVIDIA's Nemotron models, the category should be branded as "NVIDIA" with the official NVIDIA logo, consistent with how other families are branded (e.g., "llama" → "Meta", "gpt-oss" → "OpenAI").

## Changes

- **FamilySidebar.svelte**: Added `nemotron: "NVIDIA"` to the `familyNames` mapping so the sidebar displays "NVIDIA" instead of "Nemotron"
- **FamilyLogos.svelte**: Added the NVIDIA "eye" logo as an inline SVG for the `nemotron` family, matching the `viewBox="0 0 24 24"` / `fill="currentColor"` pattern used by all other brand logos
- **ModelPickerModal.svelte**: Added `"nemotron"` to the `familyOrder` array so NVIDIA appears in a consistent position in the sidebar

## Why It Works

The model picker derives categories from the `family` field in TOML model cards. Nemotron models already have `family = "nemotron"`, but the three UI components (display name, logo, sort order) lacked explicit entries for it, causing fallback behavior (auto-capitalized name, checkmark icon, alphabetical sorting). Adding explicit entries for all three aligns NVIDIA with the existing brand pattern.

### Manual Testing

- Built dashboard successfully (`npm run build`)
- Verified the NVIDIA logo renders in the sidebar alongside existing brand logos

### Automated Testing
- No test changes needed — this is a purely cosmetic dashboard change


AlexCheema:
Here are the logos now:

Screenshot 2026-03-24 at 5 28 15 PM

PR #1802, opened 2026-03-26
Add multimodality!

rltakashige:
## Motivation

Images!

TODO (in a future PR): Add audio and video support.

### Manual Testing
image

image
image

(And batching also works)

### DIFF
--- a/src/exo/worker/main.py
+++ b/src/exo/worker/main.py
@@ -155,7 +155,7 @@ class Worker:
                 self.state.instances,
                 self.state.runners,
                 self.state.tasks,
-                self.input_chunk_counts,
+                self.input_chunk_buffer,
             )
             if task is None:
                 continue
--- a/src/exo/worker/plan.py
+++ b/src/exo/worker/plan.py
@@ -2,6 +2,7 @@
 
 from collections.abc import Mapping, Sequence
 
+from exo.shared.types.chunks import InputImageChunk
 from exo.shared.types.common import CommandId, NodeId
 from exo.shared.types.tasks import (
     CancelTask,
@@ -49,7 +50,7 @@ def plan(
     instances: Mapping[InstanceId, Instance],
     all_runners: Mapping[RunnerId, RunnerStatus],  # all global
     tasks: Mapping[TaskId, Task],
-    input_chunk_counts: Mapping[CommandId, int] | None = None,
+    input_chunk_buffer: Mapping[CommandId, Mapping[int, InputImageChunk]] | None = None,
 ) -> Task | None:
     # Python short circuiting OR logic should evaluate these sequentially.
     return (
@@ -60,7 +61,7 @@ def plan(
         or _init_distributed_backend(runners, all_runners)
         or _load_model(runners, all_runners, global_download_status)
         or _ready_to_warmup(runners, all_runners)
-        or _pending_tasks(runners, tasks, all_runners, input_chunk_counts or {})
+        or _pending_tasks(runners, tasks, all_runners, input_chunk_buffer or {})
     )
 
 
@@ -271,7 +272,7 @@ def _pending_tasks(
     runners: Mapping[RunnerId, RunnerSupervisor],
     tasks: Mapping[TaskId, Task],
     all_runners: Mapping[RunnerId, RunnerStatus],
-    input_chunk_counts: Mapping[CommandId, int],
+    input_chunk_buffer: Mapping[CommandId, Mapping[int, InputImageChunk]] | None,
 ) -> Task | None:
     for task in tasks.values():
         # for now, just forward chat completions
@@ -286,8 +287,9 @@ def _pending_tasks(
         if isinstance(task, (ImageEdits, TextGeneration)):
             expected_image_chunks = task.task_params.total_input_chunks
         if expected_image_chunks > 0:
+            assert input_chunk_buffer is not None
             cmd_id = task.command_id
-            received = input_chunk_counts.get(cmd_id, 0)
+            received = len(input_chunk_buffer.get(cmd_id, {}))
             if received < expected_image_chunks:
                 continue  # Wait for all chunks to arrive
 


src/exo/api/adapters/chat_completions.py
```diff
@@ -33,25 +37,65 @@
 )
 
 
-def chat_request_to_text_generation(
+def extract_base64_from_data_url(data_url: str) -> str:
+    match = re.match(r"data:[^;]+;base64,(.+)", data_url)
```
Evanev7, line 41: nit: don't use regex, this is simple enough for python string matching

src/exo/shared/models/model_cards.py
```diff
@@ -38,13 +40,45 @@
...
+
+
 async def _load_cards_from_dir(directory: Path, *, is_custom: bool) -> None:
     """Load all TOML model cards from a directory into the cache."""
     async for toml_file in directory.rglob("*.toml"):
         try:
             card = await ModelCard.load_from_path(toml_file)
             if is_custom:
                 card = card.model_copy(update={"is_custom": True})
+            if card.vision is None:
```
Evanev7, line 68: this late-loading logic is a bit weird - why do we not expect this information to be present?

src/exo/shared/models/model_cards.py
```diff
@@ -89,6 +123,14 @@ class ComponentInfo(CamelCaseModel):
     safetensors_index_filename: str | None = None
 
 
+class VisionCardConfig(CamelCaseModel):
+    image_token_id: int
+    model_type: str
+    weights_repo: str | None = None
```
Evanev7: am i correct in assuming weights_repo is None when the weights are local (i.e. the same as the rest of the model?)

src/exo/shared/models/model_cards.py
```diff
@@ -196,6 +255,9 @@ class ConfigData(BaseModel):
             "decoder_layers",
         )
     )
+    vision_config: dict[str, Any] | None = None
+    image_token_id: int | None = None
+    model_type: str | None = None
```
Evanev7: use the model validator to grab the data you care about rather than accepting dict[str, Any] unless you actually need dict[str, Any]

Evanev7: (i.e. the visioncardconfig should come from here)

src/exo/worker/engines/mlx/generator/generate.py
```diff
@@ -55,13 +57,51 @@
...
+        if not attr.startswith("_") and not hasattr(_inject, attr):
+            with contextlib.suppress(AttributeError, TypeError):
+                setattr(_inject, attr, getattr(original_embed, attr))  # type: ignore
+
+    inner.embed_tokens = _inject
+
+    def _cleanup() -> None:
+        inner.embed_tokens = original_embed
+
+    return _cleanup
```
Evanev7: nasty. tempted to suggest making this a context manager?

src/exo/worker/engines/mlx/vision.py
```diff
@@ -0,0 +1,605 @@
...
+
+
+class VisionEncoder:
+    def __init__(self, config: VisionCardConfig, model_id: ModelId):
+        self._config = config
+        self._main_model_path = build_model_path(model_id)
+        self._model_path = (
+            build_model_path(ModelId(config.weights_repo))
+            if config.weights_repo
+            else self._main_model_path
```
Evanev7: wondering if we should make weights_repo always valid (and just equal to main model path otherwise)

src/exo/worker/engines/mlx/vision.py
```diff
@@ -0,0 +1,605 @@
...
+    valid = set(inspect.signature(cls.__init__).parameters.keys()) - {"self"}
+    return {k: v for k, v in d.items() if k in valid}  # type: ignore
+
+
+_video_processor_patched = False
+
+
+def _patch_video_processor() -> None:
+    """Patch so we don't crash horribly when torch vision isn't installed"""
+    # TODO: Consider whether we want torch vision.
```
Evanev7:
```suggestion
    # TODO: Update if we add torch vision.
```
?

src/exo/worker/engines/mlx/vision.py
```diff
@@ -0,0 +1,605 @@
...
+        if i < len(images):
+            img = decode_base64_image(images[i])
+            region.content_hash = hashlib.sha256(img.tobytes()).hexdigest()
+        else:
+            logger.warning(f"Media region {i} has no corresponding image")
+
+    return regions
+
+
+class VisionPipeline:
```
Evanev7: can we not call it a pipeline hahaha

src/exo/worker/engines/mlx/vision.py
```diff
@@ -0,0 +1,605 @@
...
+
+        return VisionResult(
+            prompt=prompt,
+            prompt_tokens=prompt_tokens,
+            embeddings=embeddings,
+            media_regions=media_regions,
+        )
+
+
+_vision_pipeline: VisionPipeline | None = None
```
Evanev7: ok can we keep this in the rest of the loading instead of a secret hidden global please ty

src/exo/worker/plan.py
```diff
@@ -49,7 +49,7 @@ def plan(
     instances: Mapping[InstanceId, Instance],
     all_runners: Mapping[RunnerId, RunnerStatus],  # all global
     tasks: Mapping[TaskId, Task],
-    input_chunk_buffer: Mapping[CommandId, dict[int, str]] | None = None,
+    input_chunk_buffer: Mapping[CommandId, Mapping[int, object]] | None = None,
```
Evanev7: we know the type of the input chunk buffer

src/exo/worker/main.py
```diff
@@ -279,12 +281,70 @@ async def plan_step(self):
...
                         del self.input_chunk_counts[cmd_id]
                     await self._start_runner_task(modified_task)
+
+                case TextGeneration() if task.task_params.image_hashes:
+                    cached_images = [
+                        self.image_cache[h]
+                        for h in task.task_params.image_hashes
+                        if h in self.image_cache
+                    ]
+                    if len(cached_images) == len(task.task_params.image_hashes):
```
Evanev7: should be an assertion right? we check this in the plan step

src/exo/worker/main.py
```diff
@@ -279,12 +281,70 @@ async def plan_step(self):
...
+                        task_id=task.task_id,
+                        command_id=task.command_id,
+                        instance_id=task.instance_id,
+                        task_status=task.task_status,
+                        task_params=task.task_params.model_copy(
+                            update={
+                                "images": assembled_images,
+                            }
+                        ),
+                    )
```
Evanev7:
```suggestion
                    modified_task = task.model_copy(
                        task_params=task.task_params.model_copy(
                            update={
                                "images": assembled_images,
                            }
                        ),
                    )
```
or are you intentionally nulling other fields

pyproject.toml
```diff
@@ -66,6 +67,9 @@ mlx-lm = { git = "https://github.com/rltakashige/mlx-lm", branch = "leo/fix-deep
 # mlx = { path = "/Users/Shared/mlx", editable=true }
 # mlx-lm = { path = "/Users/Shared/mlx-lm", editable=true }
 
+[tool.uv.extra-build-dependencies]
+"miniaudio" = ["setuptools"]
```
Evanev7:
```suggestion
"miniaudio" = ["setuptools", "cffi", "pycparser"]
```
then you can drop the parts.nix changes

src/exo/shared/models/model_cards.py
rltakashige: I've manually added vision configs to relevant models in a later PR, but for custom models, we need to detect from the model config.

src/exo/api/adapters/chat_completions.py
rltakashige: I actually prefer this regex over a find() here quite a bit..

src/exo/worker/main.py
rltakashige: No, all the fields are copied. This is better I agree

src/exo/worker/engines/mlx/vision.py
rltakashige: Yeah should've changed that mb

src/exo/worker/engines/mlx/vision.py
rltakashige: What about VisionProcessor

src/exo/worker/engines/mlx/generator/generate.py
rltakashige: Just learnt the syntax for @ contextlib and it's a bit magic. Like it though

PR #1805, opened 2026-03-27
Fix custom model add requiring two attempts + enlarge sidebar buttons

AlexCheema:
## Motivation

Adding a custom model from the Hub tab shows "Added" toast but the model doesn't appear in the All tab. You have to add it a second time for it to work. Also, the "All" button in the model picker sidebar is too small to read comfortably.

## Changes

**Race condition fix (`src/exo/api/main.py`):**
- Call `add_to_card_cache(card)` directly in `add_custom_model()` after sending the `ForwarderCommand`, before the API response returns

**Sidebar sizing (`dashboard/src/lib/components/FamilySidebar.svelte`):**
- Increased sidebar min-width from 72/64px to 80/72px
- Increased "All" icon from `w-5 h-5` to `w-6 h-6`
- Increased all sidebar labels from 9px to 11px

## Why It Works

`POST /models/add` sends a `ForwarderCommand(AddCustomModelCard)` and returns immediately. The frontend then calls `GET /models` which reads from `_card_cache`. But the cache was only updated by the worker event handler after the event round-trips through the master — a race the frontend almost always loses. By updating the cache directly in the API handler, `GET /models` immediately reflects the new model. The worker's later `add_to_card_cache` call is idempotent (dict key assignment).

### Manual Testing

- Open model picker → Hub tab → add a custom model → verify it appears in All tab on the first attempt
- Verify sidebar "All" button and other labels are visually larger and readable

### Automated Testing
- `uv run basedpyright` passes with 0 errors
- `uv run ruff check` passes

### DIFF
--- a/dashboard/src/lib/components/FamilySidebar.svelte
+++ b/dashboard/src/lib/components/FamilySidebar.svelte
@@ -48,25 +48,14 @@
   <button
     type="button"
     onclick={() => onSelect(null)}
-    class="group flex flex-col items-center justify-center p-2 sm:p-2 rounded transition-all duration-200 cursor-pointer min-h-[44px] sm:min-h-0 {selectedFamily ===
+    class="group flex items-center justify-center px-3 py-2.5 rounded transition-all duration-200 cursor-pointer min-h-[44px] sm:min-h-0 {selectedFamily ===
     null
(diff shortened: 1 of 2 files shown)


rltakashige (approved):
Don't think the change can cause any bad races - lgtm thanks!

PR #1810, opened 2026-03-27
Integrations helpers

rltakashige:


Evanev7 (approved):
this is really nice actually, for a +1300 i was expecting something pretty spooky but i'm more than happy with this

PR #1811, opened 2026-03-27
Tighten EXO bench concurrency numbers and explain methodology

rltakashige:
## Motivation

The timings in the batch generator are a little optimistic; a minor change is needed to make them more correct.

## Changes

Include the time spent in the API in the generation tps and make sure to send all requests simultaneously


rltakashige:
Actually, converting this to a draft because I want to remember to add a more thorough explanation of methodology in EXO bench, and I want to add a version compatability check, so that any major changes to methodology will get flagged (e.g. running exo-bench from v1.0.68 with the new version)

Evanev7:
approval still holds. rust change is legit, not sure why it wasn't caught by ci earlier.

PR #1813, opened 2026-03-28
fix: distinguish TB4 from TB5 for Thunderbolt connections

### LINKED ISSUE #1636: [BUG] EXO Suggest enabling RDMA on Thunderbolt 4
## Describe the bug

Exo claims there is a thunderbolt 5 connection on Mac Studios M1 and M2

## To Reproduce

1. Start EXO on Mac Studio M1/M2
2.
3.

## Expected behavior

Thunderbolt 4 hardware should be detected and no RDMA activation suggested

## Actual behavior

There is a message about Thunderbolt 5 present on M1/M2

## Environment

- macOS Version: 26.3
- EXO Version: 1.0.68
- Hardware:
  - Device 1: Mac Studio M1, 64GB RAM
  - Device 2: Mac Studio M2, 64GB RAM
  - Additional devices:
- Interconnection:
  - Apple Original Thunderbolt 5 cable between Device 1 and 2
  - 1GbE Ethernet between all devices

dogukanveziroglu:
## Summary
- TB4 devices (≤40 Gb/s) don't support RDMA, but macOS Tahoe creates `rdma_enX` devices for all Thunderbolt ports. This caused all Thunderbolt connections to be registered as `RDMAConnection`, breaking placement for TB4 clusters.
- Parse `linkSpeed` to distinguish TB4 from TB5: TB5 (>40 Gb/s) gets `RDMAConnection`, TB4 gets `SocketConnection` via Thunderbolt Bridge IP
- Dashboard "RDMA NOT ENABLED" warning now only shows for actual TB5 hardware

## Test plan
- [x] Tested on 2x Mac Mini M4 (TB4) cluster — placement now works correctly with MlxRing
- [x] TB5 behavior unchanged (RDMAConnection created as before)
- [ ] Verify on TB5 hardware

Fixes #1636


AlexCheema:
What do you mean when you say "breaking placement for TB4 clusters".
Ring should not use RDMA, so I don't see why it would break anything?

dogukanveziroglu:
> What do you mean when you say "breaking placement for TB4 clusters". Ring should not use RDMA, so I don't see why it would break anything?

You're right Ring doesn't use RDMA, so the apply.py change was unnecessary. My bad, the RDMA warning on the dashboard confused me into thinking the connection type was the issue, but it was actually ping discovery failing in my environment (macOS LNP blocking check_reachable in SSH sessions).

I'll remove the apply.py change and keep only the dashboard warning fix.

PR #1677, opened 2026-03-07
feat(dashboard): add mobile drawer support for sidebars

ArvidSU:
## Summary

- Adds mobile-friendly drawer UI for the chat sidebar and model family sidebar
- Introduces responsive header navigation with mobile-specific toggle buttons
- Uses slide-in drawer overlay pattern on small screens instead of collapsible sidebars
- Stores mobile drawer state in app store for consistent state management

## Testing

Tested on desktop (responsive mode) and mobile viewport widths. Sidebars now slide in as drawers on small screens with proper z-index layering and close-on-outside-click behavior.

PR #1814, opened 2026-03-29
feat: add Firefox AI sidebar (?q=) support to dashboard

ArvidSU:
This PR builds on https://github.com/exo-explore/exo/pull/1677 to enable custom prompts sent from Firefox `browser.ml.chat` to EXO dashboard using URL parameters in sidebar for summary and other browser interactions. See "Summarize page" example below.

## Summary
- Parse `?q=<encoded prompt>` URL parameter on page load and auto-submit it as a chat message
- Clean up the URL with `history.replaceState` to prevent re-submission on refresh
- Defer auto-send until both cluster state and model list are loaded so model auto-selection works correctly

## Context
Firefox's built-in AI sidebar (`about:config: browser.ml.chat.enabled`) integrates with chat providers by appending the user's prompt as `?q=<URL-encoded prompt>`. Previously the exo dashboard ignored this parameter. Users can now configure `http://localhost:52415` as a Firefox AI chatbot provider.

See: https://support.mozilla.org/en-US/kb/ai-chatbot

## Technical notes
- Frontend-only change in `dashboard/src/routes/+page.svelte`
- Uses a Svelte `$effect` that reacts to `pendingFirefoxQuery`, `data` (cluster state), and `models.length` — fires exactly once when all three are ready
- If no model is selected, `handleAutoSend` auto-picks the best available model; if no model fits memory, a toast is shown
- If a model is selected but not running, the message is queued until the model loads

## Testing
```
http://localhost:52415/?q=Hello+world
http://localhost:52415/?q=Summarize+this+page%3A+%5Bpage+title%5D+%5Bpage+url%5D
```

image

### DIFF
--- a/dashboard/src/routes/+page.svelte
+++ b/dashboard/src/routes/+page.svelte
@@ -265,6 +265,7 @@
 
   let mounted = $state(false);
   let localNodeId = $state<string | null>(null);
+  let pendingFirefoxQuery = $state<string | null>(null); // ?q= param deferred until state loads
 
   // ── Onboarding wizard state ──
   const ONBOARDING_COMPLETE_KEY = "exo-onboarding-complete";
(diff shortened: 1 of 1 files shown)


rltakashige:
Hey - this is really cool; it looks good to me, but I think it would be better if we put this at a /firefox endpoint so that we don't reserve the top-level ?q= for only Firefox.

ArvidSU:
@rltakashige Hmmm, I see the confusion with my variable naming and comments. However, there is not really anything functional that makes this Firefox specific, the `?q=` param is more a de facto convention used by google and chatgpt etc. Since the url is configurable in firefox I can move it to a firefox specific path but my suggestion is to generalize the naming of variables and comments to align with convention. What do you think?

Issue #953, opened 2025-12-21
[HARD] Support arbitrary tensor parallel splits


rltakashige:
**Context:**
Tensor parallelism allows models to be split among multiple devices by distributing weights among devices.
auto_parallel strategies for tensor parallelism use MLX LM's `shard_linear` and `shard_inplace` functions to split tensors evenly. These functions assume that the sharding dimensions are exactly divisible by the group size.

**Issue:**
There are several limitations to this approach:

1. Certain models are unable to be supported by tensor parallelism, particularly quantized models (see GLM Air 4.5 config.json with gs 32 quantization).
2. Most models cannot be parallelised across 3 nodes, since their intermediate size is not divisible by 3.
3. This issue is a blocker for smarter tensor parallelism on heterogeneous devices.

A potential start is smart padding.

DevKiDCosmo:
I would also add a small benchmark test to see how to split the model among the devices instead of splitting it equally like I understand. I would like to work on this problem but with someone else to get to know a little bit more about the code base.

The dimensions don't need to be exactly divisible by the group size if we implement a system where percentages or fraction of the model are distributed. To this day, all I know is that the models are always evenly divided, right?

Evanev7:
yup! we have serious issues running TP on 3 devices because most model hidden dimensions aren't divisible by it.

We do have a fixed split we need to implement, but our TP sharding code is a lot more involved than our Pipeline code so it's not just an issue of rounding some percentages better.

DevKiDCosmo:
Then I don't see a problem. Probably I misunderstood something.

By TP a layer or multiple layer is splitted across devices. But how are they splitted across those devices? I think I might get it? Instead of splitting the model vertically (by layer) it split horizontally (by neurons). But not every layer has the exact same amount of neurons in a layer. Is this the problem? And is every layer split horizontally?

DevKiDCosmo:
Is there also something hybrid like powerful devices are taking layers and less powerful devices doing it by TP?

rltakashige:
Correct me if I misunderstood your reply - a model is composed of multiple transformer blocks (as well as some other modules we don't care too much about). 

In pipeline parallelism, some of the transformer blocks are assigned to one device, some others to a second device, etc. 
In tensor parallelism, we shard the linear modules *inside* the transformer blocks. All of the linear modules are split across their intermediate dimension (although which dimension this is depends on the linear module).

[...]

PR #1821, opened 2026-03-31
feat: asymmetric tensor parallelism for heterogeneous clusters

Gajesh2007:
## Summary

Adds asymmetric tensor parallelism for clusters where nodes have different amounts of RAM. When standard 50/50 tensor parallelism fails because the smaller node can't hold half the weights, this splits each weight tensor proportionally (e.g. 75/25) so both GPUs compute every layer simultaneously.

Addresses #953 (arbitrary tensor parallel splits).

## Problem

When connecting a 128GB Mac to a 48GB Mac, a 130GB model can't use tensor parallelism today — 65GB per side doesn't fit on the 48GB node. The only option is pipeline parallelism, where one GPU sits idle while the other computes.

## Solution

Asymmetric TP splits weights proportionally to available memory. For 128GB + 48GB:
- Node A gets 75% of each weight tensor (~98GB)
- Node B gets 25% (~33GB)  
- Both GPUs compute every layer simultaneously
- `all_sum` still works correctly: `(x_a @ W_a^T) + (x_b @ W_b^T) = x @ W^T` regardless of split ratio

Includes a ratio solver that finds valid split points where all dimensions (attention heads, MoE intermediate size, etc.) divide cleanly and satisfy quantization alignment constraints.

Auto-upgrades `Tensor` → `AsymmetricTensor` in placement when it detects the equal split won't fit on the smallest node but the combined memory is sufficient.

## Testing

**Hardware tested:**
- Machine A: MacBook Pro M4 Max, 128GB unified memory, macOS 26.4
- Machine B: MacBook Pro M4 Pro, 48GB unified memory, macOS 26.3
- Connection: Thunderbolt 5 cable, RDMA via JACCL

**Model tested:**
- Qwen3.5-122B-A10B 8-bit (130.5GB) — cannot fit on either machine alone

**Results:**

| Approach | tok/s | Both GPUs active? |
|---|---|---|
| Pipeline parallelism (36/12 layer split) | ~5 | No — one idle at a time |
| **Asymmetric TP (75/25 weight split)** | **41.8** | **Yes — every token** |

Asymmetric TP is ~8x faster than pipeline on the same hardware.

**Automated tests:** 7 new unit tests for the ratio solver covering Qwen3.5, Llama-70B, and Nemotron dimensions, plus edge cases (impossible dimensions, >2 nodes).

**CI:** 0 basedpyright errors, 0 ruff errors, 347/347 tests pass.

## Architecture support

Currently supports **Qwen3.5** (GatedDeltaNet + Attention + SparseMoeBlock). Raises a clear error for unsupported architectures. Follow-up PRs can add NemotronH, DeepSeek V3, Llama, etc. — each needs its own sharding strategy tested on real hardware.

## Changes

- `src/exo/shared/types/worker/shards.py` — new `Sharding.AsymmetricTensor` enum + `AsymmetricTensorShardMetadata` with per-node `ratio` field
- `src/exo/master/placement.py` — auto-upgrade logic when equal TP won't fit
- `src/exo/master/placement_utils.py` — `get_shard_assignments_for_asymmetric_tensor_parallel()`
- `src/exo/worker/engines/mlx/utils_mlx.py` — handle `AsymmetricTensorShardMetadata` in `shard_and_load`
- **NEW** `src/exo/worker/engines/mlx/asymmetric_parallel.py` — ratio solver + per-layer sharding
- **NEW** `src/exo/worker/tests/unittests/test_mlx/test_asymmetric_parallel.py` — 7 unit tests

## Prior art

The concept of asymmetric TP exists in research (HexGen ICML 2024, HAPT, AutoHet) and has implementations for CUDA (HexGen, veScale). This is the first implementation for MLX on Apple Silicon.

### DIFF
--- a/src/exo/master/placement.py
+++ b/src/exo/master/placement.py
@@ -129,6 +129,19 @@ def place_instance(
     if len(cycles_with_sufficient_memory) == 0:
         raise ValueError("No cycles found with sufficient memory")
 
+    # Asymmetric TP currently only supports Qwen3.5 at the worker level.
+    asymmetric_tp_families = {"qwen3_5"}
+
+    if (
+        command.sharding == Sharding.AsymmetricTensor
+        and command.model_card.family not in asymmetric_tp_families
+    ):
+        raise ValueError(
+            f"Asymmetric tensor parallelism is not yet supported for "
+            f"family '{command.model_card.family}'. "
+            f"Supported: {asymmetric_tp_families}"
+        )
+
     if command.sharding in (Sharding.Tensor, Sharding.AsymmetricTensor):
         if not command.model_card.supports_tensor:
             raise ValueError(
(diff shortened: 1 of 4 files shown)


Gajesh2007:
Pushed a fix addressing four integration bugs caught in review:

1. **`ModelCard.num_attention_heads` doesn't exist** — was crashing at placement time. Now derives from `hidden_size // 128` as a conservative estimate; full validation happens at model load time when the config is available.
2. **Ratio semantics mismatch** — rank 1 was getting an inverted split point, causing misaligned weight slices. Now all shards store rank-0's ratio so both workers agree on the same split.
3. **No model family gate** — `AsymmetricTensor` was allowed for any `supports_tensor` model, but the worker only handles Qwen3.5. Now gated to `qwen3_5` family and raises a clear error for others.
4. **Cycle ordering** — if the smaller-memory node was first in the cycle, it would get the larger shard. Now sorts nodes by available memory so rank 0 is always the bigger machine.

These bugs existed because our hardware test bypassed the exo placement/loading pipeline (hardcoded ratios, called the sharding module directly). The sharding math itself is correct — verified at 41.8 tok/s on Qwen3.5-122B across 128GB + 48GB Macs.

---

**Note on model support:** This has only been tested on **Qwen3.5-122B-A10B 8-bit**. The worker currently raises a clear error for unsupported architectures. If anyone wants to help expand coverage, the main work is adding per-architecture sharding functions in `asymmetric_parallel.py` — similar to how `auto_parallel.py` has separate `ShardingStrategy` classes. NemotronH (Mamba2 + Attention + MoE) and Llama-style models are the natural next targets.

PR #1828, opened 2026-04-01
Cancel SSE keep-alive when instance is deleted

ciaranbor:
## Motivation

When a model instance is deleted (e.g. node disconnect, manual teardown), any in-flight SSE streaming connections for that instance hang indefinitely. The API never closes the response stream, so clients block forever waiting for more chunks.

## Changes

- Listen for `InstanceDeleted` events in the API event loop
- Add `_close_streams_for_instance()` to find and close any active text/image generation queues tied to tasks on the deleted instance
- Add unit tests covering text gen, image gen, and unrelated-instance-not-closed scenarios

## Why It Works

When an instance is deleted, we iterate `state.tasks` to find commands running on that instance, then close and remove their send-side queue handles. This causes the SSE generator to terminate, unblocking the client.

### Manual Testing
- This was causing issues for me on another branch (integration tests). Including this fix solved the issue

### Automated Testing
- `test_instance_deleted_stream_cleanup.py`: 3 tests covering text gen cleanup, image gen cleanup, and ensuring unrelated streams are not affected

### DIFF
--- a/src/exo/api/main.py
+++ b/src/exo/api/main.py
@@ -172,10 +172,20 @@ from exo.shared.types.events import (
     ChunkGenerated,
     Event,
     IndexedEvent,
+    InstanceDeleted,
     TracesMerged,
 )
 from exo.shared.types.memory import Memory
 from exo.shared.types.state import State
+from exo.shared.types.tasks import (
+    ImageEdits as ImageEditsTask,
+)
+from exo.shared.types.tasks import (
+    ImageGeneration as ImageGenerationTask,
+)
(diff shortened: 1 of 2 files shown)


Evanev7:
should we be getting the worker/runner supervisor to generate terminal chunks rather than catching failure in the api?

ciaranbor:
Maybe. `cancel_command` already handles this in the API though, and the API is there the stream lives. Plus this way, unexpected worker/runner failures are also handled

PR #1830, opened 2026-04-01
KV cache: per-conversation slot separation + disk persistence/offload

aidiffuser:
## Problem

`KVPrefixCache` matches slots purely by token prefix overlap. When multiple conversations share a system prompt (typical in multi-channel API gateway setups), the prefix matcher sees 90%+ overlap and overwrites the existing 
slot on every conversation switch — destroying the previous conversation's cache. A 100k-token conversation reverts to a ~20-30k system prompt match if you were to switch to a fresh session, requiring near-full reprefill. The alternative — keeping multiple slots in RAM — scales linearly with conversation count, length, and model architecture, quickly becoming prohibitive. With this patch, you only have 1 slot active at any given time so it prevents OOMs. Additionally, KV cache is RAM-only: restarts and crashes force cold prefills from zero. I am now able to hold multiple conversations in multiple channels with my agent running locally on my computer with near zero prefill downtime (instant switching from one session to the other) after every conversation is fully loaded/flushed to disk, I think of it as a context extender, I hold 3-5 conversations at a time all of which are around 70k-100k tokens on a model as large as Kimi K2.5. 

### Per-conversation slot separation

A heuristic in `_save_prefix_cache` distinguishes "same conversation extended" from "different conversation with shared prefix" using an exact prefix check:
```python
cached_len = len(cached_prompt)
is_extension = prefix_hit_length >= cached_len - 1 and cached_coverage >= hit_ratio
```

A legitimate update (same conversation, new messages appended) matches **all** cached tokens as a prefix — `prefix_hit_length >= cached_len - 1`. The new prompt is always longer, so `cached_coverage >= hit_ratio`.

A cross-conversation collision (different channel, shared system prompt) matches only a subset of the cached tokens — even at 98% overlap, the gap of hundreds of unmatched tokens at the end of the cached prompt fails the exact prefix check. This correctly routes to a new slot instead of overwriting.

### Single hot slot + disk persistence

Only **one KV cache slot stays in RAM** at a time (the "hot" slot for the active conversation). All other slots are persisted to disk via mlx-lm's `save_prompt_cache` / `load_prompt_cache`, which call `mx.save_safetensors` internally — writing directly from Apple Silicon unified memory to disk with no numpy intermediate, no bf16 casting, and no GPU→CPU copy.

**Conversation switch flow:**
1. New conversation arrives, hot slot doesn't match → flush hot slot to disk (~1-2s)
2. Search disk for a matching slot → load if found (~1-2s)  
3. If not found → fresh prefill (first time only)
4. Total switch cost: **~2-4s** vs **~2-3 minutes** cold prefill

**Three flush triggers:**
- **15-second idle timer:** Saves the hot slot when the runner has no active tasks for 15+ seconds. Protects against data loss from crashes.
- **Conversation switch:** Immediate synchronous flush before clearing RAM and loading the new conversation.
- **Shutdown:** Force flush on runner shutdown, bypassing the 15-second timer.

## Storage format

```
~/.exo/kv-cache/<model_hash>/
  ├── slot_0_cache.safetensors      # KV tensors (bf16)
  ├── slot_0_tokens.safetensors     # Token array (int32, ~200 KB)
  ├── slot_0_meta.json              # {model_id, token_count, timestamp}
  ├── slot_1_cache.safetensors
  ├── slot_1_tokens.safetensors
  └── slot_1_meta.json
```

Each model gets its own directory via `model_hash` (SHA-256[:16] of model_id). No cross-model contamination.

## Disk eviction

Stale disk slots are cleaned up automatically after every flush:

- **24-hour TTL:** Any slot with a `timestamp` older than 24 hours is deleted. Active conversations have their timestamp refreshed on every flush, so they never expire while in use.
- **No slot count cap:** Disk space is cheap on NVMe SSDs. 20 concurrent conversations ≈ 100 GB — negligible on multi-TB drives. A hard cap would risk evicting active conversations during testing bursts.
- **Hot slot protection:** The currently loaded slot is never evicted, even if its metadata timestamp is stale.
- **Three files per eviction:** `_cache.safetensors`, `_tokens.safetensors`, and `_meta.json` are all removed together.

## Model compatibility

This feature works with any model using standard `KVCache` from mlx-lm, which includes the vast majority of architectures:

- ✅ Standard attention (Llama, Mistral, Qwen, Gemma, Phi, etc.)
- ✅ Multi-Latent Attention / MLA (DeepSeek V3, Kimi-K2.5)
- ✅ Grouped Query Attention / GQA (Llama 3, Mistral, etc.)
- ⚠️ `QuantizedKVCache` — untested but should work via `from_state()` protocol
- ⚠️ `RotatingKVCache` (sliding window) — untested, may need verification
- ❌ Models with `ArraysCache` / SSM layers (Mamba, Jamba) — SSM state snapshots are not persisted to disk. These models fall through gracefully to fresh cache rather than serving corrupt state.

## Configuration

All optional. No configuration required — the feature works out of the box with sensible defaults.

| Variable | Default | Effect |
|---|---|---|
| `EXO_KV_DISK_PERSISTENCE=0` | `1` (enabled) | Disable disk persistence entirely (revert to RAM-only behavior) |
| `EXO_KV_DISK_PATH=/path` | `~/.exo/kv-cache` | Custom storage location (e.g., faster NVMe, larger drive) |
| `EXO_KV_DISK_TTL_HOURS=48` | `24` | Hours before stale slot eviction |
| `EXO_KV_DISK_MAX_SIZE_GB=200` | `500` | Evict oldest slots when directory exceeds this limit |

### Before (Checkpoint 1 only)
```
Channel A:  50k tokens → saved at slot 0
Channel B:  35k tokens → matches slot 0 at 92% → OVERWRITES slot 0
Channel A:  50k tokens → slot 0 is now Channel B → full prefill (~2-3 minutes)
Runner restart: all cache lost → full prefill for everything
```

### After
```
Channel A:  50k tokens → hot slot (index 0), saved to disk as slot_0
Channel B:  35k tokens → flush slot_0 to disk, load or create slot_1, 99.2% hit
Channel A:  50k tokens → flush slot_1 to disk, load slot_0 from disk, 99.1% hit
Runner restart: load from disk → 99% hit, no cold prefill
```

### Measured performance

| Metric | Before | After |
|---|---|---|
| Same-conversation hit rate | 93-99% | 98-99% |
| Cross-conversation hit rate | 0% (overwritten) | 98-99% (separate slots) |
| Conversation switch cost | ~2-3 minutes (full 50k prefill) | ~2-4s (disk swap) |
| Cache survives restart | No | Yes |

## Testing

Tested over multiple hours on:
- **Hardware:** 2× Mac Studio M3 Ultra (512 GB unified memory each), Thunderbolt 5 RDMA
- **Model:** Kimi-K2.5 (DeepSeek V3 architecture, MLA attention, 61 layers, bf16)
- **Conversations:** 4 concurrent Discord channels with 50k-100k token contexts, 0 prefill, instant switching.
- **Scenarios validated:**
  - Repeated channel switching with 98-99% hit rates on both channels
  - Runner reload with immediate cache restoration from disk
  - 15-second idle flush firing correctly between requests
  - Force flush on shutdown preserving cache state
  - 24-hour TTL eviction of stale slots

## Files changed

| File | Changes |
|---|---|
| `cache.py` | `model_id` param, disk persistence methods, single hot slot enforcement, TTL eviction, disk search fallback in `get_kv_cache` |
| `runner.py` | Pass `model_id` to `KVPrefixCache`, flush on idle (AllTasksComplete), force flush on shutdown |
| `generate.py` | `cached_coverage` guard in `_save_prefix_cache` |
| `batch_generate.py` | Same `cached_coverage` guard (active code path) |

### DIFF
--- a/src/exo/worker/engines/mlx/cache.py
+++ b/src/exo/worker/engines/mlx/cache.py
@@ -91,7 +91,8 @@ class KVPrefixCache:
         self._group = group
         # Disk persistence state
         self._model_id = model_id
-        self._disk_dir = self._init_disk_dir() if model_id else None
+        self._disk_enabled = os.environ.get("EXO_KV_DISK_PERSISTENCE", "1") != "0"
+        self._disk_dir = self._init_disk_dir() if model_id and self._disk_enabled else None
         self._disk_dirty = False
         self._flush_requested_at: float = 0.0
         self._hot_slot_disk_id: int | None = None
@@ -272,7 +273,8 @@ class KVPrefixCache:
 
     def _init_disk_dir(self):
         h = hashlib.sha256(self._model_id.encode()).hexdigest()[:16]
-        d = _Path.home() / ".exo" / "kv-cache" / h
+        base = _Path(os.environ.get("EXO_KV_DISK_PATH", str(_Path.home() / ".exo" / "kv-cache")))
+        d = base / h
         d.mkdir(parents=True, exist_ok=True)
         logger.info(f"KV cache disk dir: {d}")
         return d
@@ -322,10 +324,12 @@ class KVPrefixCache:
         except Exception as e:
             logger.warning(f"KV cache disk flush failed: {e}")
 
-    def _evict_stale_disk_slots(self, max_age_hours=24, max_size_gb=500):
-        """Delete disk slots older than max_age_hours, then evict oldest if over max_size_gb."""
+    def _evict_stale_disk_slots(self):
+        """Delete stale disk slots (TTL) then evict oldest if over size limit."""
         if not self._disk_dir:
             return
+        max_age_hours = float(os.environ.get("EXO_KV_DISK_TTL_HOURS", "24"))
+        max_size_gb = float(os.environ.get("EXO_KV_DISK_MAX_SIZE_GB", "500"))
         # Phase 1: TTL eviction
         cutoff = _time.time() - (max_age_hours * 3600)
         for meta_file in self._disk_dir.glob("slot_*_meta.json"):
--- a/src/exo/worker/engines/mlx/cache.py
+++ b/src/exo/worker/engines/mlx/cache.py
@@ -301,13 +301,20 @@ class KVPrefixCache:
         if len(self.caches) == 0:
             return
         try:
-            from mlx_lm.models.cache import save_prompt_cache
+            from mlx.nn.utils import tree_flatten
             self._disk_dir.mkdir(parents=True, exist_ok=True)
             slot_id = self._hot_slot_disk_id if self._hot_slot_disk_id is not None else self._next_disk_slot_id()
             base = self._disk_dir / f"slot_{slot_id}"
             # Save cache (atomic: write tmp then rename)
+            # Inline save_prompt_cache with empty-array filter for GLM/other models
+            cache = list(self.caches[0])
+            cache_data = dict(tree_flatten([c.state for c in cache]))
+            cache_data = {k: v for k, v in cache_data.items() if v.size > 0}
+            cache_info = [c.meta_state for c in cache]
+            cache_classes = [type(c).__name__ for c in cache]
+            cache_metadata = dict(tree_flatten([cache_info, {}, cache_classes]))
             tmp_cache = str(base) + "_tmp_cache.safetensors"
-            save_prompt_cache(tmp_cache, list(self.caches[0]))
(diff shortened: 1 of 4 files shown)


aidiffuser:
Closing for now — reworking this KV-cache disk-persistence feature onto the current Zenoh-based main and validating it across model architectures. Will resubmit a fresh PR once it's ready. Thanks!

PR #1832, opened 2026-04-02
fix: force gc + clear_cache after KV prefix cache eviction

adurham:
## Summary
- After `KVPrefixCache` evicts LRU entries, the MLX Metal buffers stay allocated until Python's GC runs
- This leaks ~3-4 GB between long-context requests, reducing the effective context ceiling for back-to-back requests
- Adding `gc.collect()` + `mx.clear_cache()` after eviction frees Metal buffers promptly

## Test plan
- [x] Measured on 2-node PP cluster with Qwen3.5-397B-A17B-4bit at 63K context
- [x] Before: 108.88 GB retained after eviction (3.78 GB above baseline)
- [x] After: 105.48 GB retained after eviction (0.38 GB above baseline — draft model KV + minor overhead)
- [x] `gc.collect()` adds ~2-3ms latency, runs once per eviction cycle (not per token)
- [ ] Verify with `uv run pytest`


rltakashige (approved):
Sounds reasonable! It's a little hard to test, but I believe this.

PR #1835, opened 2026-04-02
Fix BatchGenerator in line with upstream refactor (and prevent Qwen3.5 memory leak)

rltakashige:
## Motivation

MLX LM has had a massive refactor to their BatchGenerator recently. Since we'd like new features from MLX LM such as Gemma 4, we need to update the code to handle this.

Additionally this fixes a significant memory leak in GatedDeltaNet (the difference is quite substantial, up to 1GB every 1000 tokens, explaining several memory issues users were facing with Qwen3.5 models)

## Testing
Before
image

After (no memory leak, as one of the changes upstream)
image

### DIFF
--- a/bench/exo_bench.py
+++ b/bench/exo_bench.py
@@ -238,7 +238,7 @@ def run_one_completion(
         "messages": [{"role": "user", "content": content}],
         "stream": False,
         "max_tokens": tg,
-        "logprobs": True,
+        "logprobs": False,
     }
 
     t0 = time.perf_counter()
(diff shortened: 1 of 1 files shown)


src/exo/worker/engines/mlx/patches/opt_batch_gen.py
```diff
@@ -1,94 +1,69 @@
...
+from typing import cast
 
 import mlx.core as mx
-from mlx_lm.generate import BatchGenerator, generation_stream
+from mlx_lm.generate import BatchGenerator, GenerationBatch, PromptProcessingBatch
 
 _PRECOMPUTE_TOP_K = 20
 
-_original_public_next = BatchGenerator.next
+_original_step = GenerationBatch._step
```
rltakashige: I think this is the only difficult part of the code to review because the diffs are useless. The rest of the code are (I hope) trivial changes and cleanup.

PR #1837, opened 2026-04-02
Add exo version and bench compat

rltakashige:
## Changes

Adds versioning to EXO bench.

When we update the version within EXO, old benchmark scripts (which will then be invalid) will no longer work with EXO.

## Testing

image


src/exo/worker/engines/mlx/generator/generate.py
```diff
@@ -218,6 +218,8 @@ def pipeline_parallel_prefill(
             model(prompt[-1:][None], cache=_prompt_cache)
             quantize_cache_fn(_prompt_cache)
         flush_prefill_sends()
+        if distributed_prompt_progress_callback is not None:
+            distributed_prompt_progress_callback()
```
Evanev7, line 222: seperate pr for this

PR #1857, opened 2026-04-08
Add Gemma 4 parallelism

rltakashige:


Wysie:
I fixed the issue mentioned here: https://github.com/exo-explore/exo/issues/1873. Attached the patch file (that can be directly applied onto this branch).

[gemma4_26b_quant_alignment_v2.patch](https://github.com/user-attachments/files/26664829/gemma4_26b_quant_alignment_v2.patch)

@Evanev7  @rltakashige

RWL-Dittrich:
I just tested the .patch file. 

Fixed the issue for me. It's now running with RDMA/Tensor parallelism on two M4 Pro Mac's

Wysie:
@RWL-Dittrich Thanks for checking!

dakecrazy:
so .70 can fix this issue ,why gemma4-26b -8bit only run one node ,fail to run in two nodes ,version .70

PR #1868, opened 2026-04-10
build: remove pyinstaller temp artifacts

MikkoParkkola:
## Summary
- remove PyInstaller `build/` leftovers after the package recipe completes
- keep the final packaged output intact

## Notes
- release artifacts that may be used in production are preserved
- cleanup only targets debug and intermediate build artifacts by default

## Validation
- `just --dry-run package`
- `git diff --check`

### DIFF
--- a/justfile
+++ b/justfile
@@ -34,6 +34,7 @@ build-dashboard:
 
 package: build-dashboard
     uv run pyinstaller packaging/pyinstaller/exo.spec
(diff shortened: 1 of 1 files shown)


Evanev7:
thanks for the contribution! will merge this after #1869 as it adds /build to the .gitignore

Evanev7 (approved):
sweet

PR #1869, opened 2026-04-10
add env override to macos app

ciaranbor:
## Motivation

- Let users pass/override arbitrary exo env vars from the macOS app without a code change.

## Changes

- `ExoProcessController.swift`: `CustomEnvironmentVariable` struct + `@Published` list persisted to `UserDefaults`, injected into the child process after built-ins.
- `SettingsView.swift`: new **Environment** tab with add/remove rows, trim + dedup on save, and a POSIX name validator with a warning badge.

## Why It Works

- Custom vars applied last in `makeEnvironment`, so overriding a built-in works with no special-casing.

### Manual Testing

- Set `EXO_LIBP2P_NAMESPACE` via the new UI; confirmed override in  `~/.exo/exo_log/exo.log`.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -38,3 +38,4 @@ bench/**/*.json
 
 # tmp
 tmp/models
+/build/exo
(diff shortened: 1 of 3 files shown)


Evanev7 (approved):
tyvm!

PR #1877, opened 2026-04-12
fix: use configured api_port for IP connectivity probes

lawrence3699:
## Motivation

Fixes #1861

When `--api-port` is set to a non-default value (e.g., `--api-port 55555`), the IP connectivity discovery system still probes peers on the hardcoded default port 52415. Since the API is not listening on 52415, all reachability checks fail, the topology reports zero reachable nodes, and the dashboard shows "No valid configurations for current settings."

## Changes

Thread the configured `api_port` from `Args` through `Worker` into the reachability probe functions:

- `net_profile.py`: `check_reachability()` and `check_reachable()` accept an `api_port` parameter (default 52415 for backward compatibility)
- `worker/main.py`: `Worker` stores `api_port` and passes it to `check_reachable()`, uses it in `Multiaddr` construction and the mDNS connection filter
- `main.py`: passes `args.api_port` to the `Worker` constructor

## Why It Works

The `/node_id` endpoint used by reachability probes is served by the FastAPI app, which binds to `args.api_port`. The probes must use the same port the API is actually listening on. Before this fix, the port was hardcoded in three places in `net_profile.py` and `worker/main.py`; now it uses the value from the CLI flag.

### Manual Testing

- Verified ruff passes on all changed files
- Code inspection: traced `api_port` flow from `Args.parse()` → `Node.create()` → `Worker.__init__()` → `_poll_connection_updates()` → `check_reachable()` → `check_reachability()` → HTTP probe URL

### Automated Testing
- No existing automated tests cover the reachability probe code path
- The new `api_port` parameter defaults to `52415`, so all existing behavior is preserved when `--api-port` is not specified

### DIFF
--- a/src/exo/main.py
+++ b/src/exo/main.py
@@ -40,6 +40,7 @@ class Node:
 
     node_id: NodeId
     offline: bool
+    _api_port: int
     _tg: TaskGroup = field(init=False, default_factory=TaskGroup)
 
     @classmethod
@@ -139,6 +140,7 @@ class Node:
             api,
             node_id,
             args.offline,
+            args.api_port,
         )
 
     async def run(self):
@@ -251,6 +253,7 @@ class Node:
                             download_command_sender=self.router.sender(
                                 topics.DOWNLOAD_COMMANDS
                             ),
+                            api_port=self._api_port,
                         )
                         self._tg.start_soon(self.worker.run)
                     if self.api:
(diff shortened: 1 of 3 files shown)


Evanev7 (approved):
thanks for the contribution!

Evanev7:
the cluster still needs to share an api port but hey, better than it was before imo.

Issue #1871, opened 2026-04-11
Bootstrap peers connect locally in tests but fail to surface a second node in real two-Mac runtime


OffensiveGeneric:
## Summary

On a two-Mac setup using Thunderbolt networking and bootstrap peers, the worker can be configured with the master's bootstrap multiaddr and both processes start normally, but the master never surfaces the worker as a connection or topology node.

This blocks distributed placement entirely: the master's topology remains single-node, so large models that should require multiple nodes fail placement with `No cycles found with sufficient memory`.

## Upstream version

Observed on `exo-explore/exo` at:
- commit `43b3df45fb4080819f0a29d9db2aa4b76df99501`

## Why this seems like an upstream bug

The bootstrap-peer networking tests appear to validate only the local in-process case:
- `rust/networking/tests/bootstrap_peers.rs`

Those tests pass the expectation that `FromSwarm::Discovered` is emitted for a bootstrap connection.

In the real two-Mac runtime, that event never seems to make it through on the master side.

## Environment

- 2 macOS machines
- direct Thunderbolt cable between the hosts
- bootstrap peers used explicitly, not relying on mDNS
- worker bootstrap peer:
  - `/ip4/169.254.250.121/tcp/54116/p2p/12D3KooWRRRW2zZ7xMwsqfTDhmRPDYUcP35kvE1pTGZKWcxMdv5k`

## Reproduction shape

1. Start master on Mac A with a fixed libp2p port.
2. Start worker on Mac B with `--bootstrap-peers` pointing at Mac A's Thunderbolt IP + libp2p port + peer ID.
3. Wait for both nodes to come up.
4. Inspect master state and logs.

## What I observed

Worker startup shows the configured bootstrap peer:

```text
Bootstrap peers: ['/ip4/169.254.250.121/tcp/54116/p2p/12D3KooWRRRW2zZ7xMwsqfTDhmRPDYUcP35kvE1pTGZKWcxMdv5k']
```

Both nodes subscribe normally to the connection topic:

```text
Topic Router TypedTopic(topic='connection_messages', publish_policy=<PublishPolicy.Never: 'Never'>, model_type=<class 'exo.routing.connection_message.ConnectionMessage'>) ready to send
Subscribed to connection_messages
```

But on the master, the topology remains single-node:

```json
{
  "nodes": [
    "12D3KooWRRRW2zZ7xMwsqfTDhmRPDYUcP35kvE1pTGZKWcxMdv5k"
  ],
[...]

PR #1880, opened 2026-04-12
Add direct master bootstrap resolution

OffensiveGeneric:
## Motivation

This change adds a simpler control-plane bootstrap path for multi-host EXO setups.

Today a worker must be started with a fully assembled `--bootstrap-peers` multiaddr that already includes the master's current libp2p peer ID. In practice that is brittle and awkward for real multi-machine bring-up, especially on the two-Mac Thunderbolt setup used for debugging issue #1871.

This PR implements the "standard IP connectivity to the master" direction discussed in that issue by letting a worker resolve the master's `node_id` over HTTP and construct the bootstrap multiaddr itself.

## Changes

- add `--master-host`
- add `--master-api-port`
- add `--master-libp2p-port`
- resolve the master peer ID from `http://<master-host>:<master-api-port>/node_id`
- build the bootstrap multiaddr automatically from that response
- reject invalid CLI combinations:
  - `--master-host` with `--bootstrap-peers`
  - `--master-host` without `--master-libp2p-port`
- log the resolved direct-master bootstrap path at startup

## Why It Works

The master already exposes `/node_id` on its API surface. Once the worker can read that value, it has everything it needs to build a valid libp2p bootstrap address for the master's listener:

- host/IP
- libp2p port
- peer ID

That removes the requirement for operators to manually discover and encode the current peer ID into `--bootstrap-peers`, while keeping the actual libp2p transport/bootstrap behavior unchanged once the address is constructed.

### Manual Testing

Hardware:
- Mac Studio, Apple M4 Max, 64 GB RAM (`192.168.1.21`)
- Mac Studio, Apple M4 Max, 64 GB RAM (`192.168.1.19`)
- direct Thunderbolt connection between the two Macs

What I did:
- started the master on `.21` with a fixed API port and libp2p port
- started the worker on `.19` using:
  - `--master-host`
  - `--master-api-port`
  - `--master-libp2p-port`
- verified the worker resolves `/node_id` and constructs a bootstrap multiaddr automatically
- verified the worker dials the resolved master libp2p listener
- verified this path is sufficient to reach a real two-node state on the active debug branch used for issue #1871

### Automated Testing

- no new automated tests in this PR
- validation is currently manual because the active failure mode is specific to live multi-host cluster bring-up on real hardware

### DIFF
--- a/src/exo/main.py
+++ b/src/exo/main.py
@@ -1,4 +1,5 @@
 import argparse
+import ipaddress
 import multiprocessing as mp
 import os
 import resource
@@ -7,6 +8,7 @@ from dataclasses import dataclass, field
 from typing import Self
 
 import anyio
+import httpx
 from loguru import logger
 from pydantic import PositiveInt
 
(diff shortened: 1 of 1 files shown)


Evanev7:
this is not necessary as far as im concerned. libp2p is capable of doing node id resolution itself.

PR #1888, opened 2026-04-13
Add prefix cache flag to exo bench

rltakashige:
## Motivation
For using Exo-Bench extensively, there are many cases that we could use prefix caching to speed up the benchmarks, especially when the focus is on the token generation.

At the same time, it's very clear that prefix caching decode tokens is not very useful in most current scenarios. Surprisingly, even for non-thinking models, the chat template means that a continued conversation will be formatted such that the existing cache is not effective.

We already (slightly accidentally) do this for the batch generator - we should do it for the sequential generator too.

## Changes

We can now speed up exo bench by having a use prefix caching flag. Of course, for most accurate pp results, it is better to not have it, but this speeds up tg and large benchmarking significantly.
Updated methodology to match

### Manual Testing
Tested on many configurations that the difference in results is negligible, even with multiple --pp options.

### DIFF
--- a/src/exo/worker/engines/mlx/generator/batch_generate.py
+++ b/src/exo/worker/engines/mlx/generator/batch_generate.py
@@ -74,7 +74,6 @@ class _EngineTask:
     all_prompt_tokens: mx.array
     prefix_hit_length: int
     matched_index: int | None
-    cache_snapshots: list[CacheSnapshot] | None
     detokenizer: StreamingDetokenizer
     on_generation_token: Callable[[], None] | None = None
     generated_text_parts: list[str] = field(default_factory=list)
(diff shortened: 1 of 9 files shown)


src/exo/worker/engines/mlx/generator/batch_generate.py
```diff
@@ -168,8 +173,6 @@ def submit(
                     f"cached ({100 * prefix_hit_length / len(all_prompt_tokens):.1f}%)"
                 )
                 prompt_tokens = remaining_tokens
-            else:
```
rltakashige, line 171: We don't need to do this twice

src/exo/worker/engines/mlx/generator/generate.py
```diff
@@ -637,6 +640,37 @@ def mlx_generate(
         )
     cache_snapshots: list[CacheSnapshot] | None = ssm_snapshots_list or None
 
+    if kv_prefix_cache is not None and matched_index is not None and is_exact_hit:
```
rltakashige, line 643: Save prefix cache at prefill time rather than decode time

rltakashige: We don't do this in the batch generator anyway atm

src/exo/worker/engines/mlx/generator/batch_generate.py
```diff
@@ -74,14 +74,14 @@ class _EngineTask:
     all_prompt_tokens: mx.array
     prefix_hit_length: int
     matched_index: int | None
-    cache_snapshots: list[CacheSnapshot] | None
```
Evanev7, line 77: was this just unused?

Evanev7 (commented):
do you actually use the prefill tps in the kv cache anywhere or just plumbing?

Evanev7 (approved):

PR #1889, opened 2026-04-14
Try harder to clean up processes nicely

rltakashige:
## Motivation

Model loading is actually quite reliable now. No need to kill if you have a slow SSD or it's a massive model; the user can shut the instance down if necessary.

This was a major cause of signal=9 issues although not the only one (can happen during inference too?).
The reason signal=9 is so bad is that RDMA will no longer work until restart if this ever happens.

## Changes

- no more model load timeout
- no more crazy sigkills
- try harder to clean up processes on model shutdown

### Manual Testing
Tested with some RDMA instances

### DIFF
--- a/src/exo/worker/engines/mlx/auto_parallel.py
+++ b/src/exo/worker/engines/mlx/auto_parallel.py
@@ -1,10 +1,8 @@
-import os
-import threading
 from abc import ABC, abstractmethod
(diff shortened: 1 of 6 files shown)


Evanev7 (approved):
BEAUTIFUL

PR #1892, opened 2026-04-14
Try convert models

rltakashige:
## Motivation

Support Qwen/Qwen3.5-397B-A17B, Qwen/Qwen3.5-27B-FP8 and other similar models by trying to convert where we can.

### Manual Testing
No regressions on existing models


rltakashige:
Let's add this at some point, not necessary for a while.

PR #1893, opened 2026-04-14
Fix Qwen3-VL and autodetect vision config

rltakashige:
## Motivation

Qwen3 VL TP doesn't work atm, and vision is not behaving.

### Manual Testing
Works now.


AlexCheema (approved):
looks good, provided it was tested

PR #1895, opened 2026-04-15
dashboard: improve readability of long model names

one27001:
## Motivation

Model IDs in the EXO catalog are often long HuggingFace names where the suffix carries important information, including model family, architecture, quantization, and variant. In the current dashboard, those names can be truncated in the model picker and selected-model control, making it harder to distinguish similar models.

## Changes

- Increase the model picker modal width from `min(90vw,600px)` to `min(95vw,1200px)`.
- Increase the selected-model control width in the chat form from `max-w-xs` to `max-w-xl`.
- Stop truncating the selected model label in the chat form.

## Why It Works

The dashboard already uses responsive viewport-bounded sizing. This keeps the picker constrained on smaller screens while giving desktop layouts enough room to show meaningful model identifiers. The selected-model control similarly gets more horizontal space so users can distinguish long model names after selection.

### Manual Testing

Hardware: Mac Studio M1 Max 64GB

What I did:
- Used this branch locally with long HuggingFace model IDs in the dashboard.
- Verified the wider model picker makes long model names easier to distinguish.
- Verified the selected model label remains readable in the chat form.

### Automated Testing

- `cd dashboard && npm run build`

The build passes. Existing Svelte accessibility/state/chunk-size warnings are unchanged by this PR.

### DIFF
--- a/dashboard/src/lib/components/ChatForm.svelte
+++ b/dashboard/src/lib/components/ChatForm.svelte
@@ -336,14 +336,14 @@
             >MODEL:</span
           >
           <!-- Model button — opens the full model picker -->
-          <div class="relative flex-1 max-w-xs">
+          <div class="relative flex-1 max-w-xl">
             <button
               type="button"
               onclick={() => onOpenModelPicker?.()}
               class="w-full bg-exo-medium-gray/50 border border-exo-yellow/30 rounded pl-3 pr-8 py-1.5 text-xs font-mono text-left tracking-wide cursor-pointer transition-all duration-200 hover:border-exo-yellow/50 focus:outline-none focus:border-exo-yellow/70"
             >
               {#if currentModelLabel}
-                <span class="text-exo-yellow truncate">{currentModelLabel}</span
+                <span class="text-exo-yellow">{currentModelLabel}</span
                 >
               {:else}
                 <span class="text-exo-light-gray/50">— SELECT MODEL —</span>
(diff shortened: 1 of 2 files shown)


AlexCheema:
@one27001 — tested locally on a MacBook Pro at 1280×800 viewport.

## 1. Model picker modal — `min(90vw, 600px)` → `min(95vw, 1200px)`

**Before (main):**
before modal

**After (this PR):**
after modal

⚠️ **The modal is too wide now** — at a 1280px viewport it eats ~94% of the screen (only ~40px of margin per side), which doesn't feel like a modal anymore. 600px was too tight, but 1200px overshoots. Could you bring it down to something like `min(90vw, 880px)` so long model IDs get more room without the picker swallowing the whole window?

## 2. Selected-model badge — `max-w-xs` (320px) → `max-w-xl` (576px), `truncate` removed

**Before (main):** label is constrained to 320px and overflows visually
before chatform

**After (this PR):** full label fits cleanly inside the badge
after chatform

This change looks good 👍 — long names render fully without ellipsis.

## Verified
- `cd dashboard && npm run build` — passes on this branch
- Chat-form badge change is a clear win
- Modal change works mechanically but the new max width is too large; please dial it back

one27001:
thanks @AlexCheema makes sense, i will reduce the width of that modal

one27001:
done, some of the really long ones will still overflow i think but we can't make this fit all possible lengths i agree

here's what it looks like on a 1280px viewport with some oddly long names from huggingface
image

Evanev7 (approved):

PR #1897, opened 2026-04-15
add assertion to event router

Evanev7:
followup to #1894, in order to ensure every receiver receives every event, we should ensure all are created before we start the event router.


rltakashige (approved):
Nice, merging this post release!

PR #1899, opened 2026-04-15
Add usage stats to tool calls and handle multiple tool calls correctly

rltakashige:
## Motivation

Tool calls are usually not end tokens, so they didn't have usage stats.


Evanev7 (approved):
we'll find out p soon if this has broken anything

PR #1901, opened 2026-04-15
Better environment variables in MacOS app

### LINKED ISSUE #1858: [BUG] Exo from dmg does not respect EXO_DEFAULT_MODELS_DIR
## Describe the bug

I have exo 1.0.69 installed on a macos system using the dmg installer. It appears that exo does not respect the `EXO_DEFAULT_MODELS_DIR` environment variable. I have exported it in both /etc/profile, and /etc/zshrc. Exo still uses the default `~/.exo/models` directory.

## To Reproduce

1. Install the app using the dmg installer.
2. Export the variable using the env file for your shell.
3. Reboot.

## Expected behavior

Exo picks up the env variable without having to modify the Info.plist file (which I assume would get overwritten on every update).

## Actual behavior

Nothing happens.

## Environment

- macOS Version: Tahoe 26.2
- EXO Version: 1.0.69
- Hardware:
  - Irrelevant
- Interconnection:
  - Irrelevant

## Additional context

No additional context.

rltakashige:
## Motivation

Closes #1858

### DIFF
--- a/app/EXO/EXO/ExoProcessController.swift
+++ b/app/EXO/EXO/ExoProcessController.swift
@@ -9,11 +9,14 @@ private let enableImageModelsKey = "EXOEnableImageModels"
 private let offlineModeKey = "EXOOfflineMode"
 private let fastSynchEnabledKey = "EXOFastSynchEnabled"
 private let onboardingCompletedKey = "EXOOnboardingCompleted"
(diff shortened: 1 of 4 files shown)


AlexCheema (approved):
looks good

PR #1910, opened 2026-04-16
Revert "Update mlx and mlx lm to latest"

rltakashige:
Reverts exo-explore/exo#1906

The new mlx update seems to have broken a bunch of stuff with RDMA. Seems far less reliable now, so reverting!

Maybe.. Seems to have been weird caching issues. Will merge if this is necessary.


rltakashige:
Doesn't seem necessary,  it did uncover some bugs in our tooling though

PR #1913, opened 2026-04-17
Override mlx to CPU-only on Linux for ROCm compatibility

yoichi-aoyama:


yoichi-aoyama:
I apologize. That was a mistake on my part.

PR #1923, opened 2026-04-18
Treat ghost completions as Pending

mlpy0:
## Summary
In the coordinator's periodic status scan, if the shard downloader reports `status="complete"` but `resolve_existing_model` returns `None` — i.e. the HuggingFace shard cache / index says complete, but no EXO model directory passes `is_model_directory_complete` — the code previously emitted `DownloadCompleted`. That re-asserts the `state.downloads` entry and advertises a model that has no usable weights on disk.

This PR emits `DownloadPending` in that branch instead, so the stale entry clears rather than being re-advertised.

This can happen when the EXO model dir is removed or partially cleaned up out of band while the HF cache (or an HF cache index reporting complete) is still present — a narrow but real scenario.

## Test plan
- [x] `uv run pytest src/exo/download/tests/test_ghost_completion.py`
- [x] `uv run pytest src/exo/download/tests/`
- [x] `uv run basedpyright src/exo/download/coordinator.py src/exo/download/tests/test_ghost_completion.py`
- [x] `uv run ruff check src/exo/download/coordinator.py src/exo/download/tests/test_ghost_completion.py`

### DIFF
--- a/src/exo/download/coordinator.py
+++ b/src/exo/download/coordinator.py
@@ -358,10 +358,13 @@ class DownloadCoordinator:
                                 progress.shard, found, progress.total
                             )
                         else:
-                            status = DownloadCompleted(
+                            # Probe reports complete but resolve_existing_model
+                            # found nothing on disk. Emit Pending so indexed
+                            # state clears any stale DownloadCompleted instead
+                            # of re-advertising a missing model.
+                            status = DownloadPending(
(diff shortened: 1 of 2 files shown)


Evanev7:
this is solving a downstream issue of the actual one - id prefer we never emitted "complete" from the shard downloader when the shard was not complete

PR #1929, opened 2026-04-19
feat(cache): radix-trie KV prefix cache for storage-level dedup

adurham:
## Summary
- Replaces the flat `KVPrefixCache` (one full KV copy per session) with a radix trie whose edges own per-layer K/V slices. Sessions sharing a token prefix share the underlying tensors for that prefix.
- Hermes-style multi-turn workloads with a common system prompt stop paying N× for the shared context — each new session adds only its unique suffix to KV memory.
- Preserves the external API (`add_kv_cache`, `update_kv_cache`, `get_kv_cache`, `pin`, `clear`, `.prompts`, `.caches`, `.prefill_tps`) and the SSM/rotating snapshot semantics.

## Motivation
Tonight's deploy hit M4-2 at 99% RAM under Hermes traffic once `maxPrefixSessions` was turned on for MiniMax — the flat cache stored every session's full KV independently, so 4 conversations with a shared system prompt duplicated that prompt's KV four times. This PR fixes the root cause rather than dialing sessions down.

## Design
- **Sliceable cache types** (`KVCache`, `QuantizedKVCache`, `RotatingKVCache` pre-rotation) live on trie edges. Each edge holds per-layer K/V slices covering its token range.
- **Non-sliceable types** (`ArraysCache`/SSM, post-rotation `RotatingKVCache`) remain per-leaf with snapshots — partial-prefix hits walk back to the nearest snapshot <= match depth, same as before.
- **Ref-counting**: every leaf increments ref_count up its ancestor chain. Evicting a leaf decrements; any ancestor whose count hits zero detaches from its parent and its K/V tensors are freed.
- **Eviction** is LRU over leaves, driven preemptively at the start of `add_kv_cache` so `len(leaves) < max_sessions` always holds after insertion (old behaviour allowed overflow by one).
- **Token traversal** materialises the 1-D token array to numpy once per insert/lookup. Avoids hammering the MLX evaluator with per-token `.item()` calls.

## Tests
6 new trie-specific tests on top of the existing suite:
- shared-prefix node creation and ref counts
- prefix-hit materialisation returns the right match depth + remaining tokens
- LRU eviction frees only the unique branch; shared prefix survives
- pin semantics prevent eviction under session cap
- `update_kv_cache` extends an existing leaf
- media-region content-hash mismatch truncates the match at region start

## Test plan
- [x] `uv run pytest src/exo/worker/tests/unittests/test_mlx/test_kv_prefix_cache.py` (18 passed)
- [x] `uv run pytest tests/test_vision_cache.py` (10 passed)
- [x] `uv run pytest src tests` (all exo tests pass except pre-existing Metal-kernel failures in `test_extract_top_logprobs.py` and `test_cancel_download.py` — unrelated, reproducible on `main`)
- [x] `uv run ruff check src/exo/worker/engines/mlx/cache.py src/exo/worker/tests/unittests/test_mlx/test_kv_prefix_cache.py` (clean)
- [ ] Roll to Studios and verify per-node RAM drops when multiple Hermes conversations share a system prompt


adurham:
Opened against the wrong fork — closing and re-homing on adurham/exo.

Issue #1924, opened 2026-04-18
[BUG] Minimax m2.7 mxfp4 thinks in chinese


leanzero-srl:
## Describe the bug

Minimax m2.7 mxfp4 thinks in chinese

## To Reproduce

1. Download minimax m2.7 mxfp4
2. Run with qwen code
3.Thinking is in chinese

## Expected behavior

English...
## Actual behavior

Chinese
## Environment

- macOS Version:
- EXO Version: 1.0.70
- Hardware:
  - Device 1: M4 max 128gb
  - Device 2: M3 ultra 96gb
  - Additional devices:
- Interconnection:
  - (e.g., Thunderbolt 4 cable between Device 1 and 2)
  - (e.g., WiFi 6 for Device 3)
  - (e.g., 10GbE Ethernet between all devices)

leanzero-srl:
The 4 bit one does the same thing.

leanzero-srl:
Tried with claude and it's the same thing. It';s not only that it answers in chinese, it also endlessly loops.

[exo.log](https://github.com/user-attachments/files/26854437/exo.log)

leanzero-srl:
This resolves it to an extent where it actually stops chinese thinking and gibberish 

1) Repetition loops on MiniMax-M2.x

**Symptom.** MiniMax-M2.x running under exo produced endless repetition loops (often in Chinese) on typical chat-completion requests. Reproduced on both M2.5 and M2.7 at 4-bit; indifferent to prompt content.

**Root cause.** `TextGenerationTaskParams.repetition_penalty` defaults to `None`. mlx_lm's `make_logits_processors` treats `None` as "install no repetition-penalty processor at all":
```python
# mlx_lm/sample_utils.py line ~117-124
for make_penalty, penalty, context_size in repetition_penalties:
    if penalty is not None and penalty != 0:
        logits_processors.append(make_penalty(penalty, context_size))
```
Clients like Claude Code / Qwen Code don't set `repetition_penalty` in the request body, so it arrives `None` and the model generates with zero suppression on repeating tokens. Thinking models degrade into loops fast under that config.

**Fix.** Per-model sampling defaults in the model card, resolved at the runner boundary.

- `src/exo/shared/models/model_cards.py` — new optional `SamplingDefaults` nested model on `ModelCard`. Fields: `temperature`, `top_p`, `top_k`, `min_p`, `repetition_penalty`, `repetition_context_size`, all optional.
[...]

PR #1935, opened 2026-04-19
fix: add per-model sampling defaults to prevent MiniMax repetition loops

qtkt2hmr2v-debug:
## Summary

This PR adds a `SamplingDefaults` infrastructure to `ModelCard` — a mechanism
to apply per-model sampling parameter defaults when the request omits them.
It also adds upstream-sourced defaults (`temperature`, `top_p`, `top_k`) to
all MiniMax model cards.

**Note on `repetition_penalty`:** Our manual testing revealed that setting
`repetition_penalty=1.1` in the request payload causes EXO to crash and
unload the model on complex prompts (see Test Plan). We are *not* including
`repetition_penalty` in the TOML defaults until the root cause of that crash
is understood. The infrastructure is in place to add it safely once resolved.

---

### `SamplingDefaults` (model_cards.py)

A new optional `SamplingDefaults` field on `ModelCard` — Pydantic-typed,
loaded from TOML, ignored when absent (zero-regression guarantee for all
existing models). `fill_missing()` merges card defaults with an overrides
dict; explicit caller-supplied values always win.

### `apply_sampling_defaults` (utils_mlx.py)

Pure function: fills `None` slots in `TextGenerationTaskParams` from card
defaults. Returns the original object unchanged when nothing needs filling
(no allocation in the common case). Emits a `logger.debug` line listing
each applied field and its value.

### Hook in `runner.py`

`submit_text_generation` looks up the model card and calls
`apply_sampling_defaults` before handing the task to the generator.
One card lookup per submitted task.

### Model card TOML updates (11 cards — M2.1, M2.5, M2.7 all quants)

```toml
[sampling_defaults]
temperature = 1.0  # from MiniMaxAI/MiniMax-M2.7 upstream generation_config.json
top_p = 0.95
top_k = 40
# repetition_penalty intentionally omitted — see issue #1924 and Test Plan
```

### Tests (test_sampling_defaults.py)

Parametrized pytest suite covering:
- card defaults fill absent fields
- explicit request values are never overwritten
- no-op fast path when defaults are empty (identity return)
- `fill_missing` dict merge semantics
- individual field preservation (5 parametrized cases)

---

## Motivation

MiniMax M2.x models produce repetition loops and empty response bodies
under exo when `enable_thinking=false`. Tracked in #1924.

During investigation we also found that `temperature`, `top_p` and `top_k`
are absent from all MiniMax model cards, even though MiniMaxAI's upstream
`generation_config.json` defines recommended values. This PR adds those.

The `repetition_penalty` story is more complex — see Test Plan below.

---

## Changes

- New `SamplingDefaults` Pydantic model + optional field on `ModelCard`
- `apply_sampling_defaults()` in utils_mlx.py (pure function, zero-allocation fast-path)
- Hook in `Runner.submit_text_generation()` — one card lookup per task
- `[sampling_defaults]` with `temperature/top_p/top_k` added to 11 MiniMax TOML cards
- Parametrized test suite in `test_sampling_defaults.py`

---

## Why It Works

Card defaults only fill `None` slots — explicit request values always win.
This means any existing API client that sends its own `temperature` is
unaffected. The infrastructure is generic and can be used for any model family.

---

### Manual Testing

**Hardware:** Mac Studio M3 Ultra 512 GB RAM + Mac Mini M4 Pro 48 GB,
exo cluster connected via Thunderbolt 5 RDMA.

**MiniMax-M2.7-6bit — `enable_thinking=false`, no `repetition_penalty`**

| Run | OK | Empty-body errors | Timeouts | Notes |
|-----|----|-------------------|----------|-------|
| Run A | 12/20 (60%) | 7 | 1 | Immediate failures on complex prompts |
| Run B (fresh cluster) | 17/20 (85%) | 0 | 3 | Hangs instead of fast-fail |

Both runs confirm M2.7-6bit is unreliable in `think=OFF`. In both cases EXO
entered a reload loop, requiring manual restart.

**MiniMax-M2.5-6bit — `enable_thinking=false`, no `repetition_penalty`**

5/5 ✅ on the hardest prompts (CAP theorem, CRDT, Quicksort worst-case,
Theseus paradox, Python regex). **Stable without any fix applied.**

**MiniMax-M2.5-6bit — `enable_thinking=false`, `repetition_penalty=1.1`**

| Prompt | Result |
|--------|--------|
| CAP theorem | ✅ OK |
| Quicksort O(n²) explanation | ❌ EXCEPTION at 48.5s → EXO crash |
| CRDT explanation | ❌ HTTP 404 (model unloaded after crash) |
| Theseus paradox | ❌ HTTP 404 |
| Python regex | ❌ HTTP 404 |

Adding `repetition_penalty=1.1` to the payload crashes EXO on M2.5 at the
first complex prompt, causing the model to unload entirely. This is why we
are **not** including `repetition_penalty` in the TOML defaults in this PR.
The crash root cause needs separate investigation — possibly mlx_lm's
repetition processor behavior on MiniMax architectures.

### Automated Testing

New: `src/exo/worker/engines/mlx/tests/test_sampling_defaults.py`

- 7 parametrized test cases covering all invariants
- Run with: `pytest src/exo/worker/engines/mlx/tests/test_sampling_defaults.py`

### DIFF
--- a/resources/inference_model_cards/mlx-community--MiniMax-M2.1-3bit.toml
+++ b/resources/inference_model_cards/mlx-community--MiniMax-M2.1-3bit.toml
@@ -13,3 +13,12 @@ context_length = 196608
 
 [storage_size]
 in_bytes = 100086644736
+
+# Sampling defaults sourced from MiniMaxAI upstream generation_config.json
+# (temperature=1.0, top_p=0.95, top_k=40) plus repetition_penalty=1.1 added
+# empirically to prevent repetition loops observed in exo (see issue #1924).
+[sampling_defaults]
+temperature = 1.0
+top_p = 0.95
+top_k = 40
+repetition_penalty = 1.1
--- a/resources/inference_model_cards/mlx-community--MiniMax-M2.1-8bit.toml
+++ b/resources/inference_model_cards/mlx-community--MiniMax-M2.1-8bit.toml
@@ -13,3 +13,12 @@ context_length = 196608
 
 [storage_size]
 in_bytes = 242986745856
+
+# Sampling defaults sourced from MiniMaxAI upstream generation_config.json
+# (temperature=1.0, top_p=0.95, top_k=40) plus repetition_penalty=1.1 added
+# empirically to prevent repetition loops observed in exo (see issue #1924).
+[sampling_defaults]
+temperature = 1.0
+top_p = 0.95
+top_k = 40
+repetition_penalty = 1.1
--- a/resources/inference_model_cards/mlx-community--MiniMax-M2.5-4bit.toml
+++ b/resources/inference_model_cards/mlx-community--MiniMax-M2.5-4bit.toml
@@ -13,3 +13,12 @@ context_length = 196608
 
 [storage_size]
 in_bytes = 128666664960
+
+# Sampling defaults sourced from MiniMaxAI upstream generation_config.json
+# (temperature=1.0, top_p=0.95, top_k=40) plus repetition_penalty=1.1 added
+# empirically to prevent repetition loops observed in exo (see issue #1924).
+[sampling_defaults]
+temperature = 1.0
+top_p = 0.95
+top_k = 40
+repetition_penalty = 1.1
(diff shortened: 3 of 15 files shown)


rltakashige:
Hi there! Thanks for this contribution!

I've added sampling defaults in #1947 as some functions here weren't wired up, and I wanted to fix the repetition penalty crash. 

I'll close this PR in favour of that, but thanks again!

PR #1938, opened 2026-04-19
fix(download): treat HF size=None as "size unknown, trust filesystem"

### LINKED ISSUE #1918: [BUG] Model downloaded status gets lost
## Describe the bug

On our two Mac Studio setup the "fully downloaded" status of Kimi K2.5 always gets lost after a few minutes.

What I mean with that is that it stops showing up in the models list when the `downlaoded` filter is enabled. 
In the Download manager it doesn't have the green checkmark but shows 612 / 612 GB (on both nodes).

Clicking the download arrow immediately makes the green check appear and then it also appears in the launcher with the downloaded filter enabled. 

However it reverts to the "almost downloaded" status after a very short time (1 - 2 minutes).

Other large models don't show this behaviour.

## To Reproduce

1. Download Kimi K2.5
2. Observe the download status

## Expected behavior

The downloaded status to stick

## Actual behavior

See above

## Environment

- macOS Version: 26.3
- EXO Version: 1.0.69
- Hardware:
  - Device 1: Mac Studio M3 Ultra 512GB
  - Device 2: Mac Studio M3 Ultra 512GB
- Interconnection:
  - TB cable

nhwaani:
Fixes #1918 — the "fully-downloaded status keeps reverting" bug reported for Kimi K2.5 on a 2-node Mac Studio setup: a model that's already 612/612 GB on disk loses its green checkmark every 1–2 minutes and disappears from the "downloaded" filter, reappearing only when the user clicks the download arrow.

## Root cause

`DownloadCoordinator._emit_existing_download_progress` runs a periodic re-scan every 60 s (`src/exo/download/coordinator.py:345`, `await anyio.sleep(60)`) that calls `shard_downloader.get_shard_download_status()` for every known model. That ultimately ends up in `download_shard(skip_download=True)` at `src/exo/download/download_utils.py:763`, which loops over the file list returned by HuggingFace's tree API and builds per-file status with:

```python
status="complete"
if final_file_exists and downloaded_bytes == file.size
else "not_started",
```

`FileListEntry.size` is `int | None`. For very large repos (Kimi K2.5 is the reproducer) HuggingFace's tree API can return `size=None` for at least one entry. The moment that happens:

1. `downloaded_bytes == None` is always `False`, regardless of what's on disk
2. That file gets tagged `not_started`
3. `calculate_repo_progress` (line 654) then classifies the whole repo as `in_progress` (any not-complete → repo not complete)
4. Back in the coordinator periodic loop:
   ```python
   elif progress.status in ["in_progress", "not_started"]:
       if progress.downloaded_this_session.in_bytes == 0:
           status = DownloadPending(...)    # ← demotes Completed → Pending
   ```
5. The `DownloadCompleted` status gets overwritten every 60 s, which the UI renders as losing the green checkmark and being removed from the "downloaded" filter.

Why clicking the download arrow fixes it temporarily: `StartDownload` goes through `_start_download` which calls `resolve_existing_model` → `is_model_directory_complete`. That verification uses the safetensors index on disk (`_scan_model_directory`), not the HF tree API, so it correctly reports "complete". But 60 s later the periodic scan runs again and demotes it.

Why other large models don't trigger this: they happen not to contain any file for which HF's tree API returns `size=None`. Kimi K2.5 does, presumably because of its size/structure (384 safetensor shards plus custom Python modeling files and assorted configs).

## The fix

In `download_shard`, when `file.size is None` we have no authoritative remote size to compare against — but we still know whether the file exists on disk. File existence is a strictly stronger signal than absence of HF-reported size info, so the check becomes:

```python
status="complete"
if final_file_exists
and (file.size is None or downloaded_bytes == file.size)
else "not_started",
```

The `total` field is also adjusted from a hard-coded `0` to the on-disk byte count in the `size=None` case, so the UI shows an accurate total.

All other verification paths are untouched. In particular, `_download_file` (the active-download path that verifies hashes and sizes during actual downloads) is unchanged — this PR only loosens the no-remote-size-available branch of the passive status check.

## Tests

Three new regression tests in `TestSkipDownloadStatusWithMissingSize`:

| Test | Asserts |
|---|---|
| `test_status_complete_when_file_size_is_none` | Fully-downloaded model with a `size=None` entry reports `"complete"` |
| `test_status_not_started_when_file_with_none_size_missing` | A file with `size=None` that's **missing** on disk still reports `"not_started"` — the fix only waives size-verification, not existence-verification |
| `test_status_not_started_when_sized_file_size_mismatch` | A genuine size mismatch on a sized file still reports `"not_started"` (no regression) |

All three pass; existing 44 download-module tests continue to pass; full non-slow test suite (364 tests) passes.

## Why I can't verify end-to-end

I don't have the hardware to reproduce the original report (no 2× Mac Studio, no 612 GB free for Kimi K2.5). The fix is derived from reading the code path, and the unit tests encode the exact scenario (`FileListEntry(path=..., size=None)` + file present on disk). Verification against a real Kimi K2.5 download would be valuable before merge.

## Checklist

- [x] `uv run ruff check` passes
- [x] `uv run basedpyright` passes with 0 errors, 0 warnings
- [x] `uv run ruff format --check` passes
- [x] `uv run pytest` passes (364 passed, 1 skipped, 154 deselected)
- [ ] `nix fmt` — not available in my dev environment; `ruff format` was used as the Python-side proxy
- [x] Single focused change per CONTRIBUTING.md

### DIFF
--- a/src/exo/download/download_utils.py
+++ b/src/exo/download/download_utils.py
@@ -900,17 +900,26 @@ async def download_shard(
     for file in filtered_file_list:
         downloaded_bytes = await get_downloaded_size(target_dir / file.path)
         final_file_exists = await aios.path.exists(target_dir / file.path)
+        # HuggingFace's tree API occasionally returns ``size=None`` for entries
+        # it can't report a size for (observed in the wild for very large repos
+        # such as Kimi K2.5). When that happens, ``downloaded_bytes == None`` is
+        # always False and the file would be permanently marked ``not_started``
+        # — which causes the periodic re-scan in ``DownloadCoordinator`` to demote
+        # a fully-downloaded model back to ``DownloadPending`` every 60s.
+        # File existence is a strictly stronger signal than absence of HF-reported
+        # size info, so trust the filesystem in that case.
         file_progress[file.path] = RepoFileDownloadProgress(
             repo_id=model_id,
             repo_revision=revision,
             file_path=file.path,
             downloaded=Memory.from_bytes(downloaded_bytes),
             downloaded_this_session=Memory.from_bytes(0),
-            total=Memory.from_bytes(file.size or 0),
+            total=Memory.from_bytes(file.size or downloaded_bytes),
             speed=0,
             eta=timedelta(0),
             status="complete"
-            if final_file_exists and downloaded_bytes == file.size
+            if final_file_exists
+            and (file.size is None or downloaded_bytes == file.size)
             else "not_started",
             start_time=time.time(),
         )
--- a/src/exo/download/tests/test_download_verification.py
+++ b/src/exo/download/tests/test_download_verification.py
@@ -12,11 +12,14 @@ from pydantic import TypeAdapter
 
 from exo.download.download_utils import (
     delete_model,
+    download_shard,
     fetch_file_list_with_cache,
 )
+from exo.shared.models.model_cards import ModelCard, ModelTask
 from exo.shared.types.common import ModelId
 from exo.shared.types.memory import Memory
 from exo.shared.types.worker.downloads import FileListEntry, RepoFileDownloadProgress
+from exo.shared.types.worker.shards import PipelineShardMetadata, ShardMetadata
 
 
 @pytest.fixture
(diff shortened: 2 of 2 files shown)


nhwaani:
@ciaranbor

ciaranbor:
Thanks! I'll try to reproduce and verify the fix tomorrow

nhwaani:
> Thanks! I'll try to reproduce and verify the fix tomorrow

Keep me updated, Thanks

ciaranbor:
I haven't been able to reproduce this issue. Kimi k2 downloads fully and stays downloaded for me

ciaranbor:
I found the same issue naturally with a different model. The issue was that the remote HF repo had updates which broke the file comparison. Fixed by https://github.com/exo-explore/exo/pull/1958

nhwaani:
> I found the same issue naturally with a different model. The issue was that the remote HF repo had updates which broke the file comparison. Fixed by #1958

Thanks for the fix

PR #1950, opened 2026-04-21
fix: use official mlx==0.31.1 from PyPI instead of custom git fork

jmalekos:
## Summary

- Removes the `[tool.uv.sources]` override that pointed `mlx` at a custom dev fork (`rltakashige/mlx-jaccl-fix-small-recv`, branch `address-rdma-gpu-locks`)
- `uv.lock` now resolves `mlx==0.31.1` + `mlx-metal==0.31.1` from PyPI

## Why

The custom fork predates macOS 26 Tahoe's updated Metal toolchain. On macOS 26, loading any MLX model produced:

```
unknown type name 'bfloat16_t'
unknown type name 'complex64_t'
```

`mlx==0.31.1` (released on PyPI) ships `mlx-metal==0.31.1` which includes Metal shaders compiled against the macOS 26 SDK. The fix the custom fork was carrying is now upstream in the official release.

## Test plan

- [x] `uv sync` resolves `mlx==0.31.1` from PyPI (no git checkout)
- [x] Metal matrix multiply test passes on macOS 26.3.1 arm64: `Metal OK: [[7.0, 10.0], [15.0, 22.0]]`
- [x] 361 tests pass (`uv run pytest`)

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -70,7 +70,6 @@ members = ["rust/exo_pyo3_bindings", "bench"]
 
 [tool.uv.sources]
 exo-pyo3-bindings = { workspace = true }
-mlx = { git = "https://github.com/rltakashige/mlx-jaccl-fix-small-recv.git", branch = "address-rdma-gpu-locks", marker = "sys_platform == 'darwin'" }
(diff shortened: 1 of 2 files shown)


Evanev7:
please raise an issue if one exists, not everything in our fork has been upstreamed.

PR #1955, opened 2026-04-21
Expose sampling_defaults on /v1/models

mlpy0:
#1947 added `[sampling_defaults]` blocks to model cards, but the `/v1/models` response schema didn't include the field — the parsed defaults never made it over the wire to clients.

This wires `ModelCard.sampling_defaults` through `ModelListModel` so clients can surface or apply the recommended per-mode sampling values without re-reading the TOMLs themselves.

### DIFF
--- a/src/exo/api/main.py
+++ b/src/exo/api/main.py
@@ -1675,6 +1675,7 @@ class API:
                     base_model=card.base_model,
                     capabilities=card.capabilities,
                     context_length=card.context_length,
+                    sampling_defaults=card.sampling_defaults,
                 )
                 for card in cards
             ]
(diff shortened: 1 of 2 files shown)


Evanev7:
the defaults don't need to be applied by clients. do you have something in mind?

mlpy0:
> the defaults don't need to be applied by clients. do you have something in mind?

I'm running a separate frontend on top of exo, not the exo dashboard. On model switches I want the recommended sampling params applied automatically (thinking vs non-thinking picked per request), without duplicating them in my client or re-parsing the TOMLs. /v1/models is the natural place to read them from.

If it doesn't fit your design, I can close it

Evanev7:
they are currently applied by the api if no overrides are passed- thats all. you shouldnt need to ingest these values unless you want further front end integration.

PR #1956, opened 2026-04-21
Handle missing total_size in safetensors index files

ciaranbor:
## Motivation

Image models fail to load after a mid-download instance deletion and recreation. The system skips the download and crashes with `FileNotFoundError: No safetensors files found in .../vae`.

## Changes

- Make `ModelSafetensorsIndexMetadata.total_size` optional (`PositiveInt | None = None`)
- Add null guard in `fetch_safetensors_size`
- Add regression test

## Why It Works

Exolabs quantized image models have safetensors index files with mflux metadata (`quantization_level`, `mflux_version`) but no `total_size`. The required `PositiveInt` field caused Pydantic validation to fail, which was silently swallowed by `except Exception: continue` in `_scan_model_directory`. This skipped all weight map checks, making incomplete models appear complete.

### Manual Testing

- Hardware: Mac Studio
- Before: `CreateRunner → LoadModel` (crash). After: `CreateRunner → DownloadModel` (correct).

### Automated Testing

- `test_safetensors_index.py`: 3 cases covering missing, valid, and null metadata

### DIFF
--- a/src/exo/download/tests/test_safetensors_index.py
+++ b/src/exo/download/tests/test_safetensors_index.py
@@ -1,0 +1,35 @@
+"""Tests for safetensors index file parsing."""
+
+from exo.shared.types.worker.downloads import ModelSafetensorsIndex
+
+
+def test_safetensors_index_missing_total_size():
+    """Index files from image models (e.g. FLUX/mflux) have metadata without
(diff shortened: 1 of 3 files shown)


Evanev7 (approved):
man i hate except exception:

PR #1958, opened 2026-04-21
Ignore HF remote repo changes (temporary fix)

### LINKED ISSUE #1918: [BUG] Model downloaded status gets lost

ciaranbor:
## Motivation

Fixes #1918. Downloaded model status reverts from "completed" to "pending" during each download scan. Reproduced with `zai-org/GLM-5.1`.

## Changes

- `coordinator.py`: In the periodic rescan, don't downgrade already-completed models; fall back to `resolve_existing_model()` (safetensors weight check) when per-file size check reports incomplete
- New `test_download_status_not_lost.py`: 3 regression tests

## Why It Works

The rescan compares local file sizes against HF's `main` revision. When HF updates text files (README, jinja, etc.), remote sizes change but local files still match the old revision — causing a false "incomplete". The fix uses the safetensors weight check as ground truth instead.

Long-term: pin the downloaded revision SHA rather than always checking against `main`.

### Manual Testing

- Mac Studio M3 Ultra with GLM-5.1 downloaded (natural reproduction of the issue)
- Confirmed GLM-5.1 stays `DownloadCompleted` through multiple rescan cycles

### Automated Testing

- 3 new tests: completed-not-downgraded, fallback-to-resolve, genuinely-incomplete-stays-pending

### DIFF
--- a/src/exo/download/coordinator.py
+++ b/src/exo/download/coordinator.py
@@ -365,7 +365,28 @@ class DownloadCoordinator:
                                 model_directory=self._default_model_dir(model_id),
                             )
                     elif progress.status in ["in_progress", "not_started"]:
-                        if progress.downloaded_this_session.in_bytes == 0:
+                        # TODO(ciaran): temporary solution
+                        # Don't downgrade a model that is already confirmed complete.
+                        if isinstance(
+                            self.download_status.get(model_id), DownloadCompleted
+                        ):
+                            continue
+                        # The per-file size check compares local files against
+                        # the latest HF "main" revision, which is a moving
+                        # target.  When HF updates text files (README, YAML,
+                        # jinja) in a new commit, the cached file list has new
+                        # sizes while local files still match the old revision.
+                        # Fall back to the authoritative completeness check
+                        # (is_model_directory_complete) which validates that all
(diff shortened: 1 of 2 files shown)


Evanev7 (approved):
explicitly temporary, sgtm

PR #1967, opened 2026-04-22
Fix Gemma 4 E2B TP + DeepSeek V32 thinking parsing

rltakashige:


Evanev7 (approved):

AlexCheema (approved):
- make sure we test deepseek v3.2
- going backwards on transformers version is a little scary

PR #1978, opened 2026-04-25
Add DeepSeek V4 Flash/Pro

### LINKED ISSUE #1974: we need deepseekv4 please
**EMERGENCY**  We need ds-v4 both pro and flash! PLEASE!!!

rltakashige:
Wait for upstream merge.


src/exo/worker/engines/mlx/patches/v4_offset_sync.py
```diff
@@ -0,0 +1,68 @@
...
+        key: str = "compressor",
+    ) -> mx.array:
+        if isinstance(offset, mx.array) and _current_int_offset is not None:
+            offset = _current_int_offset
+        return original_compressor_call(self, x, cache, offset, key)
+
+    Compressor.__call__ = patched_compressor_call
+
+
+apply()
```
Evanev7: patch is double applied at import time and at call time.

Evanev7 (approved):

PR #1981, opened 2026-04-25
Stabilize GPT-OSS tool support

team-wcv:
## Summary
- advertise tool support for GPT-OSS 20B/120B model cards
- let custom model cards override cached built-ins and refresh without restarting
- prevent batch runner template/history errors from crashing or wedging the active model instance

## Validation
- uv run pytest src/exo/shared/tests/test_model_cards.py src/exo/worker/tests/unittests/test_runner/test_batch_generator_errors.py src/exo/worker/tests/unittests/test_runner/test_parse_gpt_oss.py
- live /v1/models reports gpt-oss 20B/120B as text,thinking,tools
- live GPT-OSS 120B Responses tool-call probe emitted a function_call and stayed RunnerReady after tool output replay


team-wcv:
Closing in favor of #2062 ('Stabilize GPT-OSS tool support; echo reasoning config in Responses API'), which supersedes this PR with a more complete fix and adds the reasoning-config echo behaviour. Thanks for the original review attention.

PR #1982, opened 2026-04-26
fix: walk nested Thunderbolt _items so hub-attached peers register

team-wcv:
## Summary

`system_profiler SPThunderboltDataType -json` represents transparent TB hubs (e.g. an iVANKY Fusiondock Ultra, OWC TB hub, etc.) as an extra layer in `_items`, nesting the peer Mac one level deeper than a direct cable would. Apple makes the call asymmetrically: the side whose receptacle is plugged directly into the hub sees the hub at `_items[0]` with the peer Mac at `_items[0]._items[0]`, while the other side sees the peer at `_items[0]` directly.

The previous parser walked only the top level of `_items` and returned `None` if it didn't find a `domain_uuid_key` there. So on the "deeper" side, `MacThunderboltConnections` events never included that link, and the master ended up with an asymmetric RDMA mesh: the connection was visible in one direction but missing in the other. That blocked all-to-all RDMA cycles, which the JACCL backend requires for 3-way placement (`MlxJaccl`).

## Changes

* Extend `_ConnectivityItem` to recursively model `_items`.
* Add a depth-first helper, `_first_descendant_domain_uuid`, and use it from `ThunderboltConnectivityData.conn()`. The peer endpoint always has a `domain_uuid_key`, regardless of how many transparent hub layers sit between us; the helper returns the first one it finds.
* Three deterministic unit tests in `test_tb_parsing.py`: the iVANKY-hub case observed in production, a direct-cable sanity case, and an empty-receptacle case.

## Repro

`system_profiler SPThunderboltDataType -json` from a Mac whose TB cable to its peer goes through a hub, then check whether `_items[0].domain_uuid_key` is `null` while `_items[0]._items[0].domain_uuid_key` is the peer.

## Verification

Reproduced on a 3-node `wc-bmbp ↔ wc-smbp ↔ wc-smbpt` cluster with one cable routed through an iVANKY Fusiondock Ultra. Before the fix: `bmbp → smbp` RDMA edge missing, all 5 other RDMA edges present; `Pipeline + MlxJaccl` 3-way placement failed with "no RDMA-connected cycles available". After the fix: 6/6 directional RDMA pairs and `Pipeline + MlxJaccl` 3-way passes.

## Test plan

- [x] `uv run pytest src/exo/utils/info_gatherer/tests/test_tb_parsing.py` — 5/5 passing
- [x] `uv run basedpyright src/exo/shared/types/thunderbolt.py src/exo/utils/info_gatherer/tests/test_tb_parsing.py` — 0 errors
- [x] `uv run ruff check` — clean on touched files
- [x] Manual: cluster restart shows `bmbp → smbp rdma_en6 → rdma_en6` edge in `/state.topology`
- [x] Manual: `/instance/previews?model_id=mlx-community/gpt-oss-120b-MXFP4-Q8` shows `Pipeline + MlxJaccl` `OK` at 3 nodes (was `FAIL`)


Evanev7:
does inference work with this setup? any perf degradations?

team-wcv:
Updated the branch against current `exo-explore/main` and resolved the conflicts.

On the inference question: this PR's functional change is limited to walking nested Thunderbolt `_items` during topology discovery so hub-attached peers register; it does not change the generation path or placement scoring. I did not run a full multi-node inference benchmark in this workspace.

Targeted checks passed on the conflict-resolved branch:

- `uv run ruff check src/exo/worker/runner/llm_inference/model_output_parsers.py src/exo/worker/runner/llm_inference/batch_generator.py`
- `EXO_DASHBOARD_DIR=/tmp uv run pytest src/exo/shared/tests/test_model_cards.py`

Evanev7:
im not merging this if you haven't tested it - i view it as niche hardware support and without proper validation there is no point merging.

PR #1984, opened 2026-04-26
fix(dashboard): treat RunnerIdle as ready status

raajkumars:
## Summary
- map RunnerIdle in dashboard instance status derivation
- treat RunnerIdle as READY so active instances do not display as PREPARING

## Why
Runners can report RunnerIdle between requests. This state was not handled, which produced false PREPARING status in the dashboard.

## Validation
- dashboard build succeeds
- state-based status derivation resolves RunnerIdle to READY


Evanev7:
idle runners are not ready to serve requests. preparing is appropriate - this is a synchronization step.

PR #1989, opened 2026-04-26
fix: route by in-flight tasks only — completed tasks were skewing load balance

adurham:
The load balancer counted ALL tasks (Complete, Cancelled, TimedOut, Failed) instead of only Pending/Running ones. With 138 accumulated tasks and only 7 active, routing decisions were based on historical distribution, causing one node to appear permanently 'busier' and starving the other of work.


rltakashige (approved):
Looks good, thanks for catching this!

PR #1991, opened 2026-04-26
fix: map presence_penalty and frequency_penalty from ChatCompletionRequest

adurham:
Upstream PR #1947 added `presence_penalty` and `frequency_penalty` to `TextGenerationTaskParams` and the mlx-lm generator call sites, but missed wiring them up in the API adapter so they were silently dropped from incoming requests. This fixes the API mapping.


Evanev7 (approved):
huge

PR #1993, opened 2026-04-27
MLX P/D

rltakashige:
## Motivation

MLX only prefill server for Apple Silicon


Evanev7 (approved):
eh its alright i guess

PR #2015, opened 2026-05-02
Add WarnExtraModel: log warnings for unknown fields in API request models

imapotato123:
# Warn on dropped extra fields in API request models

## Problem

Pydantic v2 `BaseModel` defaults to `extra="ignore"`, silently accepting
and dropping unknown JSON fields on request bodies. For internal data
structures this is fine, but for public API request bodies it makes
mistyped or out-of-spec fields invisible.

We hit this on a 2x Mac Studio M4 Max cluster (TB5 RDMA, MLX-Jaccl
backend) when validating `/place_instance`: a mistyped `instance_meta`
field was silently dropped, falling back to the default
`Pipeline + MlxRing` placement instead of the intended
`Tensor + MlxJaccl`. The request returned 200 OK, placement looked
successful, and the only symptom was ~1.5x lower decode throughput.
There was no log line, no warning, no 4xx — debugging required diffing
the placement response against the request payload by hand.

This is the kind of footgun where the API silently does the wrong
thing. A 4xx would be the right end state, but a hard switch to
`extra="forbid"` is breaking for any client today that's relying on
extras being dropped (including future-spec fields the server simply
hasn't learned about yet).

## Change

Adds a `WarnExtraModel` base class (in `src/exo/utils/extra_fields_warner.py`)
that keeps `extra="ignore"` semantics — so behavior is unchanged — but
attaches a `model_validator(mode="before")` that compares the input
dict's keys against the model's declared fields (and aliases / alias
choices / alias paths) and logs a one-line warning per unknown key.

Warnings are rate-limited per `(model_name, field_name)` pair to once
per 60s. A burst of 100 mistyped requests produces one log line per
distinct field, not one hundred.

Then applies `WarnExtraModel` as the base class for the public API
request models. Existing `BaseModel(frozen=True)` and inherited
classes are preserved exactly — only the base swaps in the validator.

## Models updated

- `src/exo/api/types/api.py`:
  `ChatCompletionRequest`, `BenchChatCompletionRequest` (transitively),
  `PlaceInstanceParams`, `CreateInstanceParams`, `AddCustomModelParams`,
  `DeleteInstanceTaskParams`, `InstanceLinkBody`,
  `ImageGenerationTaskParams`, `BenchImageGenerationTaskParams`
  (transitively), `ImageEditsTaskParams`
- `src/exo/api/types/openai_responses.py`: `ResponsesRequest`
- `src/exo/api/types/ollama_api.py`: `OllamaChatRequest`,
  `OllamaGenerateRequest`, `OllamaShowRequest`
- `src/exo/api/types/claude_api.py`: `ClaudeMessagesRequest`

`FrozenModel` descendants (e.g. `StartDownloadParams`,
`CancelDownloadParams`, `DeleteTracesRequest`) already use
`extra="forbid"` and are intentionally untouched — they 4xx today and
should keep doing so.

## Test plan

- [x] Existing `src/exo/utils/tests/` (30 tests) pass unchanged
- [x] Existing `src/exo/api/tests/` (42 tests) pass unchanged
- [x] New tests in `src/exo/utils/tests/test_extra_fields_warner.py`
  (10 cases) cover:
  - Known fields parse without warning
  - Unknown field is dropped + warned, request still parses
  - Repeat unknown field on same model is rate-limited
  - Distinct unknown fields each warn once
  - Aliased fields (`Field(alias=...)`) are recognized
  - `AliasChoices` accepts any declared name without warning
  - `AliasPath` recognises the top-level path entry
  - Subclasses inherit the validator + warning names the subclass
  - Non-dict inputs (model instances) pass through cleanly
  - Plain `BaseModel` subclasses are unaffected (mixin is opt-in)
- [x] End-to-end repro: mistyped `instanceMeta` on `PlaceInstanceParams`
  now logs:
  ```
  WARNING | exo.utils.extra_fields_warner:_warn_unknown_fields - Dropping unknown field 'instanceMeta' on PlaceInstanceParams request — this field is not declared on the model and will be ignored. Check for a typo or an out-of-spec field. (Subsequent occurrences of this same field on this model are rate-limited to once per 60s.)
  ```
  while still placing on the default backend (existing fallback
  behavior preserved).

## Migration notes

- **Non-breaking.** Clients sending unknown fields still get 200 OK
  and the same default behavior. Only the log gains a warning.
- The intent is to flip `extra="forbid"` (or move to a stricter base
  class) in a follow-up PR after a deprecation window long enough for
  the warning to surface miswired clients in operator logs.
- The rate-limit cache is process-local (a module-level dict guarded
  by a threading lock). Multi-worker deployments will independently
  rate-limit per-process; this is the desired behavior for log-noise
  control.

## Repro

```bash
# Trigger the warning by sending a typo on /place_instance:
curl -sS -X POST http://<exo-host>:52415/place_instance \
  -d '{"model_id":"qwen-3.6-35b","sharding":"Tensor","instanceMeta":"MlxJaccl"}'

# Check logs:
tail -f ~/.local/share/exo/exo.log | grep "Dropping unknown field"
```

### DIFF
--- a/src/exo/api/types/api.py
+++ b/src/exo/api/types/api.py
@@ -11,6 +11,7 @@ from exo.shared.types.memory import Memory
 from exo.shared.types.text_generation import ReasoningDialect, ReasoningEffort
 from exo.shared.types.worker.instances import Instance, InstanceId, InstanceMeta
 from exo.shared.types.worker.shards import Sharding, ShardMetadata
+from exo.utils.extra_fields_warner import WarnExtraModel
 from exo.utils.pydantic_ext import FrozenModel
 
 FinishReason = Literal[
@@ -204,7 +205,7 @@ class StreamOptions(BaseModel):
     include_usage: bool = False
 
 
-class ChatCompletionRequest(BaseModel):
+class ChatCompletionRequest(WarnExtraModel):
     model: ModelId
     frequency_penalty: float | None = None
     messages: list[ChatCompletionMessage]
@@ -237,7 +238,7 @@ class BenchChatCompletionRequest(ChatCompletionRequest):
     use_prefix_cache: bool = False
 
 
-class AddCustomModelParams(BaseModel):
+class AddCustomModelParams(WarnExtraModel):
     model_id: ModelId
 
 
@@ -250,14 +251,14 @@ class HuggingFaceSearchResult(BaseModel):
     tags: list[str] = Field(default_factory=list)
 
 
-class PlaceInstanceParams(BaseModel):
+class PlaceInstanceParams(WarnExtraModel):
     model_id: ModelId
     sharding: Sharding = Sharding.Pipeline
     instance_meta: InstanceMeta = InstanceMeta.MlxRing
     min_nodes: int = 1
 
 
-class CreateInstanceParams(BaseModel):
+class CreateInstanceParams(WarnExtraModel):
     instance: Instance
 
 
@@ -275,7 +276,7 @@ class PlacementPreviewResponse(BaseModel):
     previews: list[PlacementPreview]
 
 
-class DeleteInstanceTaskParams(BaseModel):
+class DeleteInstanceTaskParams(WarnExtraModel):
     instance_id: str
 
 
@@ -296,7 +297,7 @@ class CancelCommandResponse(BaseModel):
     command_id: CommandId
 
 
-class InstanceLinkBody(BaseModel):
+class InstanceLinkBody(WarnExtraModel):
     prefill_instances: list[InstanceId]
     decode_instances: list[InstanceId]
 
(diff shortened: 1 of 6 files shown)


imapotato123:
Refreshed this onto current `main` — the branch was 31 commits behind, now up to date and mergeable with no conflicts.

Still applies as written. The change is additive: `WarnExtraModel` logs a warning when a public API request model receives unknown fields, rather than silently dropping them the way Pydantic v2's `extra="ignore"` default does. Nothing changes for internal models, and no request that worked before starts failing — it only makes an existing silent failure visible.

Scope is `src/exo/utils/extra_fields_warner.py` plus the four API request-type modules that opt into it, with 179 lines of tests.

Would appreciate a look when someone has time. Happy to rebase again or narrow the scope if the maintainers would rather this start with a single endpoint.

PR #2026, opened 2026-05-03
Fix reliable Darwin mDNS discovery

AlexCheema:
## Root Cause

There were two separate issues behind the unreliable no-bootstrap startup on the four-machine macOS cluster.

1. **Darwin mDNS advertisement was not reliable from the background exo process.**

   On the affected machines, the normal libp2p mDNS path could fail to advertise in the daemonized/background context. In particular, multicast / limited broadcast paths were not reliable, while the interface's directed subnet broadcast address worked consistently.

2. **Discovery alone was not enough because election transitions could stall.**

   After peers connected and exchanged election traffic, a node could get stuck while demoting/rebuilding because `Worker.shutdown()` could wait indefinitely. That left the node connected at the libp2p layer but without reliably rebuilding worker/API/event-router state, so `/state` could remain incomplete.

## What Changed and Why

- Added a Darwin-only mDNS announcer in Python.

  `src/exo/routing/mdns_announcer.py` builds a libp2p-mdns-compatible DNS response containing:

  ```text
  dnsaddr=/ip4/<node-ip>/tcp/<libp2p-port>/p2p/<node-id>
  ```

  `src/exo/main.py` starts this announcer on macOS when exo has a fixed libp2p port. It discovers the `en0` address and subnet mask, computes the directed broadcast address, and advertises there first. This is what made no-bootstrap peer discovery reliable on the test machines.

- Kept the announcer active through startup.

  The announcer sends every second for the first 60 sends, then backs off to every 10 seconds. This covers cold starts where machines come up at slightly different times without relying on bootstrap peers.

- Made worker shutdown bounded and proactive.

  `src/exo/worker/main.py` now closes senders, shuts down runners, cancels background tasks, and waits for shutdown with a timeout. This prevents election demotion from blocking forever and lets the node rebuild its runtime state after topology/election changes.

## What This Is Not

- This PR does **not** vendor or patch Rust/libp2p code.
- An earlier version tried a Rust `libp2p-mdns` patch, but that was removed after verification showed the Python announcer plus bounded worker shutdown was sufficient.

## Validation

Local checks:

- `uv run basedpyright src/exo/main.py src/exo/routing/mdns_announcer.py src/exo/worker/main.py`
- `uv run ruff check src/exo/main.py src/exo/routing/mdns_announcer.py src/exo/worker/main.py`
- `uv run pytest`
- `uv run --reinstall-package exo-pyo3-bindings python -c "import exo_pyo3_bindings; print('rebuilt bindings')"`

Cluster verification:

- Final slim Python-only PR, no Rust vendoring: **10/10 cold-start cycles passed** across `james`, `mike`, `s13`, and `s14`.
- Each cycle killed previous listeners, started all four machines from scratch with `EXO_BOOTSTRAP_PEERS` unset, and verified every node's `/state` reported `topology.nodes == 4`.
- The final test used commit `49b2c5f2` from `alexcheema/reliable-mdns-discovery`.

## Notes

- PR is stacked on `alexcheema/profilers-dashboard` so the diff contains only this mDNS reliability fix.

### DIFF
--- a/src/exo/main.py
+++ b/src/exo/main.py
@@ -1,8 +1,11 @@
 import argparse
+import ipaddress
 import multiprocessing as mp
 import os
 import resource
 import signal
+import subprocess
+import sys
 from dataclasses import dataclass, field
 from typing import Self
 
@@ -41,6 +44,7 @@ class Node:
     node_id: NodeId
     offline: bool
     _api_port: int
+    _libp2p_port: int
     _tg: TaskGroup = field(init=False, default_factory=TaskGroup)
 
     @classmethod
@@ -141,6 +145,7 @@ class Node:
             node_id,
             args.offline,
             args.api_port,
+            args.libp2p_port,
         )
 
     async def run(self):
@@ -158,6 +163,12 @@ class Node:
                 tg.start_soon(self.master.run)
             if self.api:
                 tg.start_soon(self.api.run)
+            if sys.platform == "darwin" and self._libp2p_port != 0:
+                tg.start_soon(
+                    _darwin_mdns_broadcast_announcer,
+                    self.node_id,
+                    self._libp2p_port,
+                )
             tg.start_soon(self._elect_loop)
 
     def shutdown(self):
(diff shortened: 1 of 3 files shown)


Evanev7:
i will be moving us off mdns; closing as stale

PR #2037, opened 2026-05-03
Retry custom card reconciliation failures

AlexCheema:
## Why

Custom model cards are now reconciled from durable `State`, which is the right event-sourcing shape. That reconciliation loop should be robust to transient filesystem failures: one failed write/delete should not kill the loop and leave the worker permanently out of sync with state.

## How

- Catch `OSError` while writing a custom card to disk.
- Catch `OSError` while deleting a stale custom card from disk.
- Leave `_synced_custom_cards` unchanged on failure so the next reconciliation tick retries.
- Add focused tests for save and delete failure retry behavior.

## Tests

- `uv run pytest src/exo/worker/tests/unittests/test_worker_custom_cards.py`
- `uv run pytest`
- `uv run basedpyright`
- `uv run ruff check`
- `nix fmt`

### DIFF
--- a/src/exo/worker/main.py
+++ b/src/exo/worker/main.py
@@ -226,15 +226,25 @@ class Worker:
         for model_id, card in target.items():
             if self._synced_custom_cards.get(model_id) == card:
                 continue
-            await card.save_to_custom_dir()
-            add_to_card_cache(card)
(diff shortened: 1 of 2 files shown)


Evanev7:
manually merged into #2024

PR #2039, opened 2026-05-03
Improve prefill decode benchmark reporting

AlexCheema:
## Why

The prefill/decode benchmark was losing useful metadata from benchmark responses and could select stale node identities from previous cluster sessions. Local testing also showed that instance readiness can precede the prefill-server socket becoming reachable, but that wait needs to be opt-in rather than hardcoded to one cluster IP.

## How

- Filter node identities to currently-live topology nodes when resolving friendly names.
- Preserve `power_usage` from benchmark chat completion responses in per-run rows.
- Add optional `--prefill-server-host` / `prefill_server_host` TOML support to poll the prefill server port after linking.
- Keep the default behavior as the existing fixed 2s wait when no host is configured.

## Tests

- `uv run ruff check bench/prefill_decode_bench.py`
- `uv run basedpyright`
- `uv run python bench/prefill_decode_bench.py --help`
- `nix fmt`

### DIFF
--- a/bench/prefill_decode_bench.py
+++ b/bench/prefill_decode_bench.py
@@ -22,6 +22,7 @@ import contextlib
 import copy
 import itertools
 import json
+import socket
 import sys
 import time
 import tomllib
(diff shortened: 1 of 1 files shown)


AlexCheema:
Closing because this prefill/decode benchmark workaround belongs out of the public repo, and the real fix should be in exo-nvidia readiness semantics rather than benchmark-side waiting.

PR #2040, opened 2026-05-03
Add rambo prefill decode bench config

AlexCheema:
## Why

The rambo/gx10 split prefill benchmark setup is useful to keep as a checked-in, reproducible config instead of a local-only file. This is stacked on #2039 because it uses the new `prefill_server_host` and `prefill_server_wait_timeout` config keys added there.

## How

- Adds `bench/prefill-decode-rambo.toml` for the rambo decode node and gx10 prefill node.
- Pins the prefill server host to the Thunderbolt/local link address and gives startup polling a bounded timeout.

## Testing

- `nix fmt`
- `uv run python -c 'import tomllib; tomllib.load(open("bench/prefill-decode-rambo.toml", "rb"))'`

### DIFF
--- a/bench/prefill-decode-rambo.toml
+++ b/bench/prefill-decode-rambo.toml
@@ -1,0 +1,21 @@
+port = 52415
+timeout = 72000.0
+settle_timeout = 60.0
+prefill_server_host = "169.254.100.1"
+prefill_server_wait_timeout = 60.0
+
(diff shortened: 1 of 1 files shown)


AlexCheema:
Closing this because the rambo/gx10 config is local test-only and should stay out of the PR stack.

PR #2051, opened 2026-05-05
Initialise _cancelled_tasks in ImageEngine

Evanev7:
we yielded nonsense chunks from engines; we didn't initialize the image engine correctly. mostly rewrite of #2049


ciaranbor (approved):
Makes sense

PR #2056, opened 2026-05-06
Andrei/mp capture stdio

AndreiCravtov:
## Motivation

Process-isolated runner crashes and C-extension failures can write directly to fd-level stdout/stderr, bypassing Python/loguru. We need to capture that output per runner process without polluting the main process or other workers, and without breaking operation when the parent stdio is detached.

## Changes

- Added `AsyncProcess`, a spawn-only multiprocessing wrapper that redirects child stdout/stderr to pipes and exposes them as in-memory `Receiver[bytes]`s
- Replaced runner-supervisor's raw `multiprocessing.Process` usage with `AsyncProcess`
- Added `--no-stdio`, redirecting stdin/stdout/stderr to `/dev/null` after logging is configured
- Disabled verbose MLX
- Added tests covering stdio capture, child crashes, repeated bad children, SIGTERM/SIGKILL shutdown escalation, stdio detachment, and spawning captured children from a stdio-detached parent

## Why It Works

The parent can redirect its own stdio fds to `/dev/null`, while `AsyncProcess` installs fresh pipe fds over fd 1 and 2 inside each spawned child. That keeps stdio-detached parents quiet while preserving per-runner stdout/stderr capture. Runner shutdown is still bounded: SIGTERM grace first, then SIGKILL escalation if needed.

Next direction: the runner supervisor currently drains captured output and logs it as stdout/debug and stderr/warning. This should be split into more useful process-isolated error reporting instead of just log forwarding (regex match on errors to obtain "reason" string, best effort).

### Manual Testing

Ran on 4 Mac Minis in a Thunderbolt 4 ring, can see that runner's stdout/stderr contents are being captured.

### Automated Testing

- Added async-process tests for fd-level stdout/stderr capture, Python traceback capture, bounded-buffer output, child `exit`/abort, parent stdio preservation, fd leak checks, spawn-context mp channels, and SIGTERM/SIGKILL shutdown behavior
- Added stdio-detach tests proving stdio detaches to `/dev/null`, a stdio-detached parent can still spawn and capture a child, and the same stdio-detached parent can spawn/capture multiple children sequentially
- Updated runner-supervisor tests for the new `AsyncProcess.exitcode` path


Evanev7:
1000 lines? clean this up man... mostly the async_process - implement the functionality you need rather than reimplementing all of mp.Process.

src/exo/utils/async_process.py
```diff
@@ -0,0 +1,355 @@
...
+        self._buffer.clear()
+        await self._receive_stream.aclose()
+
+
+@final
+class AsyncSpawnProcess:
+    @staticmethod
+    def context() -> SpawnContext:
+        return mp.get_context("spawn")
+
```
Evanev7:
```suggestion
```

src/exo/utils/async_process.py
```diff
@@ -0,0 +1,355 @@
...
+        target: Callable[..., object] | None = None,
+        name: str | None = None,
+        args: Iterable[object] = (),
+        kwargs: Mapping[str, object] | None = None,
+        *,
+        daemon: bool | None = None,
+        stream_buffer_size: int = 16,
+    ) -> None:
+        if stream_buffer_size <= 0:
+            raise ValueError("stream_buffer_size must be positive")
```
Evanev7: make const

src/exo/utils/async_process.py
```diff
@@ -0,0 +1,355 @@
...
+                    await self._terminate_if_still_alive()
+            finally:
+                self._has_stopped = True
+                self._stopped.set()
+
+    async def wait_started(self) -> None:
+        await self._started.wait()
+        if self._start_error is not None:
+            raise self._start_error
+
```
Evanev7:
```suggestion
```

src/exo/utils/async_process.py
```diff
@@ -0,0 +1,355 @@
...
+_STDOUT_FD = 1
+_STDERR_FD = 2
+_READ_CHUNK_SIZE = 64 * 1024
+_TERMINATE_GRACE_SECONDS = 5.0
+_KILL_GRACE_SECONDS = 5.0
+
+
+@final
+class MemoryByteReceiveStream(ByteReceiveStream):
```
Evanev7: this is a bizarre construct that reads structured objects of bytes into a buffer and then returns them - expose the byte receive stream, or process the byte streams into something more useful. you can wrap this in a bufferedbytereceivestream and then a textreceivestream to achieve a lines type reader.

src/exo/utils/channels.py
```diff
@@ -79,16 +80,16 @@ class _MpEndOfStream:
 
 
 class MpState[T]:
-    def __init__(self, max_buffer_size: float):
+    def __init__(self, max_buffer_size: float, mp_ctx: BaseContext):
```
Evanev7: please dont

src/exo/worker/runner/supervisor.py
```diff
@@ -129,41 +151,13 @@ async def run(self):
...
-                logger.info("Runner process succesfully terminated")
-
-            self.runner_process.close()
+            if runner_started:
+                with anyio.CancelScope(shield=True):
+                    self.runner_process.shutdown()
+                    await self.runner_process.wait_stopped()
+                    if not self.runner_process.is_alive():
+                        logger.info("Runner process succesfully terminated")
+                    self.runner_process.close()
```
Evanev7: move all of this code into the process shutdown or none of it. this is a significant change in behaviour and as far as i can tell there isn't any replication of the repeat interrupt logic - it was there for a reason.

Evanev7 (changes requested):

src/exo/utils/async_process.py
AndreiCravtov: Ill just make inf (thats default for channels anyways)

src/exo/utils/channels.py
AndreiCravtov: Gone

AndreiCravtov:
> 1000 lines? clean this up man... mostly the async_process - implement the functionality you need rather than reimplementing all of mp.Process.

tbf most of the 1k is test-code, the actual meat is like 250 ish lines

src/exo/utils/async_process.py
```diff
@@ -0,0 +1,291 @@
...
+                self._start_error = exc
+                self._started.set()
+            cleanup_stdio_fd()
+            raise
+        finally:
+            try:
+                with CancelScope(shield=True):
+                    await self._terminate_if_still_alive()
+            finally:
+                cleanup_stdio_fd()
```
Evanev7: runs cleanup_stdio_fd twice btw.

Evanev7 (approved):
fix the ci but this'll do

src/exo/utils/async_process.py
AndreiCravtov: true lol. removed it. 

for some reason I didnt believe that python's final blocks were truly final

PR #2059, opened 2026-05-06
Stabilize master election and event replay

team-wcv:
## Motivation

We hit two distinct master/replay correctness bugs running exo across a 4-5 node Apple Silicon + DGX cluster on Thunderbolt RDMA. Both are small, surgical fixes with regression tests:

1. **Election flapping when a forced-master message is delayed.** A connection hiccup triggered a new election; the forced-master node's message arrived after the campaign timeout, which caused a brief master flip that killed running instances and desynchronized the event log.
2. **Event-replay split-brain.** Old master session logs were replayed during cold start, so previously-deleted instances came back, and runners assigned to those instances lingered in state with no owner. After enough sessions, the master also stalled replaying a giant cumulative log.

### 1. Track max observed seniority across rounds (`src/exo/shared/election.py`)
- When a campaign resolves with a winner whose seniority is lower than any seniority observed across earlier rounds, reject the result and keep the current master.
- Adds 70 lines of regression tests covering the delayed-forced-master scenario.

### 2. Cascade runner cleanup on instance deletion (`src/exo/shared/apply.py`)
- `apply_instance_deleted` now also prunes runners whose `runner_id` is in the deleted instance's `shard_assignments.runner_to_shard`. This change is layered on top of the existing `instance_links` cleanup so both invariants hold.
- Companion test in `test_apply_runner_deleted.py` asserting both instance and runner pruning.

### 3. Scope master event logs by session (`src/exo/master/main.py`)
- Each master session writes its event log under a session-specific directory, and old session log directories are pruned at startup (default cap = 10).
- Prevents replay of stale sessions resurrecting deleted instances/runners and prevents the master from getting stuck replaying ever-growing cumulative logs.

### 4. Misc fixes that fall out of (1)-(3)
- `master/placement.py`: defensive handling for runners no longer in state.
- `download/coordinator.py`: don't lose download status when we receive a stale event after a master flip.
- Dashboard: clear `runners`/`instances` when the master change-id changes so the UI doesn't show ghost rows.

## Why It Works

Election seniority is monotonic per node by design; if a campaign result reports a winner with seniority lower than what some peer has already broadcast, that result is provably stale. Rejecting it keeps the existing (higher-seniority) master, which is the only safe action — the in-flight forced-master message will arrive shortly and resolve the round.

Cascading runner cleanup on instance deletion is just enforcing the existing invariant that a runner's lifetime is owned by its instance. Without this, a delete event leaves dangling runners that subsequent placement code has to special-case.

Scoping event logs per session is the standard fix for "replay may resurrect deleted state" — if an old session's events are not on the replay path, they can't resurrect anything.

### Automated Testing

- Adds 70 lines to \`test_election.py\` covering: delayed forced-master, cross-round seniority tracking, multi-round resolution.
- Adds 60 lines to \`test_apply_runner_deleted.py\` covering: instance deletion cascade-cleans runners; runners preserved when no instance match.
- Adds 90 lines to \`test_master.py\` covering: session log directory rotation, capped session count, replay scoping.
- All existing tests in touched modules still pass.

\`\`\`
src/exo/shared/tests/test_apply/test_apply_runner_deleted.py ...
src/exo/shared/tests/test_election.py ......................
src/exo/master/tests/test_master.py ...
src/exo/master/tests/test_placement.py ..................
src/exo/download/tests/test_download_status_not_lost.py ...
=== 56 passed in 9.18s ===
\`\`\`

### Manual Testing

Hardware: 4-node Apple Silicon cluster (3x M5 Max + 1x M1 Ultra) over Thunderbolt 4/5 RDMA + LAN.

- Reproduced the original flap by killing the forced-master node, watching it rejoin late, and confirming the cluster no longer flips master mid-rejoin.
- Reproduced the replay split-brain by starting cold with stale session logs in place and confirming the master no longer resurrects deleted instances/runners.

## Notes for reviewers

- One conflict at `apply_instance_deleted` was hand-resolved against current upstream main — preserving the existing `instance_links` cleanup and layering in the new `runners` cleanup. Both branches of the if/else return the same shape with both new keys included. Diff for that resolution is small and readable in the second commit.
- The third commit references "(#8)" — that's our internal fork PR number, not an upstream one. Happy to amend the commit message if upstream prefers.

### DIFF
--- a/dashboard/src/lib/stores/app.svelte.ts
+++ b/dashboard/src/lib/stores/app.svelte.ts
@@ -1455,10 +1455,10 @@ class AppStore {
   /**
    * Clear the preview node filter and re-fetch placements
    */
-  clearPreviewNodeFilter() {
+  clearPreviewNodeFilter(refetch = true) {
     this.previewNodeFilter = new Set();
     // Re-fetch with no filter if we have a selected model
-    if (this.selectedPreviewModelId) {
+    if (refetch && this.selectedPreviewModelId) {
       this.fetchPlacementPreviews(this.selectedPreviewModelId, false);
     }
   }
@@ -3543,7 +3543,8 @@ export const selectPreviewModel = (modelId: string | null) =>
   appStore.selectPreviewModel(modelId);
 export const togglePreviewNodeFilter = (nodeId: string) =>
   appStore.togglePreviewNodeFilter(nodeId);
-export const clearPreviewNodeFilter = () => appStore.clearPreviewNodeFilter();
+export const clearPreviewNodeFilter = (refetch = true) =>
+  appStore.clearPreviewNodeFilter(refetch);
 export const previewNodeFilter = () => appStore.previewNodeFilter;
 export const deleteMessage = (messageId: string) =>
   appStore.deleteMessage(messageId);
--- a/dashboard/src/routes/+page.svelte
+++ b/dashboard/src/routes/+page.svelte
@@ -889,7 +889,7 @@
   type InstanceMeta = "MlxRing" | "MlxJaccl";
 
   // Launch defaults persistence
-  const LAUNCH_DEFAULTS_KEY = "exo-launch-defaults-v2";
+  const LAUNCH_DEFAULTS_KEY = "exo-launch-defaults-v3";
   interface LaunchDefaults {
     modelId: string | null;
     sharding: "Pipeline" | "Tensor";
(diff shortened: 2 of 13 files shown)


team-wcv:
Consolidated into #2080 (Cluster reliability: master election stability + bootstrap peer dial-in + local model symlink fix), which combines this PR's election + event-replay fixes with #2060's bootstrap and local-model fixes into a single review unit.

PR #2060, opened 2026-05-06
Dial bootstrap peers on swarm startup; fix symlinked local model deletion

team-wcv:
## Motivation

Two small, independent fixes that surfaced running exo with bootstrap peers behind LAN/Tailscale and with locally-symlinked model directories:

1. **Bootstrap peers were not actively dialed at swarm startup**, so a node that came up before its bootstrap peer was reachable would sit idle until libp2p's gossip discovery happened to find a path. On clusters where the master is the only well-known node, this manifested as workers staying disconnected for tens of seconds.
2. **Deleting a model whose `~/.exo/models/<owner>--<model>` is a symlink** (e.g. into LM Studio's model store) recursively walked through the symlink and tried to delete the *target's* contents, which either failed with permission errors or, in the worst case, corrupted the LM Studio cache.

### 1. `rust/networking/src/discovery.rs` — proactively dial bootstrap peers on startup
- After registering bootstrap multiaddrs, immediately issue an outbound dial to each rather than waiting for gossip-driven discovery.
- Idempotent: libp2p de-duplicates concurrent dials to the same peer ID.

### 2. `src/exo/download/download_utils.py` + `coordinator.py` — symlink-aware model deletion
- Detect when a model directory is a symlink before recursing.
- For symlinks: unlink the symlink itself and leave the target untouched.
- For regular dirs: existing recursive delete behavior is unchanged.
- New regression test in `test_model_dirs.py` covers both branches.

## Why It Works

For (1), libp2p gossip-based discovery is fundamentally pull-driven — a peer learns about another peer when someone gossips about it. With only a single bootstrap peer, that gossip may take a long time or never happen if topology is fragmented. Explicit dial pushes the dial state machine forward immediately and is the same pattern used by libp2p's own bootstrap example.

For (2), the existing code path used `shutil.rmtree(path)` which follows symlinks by default on macOS. Switching to `path.is_symlink()` + `path.unlink()` gives correct semantics for both file and directory symlinks, and `shutil.rmtree(path, ignore_errors=False)` only runs on real directories.

### Automated Testing

- New unit tests in `src/exo/download/tests/test_model_dirs.py` covering symlink and non-symlink delete paths (20 tests pass).
- Existing download/coordinator tests still pass:

\`\`\`
src/exo/download/tests/test_cancel_download.py ...
src/exo/download/tests/test_download_status_not_lost.py ...
src/exo/download/tests/test_download_verification.py ...........
src/exo/download/tests/test_model_dirs.py ....................
src/exo/download/tests/test_offline_mode.py ..........
src/exo/download/tests/test_rate_limit_handling.py ...................
src/exo/download/tests/test_re_download.py .
src/exo/download/tests/test_safetensors_index.py ...
=== 70 passed in 5.76s ===
\`\`\`

### Manual Testing

Hardware: 4-node Apple Silicon cluster (3x M5 Max + 1x M1 Ultra) over Thunderbolt 4/5 RDMA + LAN.

- Cold-started workers with master not yet ready, then started master. With this patch, workers connect within ~2s of master becoming reachable; before, observed delays of 30-60s.
- Symlinked `~/.exo/models/mlx-community--Qwen3.5-122B-A10B-4bit` to LM Studio's copy, deleted via the dashboard, confirmed the LM Studio target was untouched.

### DIFF
--- a/rust/networking/src/discovery.rs
+++ b/rust/networking/src/discovery.rs
@@ -114,19 +114,29 @@ pub struct Behaviour {
 
 impl Behaviour {
     pub fn new(keypair: &identity::Keypair, bootstrap_peers: Vec<Multiaddr>) -> io::Result<Self> {
-        Ok(Self {
+        let mut behaviour = Self {
             managed: managed::Behaviour::new(keypair)?,
             mdns_discovered: HashMap::new(),
             bootstrap_peers,
             retry_delay: Delay::new(RETRY_CONNECT_INTERVAL),
             pending_events: WakerDeque::new(),
-        })
+        };
+        behaviour.dial_bootstrap_peers();
+        Ok(behaviour)
     }
 
     fn dial(&mut self, peer_id: PeerId, addr: Multiaddr) {
         self.pending_events.push_back(ToSwarm::Dial {
             opts: DialOpts::peer_id(peer_id).addresses(vec![addr]).build(),
         })
+    }
+
+    fn dial_bootstrap_peers(&mut self) {
+        for addr in &self.bootstrap_peers {
+            self.pending_events.push_back(ToSwarm::Dial {
+                opts: DialOpts::unknown_peer_id().address(addr.clone()).build(),
+            })
+        }
     }
 
     fn close_connection(&mut self, peer_id: PeerId, connection: ConnectionId) {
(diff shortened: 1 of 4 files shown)


team-wcv:
Consolidated into #2080 (Cluster reliability: master election stability + bootstrap peer dial-in + local model symlink fix), which combines this PR's bootstrap + local-model fixes with #2059's election + event-replay fixes into a single review unit.

PR #2072, opened 2026-05-08
Create PID file locking for EXO

AndreiCravtov:
## Motivation

EXO should be PID file locked, to prevent duplicate processes from clobbering the log, right now this isn't the case.

## Changes

I added a wrapper around a Rust PID file lock library, and used it to implement PID locking for EXO, with the PID file being in exo cache directory.

### Manual Testing
Tested on e11, trying to spawn duplicate EXO processes prevented.

### DIFF
--- a/rust/exo_pyo3_bindings/Cargo.toml
+++ b/rust/exo_pyo3_bindings/Cargo.toml
@@ -46,9 +46,12 @@ pyo3-async-runtimes = { version = "0.27.0", features = [
 ] }
 pyo3-log = "0.13.2"
 
+pidfile-rs = "0.3"
+
 # macro dependencies
 extend = { workspace = true }
 delegate = { workspace = true }
+thiserror = "2.0"
 
 # async runtime
 tokio = { workspace = true, features = ["full", "tracing"] }
--- a/rust/exo_pyo3_bindings/src/pidfile.rs
+++ b/rust/exo_pyo3_bindings/src/pidfile.rs
@@ -1,0 +1,87 @@
+use pidfile_rs::{Pidfile, PidfileError};
+use pyo3::exceptions::PyException;
+use pyo3::prelude::{PyModule, PyModuleMethods};
+use pyo3::{Bound, PyErr, PyResult, Python, pyclass, pymethods};
+use pyo3_stub_gen::derive::{gen_stub_pyclass, gen_stub_pymethods};
+use std::fs::Permissions;
+use std::os::unix::prelude::PermissionsExt;
+use std::path::PathBuf;
+
+#[gen_stub_pyclass]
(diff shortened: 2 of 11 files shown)


rust/exo_pyo3_bindings/src/pidfile.rs
```diff
@@ -0,0 +1,87 @@
...
+                .map_err(|e| PyPidfileError(e).into_pyerr(py))?,
+        ))
+    }
+
+    /// Writes the current process ID to the PID file.
+    ///
+    /// The file is truncated before writing.
+    fn write<'py>(&mut self, py: Python<'py>) -> PyResult<()> {
+        self.0.write().map_err(|e| PyPidfileError(e).into_pyerr(py))
+    }
```
Evanev7, line 79: why is this not called immediately after the constructor?

src/exo/utils/tests/test_pidfile.py
```diff
@@ -0,0 +1,84 @@
...
+import textwrap
+from pathlib import Path
+from typing import Final
+
+import pytest
+
+import exo.utils.pidfile as pidfile
+from exo.utils.pidfile import acquire_exo_pidfile
+
+_CHILD_ACQUIRE_PIDFILE_SCRIPT: Final = textwrap.dedent(
```
Evanev7, line 16: textwrap? codex inventing new python stdlib modules tonight

src/exo/utils/pidfile.py
```diff
@@ -0,0 +1,26 @@
...
+    path = EXO_PID_FILE
+    try:
+        pidfile = Pidfile(path, _PIDFILE_MODE)
+        pidfile.write()
+    except (OSError, PidfileError) as exception:
+        raise PidfileLockError(
+            f"Failed to acquire EXO pidfile at {path}: {exception}"
+        ) from exception
+
+    return pidfile
```
Evanev7, line 26: instead of having this file - just set the pidfile mode to 0o600 in the rust code. I don't see the benefit in this wrapper.

rltakashige (approved):
:) Thanks, reviewed in person

rust/exo_pyo3_bindings/Cargo.toml
```diff
@@ -46,9 +46,12 @@ pyo3-async-runtimes = { version = "0.27.0", features = [
 ] }
 pyo3-log = "0.13.2"
 
+pidfile-rs = "0.3"
+
 # macro dependencies
 extend = { workspace = true }
 delegate = { workspace = true }
+thiserror = "2.0"
```
Evanev7, line 54: not used?

src/exo/utils/pidfile.py
Evanev7: especially the double wrapped exceptions. just propagate the exception you want.

PR #2076, opened 2026-05-09
libp2p -> zenoh pure

Evanev7:
same impl as #2073 but purely the zenoh to libp2p changes.
offers no benefit over libp2p but should make the other PR more comprehensible.


Evanev7:
alright, ready to merge.

PR #2084, opened 2026-05-11
Redirect runner stdout/stderr to file logs 

AndreiCravtov:
## Motivation

We want to use log mining tools like [Drain3](https://github.com/logpai/Drain3) to get standardized error formats, but for that we should record runner stdout/stderr in a massive append-only log to gather training data for such tools. Also useful for future opt-in telemetry.

## Changes

The stdout/stderr from runner now splits into 3 tasks: 
1) raw write to dedicated runner logs 
2) sanitized line-by-line logging with log-guru 
3) stub for further error-processing (i.e. turning lines into errors)

### Manual Testing
Works on 4x mac mini clusted connected as TB4 ring.

### DIFF
--- a/src/exo/worker/runner/supervisor.py
+++ b/src/exo/worker/runner/supervisor.py
@@ -1,17 +1,17 @@
+import codecs
 import contextlib
 import signal
 from dataclasses import dataclass, field
 from os import PathLike
-from typing import Self, Callable
+from typing import Callable, Self
 
 import anyio
 from anyio import (
+    AsyncFile,
     BrokenResourceError,
     ClosedResourceError,
     EndOfStream,
-    AsyncFile,
 )
-from anyio.streams.text import TextReceiveStream
 from loguru import logger
 
 from exo.shared.constants import EXO_RUNNER_STDERR_LOG, EXO_RUNNER_STDOUT_LOG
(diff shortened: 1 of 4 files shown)


src/exo/worker/runner/supervisor.py
```diff
@@ -42,18 +46,139 @@
...
+                tg.start_soon(  # pyright: ignore[reportUnknownArgumentType]
+                    self._handle_runner_output,
+                    self._stderr_rx,
+                    self._stderr_log,
+                    lambda line: logger.warning(f"Runner stderr: {line}"),  # pyright: ignore[reportUnknownLambdaType]
+                )
+        finally:
+            # close up files
+            await self._stdout_log.aclose()
+            await self._stderr_log.aclose()
```
Evanev7: you should shield async cleanup logic, the awaits will reraise cancellations.

src/exo/worker/runner/supervisor.py
```diff
@@ -42,18 +46,139 @@
...
+            await logfile.write(text)
+            await logfile.flush()
+
+            # newline buffering
+            pending_line += text
+            lines = pending_line.split("\n")
+            pending_line = lines.pop()
+
+            for line in lines:
+                await handle_line(line)
```
Evanev7, line 154: can you inline this? the nonlocal/async closure seems to be strictly unnecessary

src/exo/worker/runner/supervisor.py
```diff
@@ -42,18 +46,139 @@
...
+
+        try:
+            while True:
+                await handle_text(decoder.decode(await rx.receive(), final=False))
+        except EndOfStream:
+            await finish()
+        except (ClosedResourceError, BrokenResourceError):
+            await finish()
+            logger.warning("Runner stdio stream closed before clean EOF")
+
```
Evanev7, line 170:
```suggestion
        except EndOfStream:
            pass
        except (ClosedResourceError, BrokenResourceError):
            pass
            logger.warning("Runner stdio stream closed before clean EOF")
        finally:
        	with CancelScope(shield=True):
        		await finish()
```
assuming logfile.flush() should run on other terminating exceptions.
though again at that point inline the finish method....

Evanev7 (changes requested):
nice, just some cancellation stuff & this is good to go

AndreiCravtov:
gotcha, nws

Evanev7 (approved):
lgtm

Issue #2042, opened 2026-05-03
[BUG] Cannot use ollama API with VS Code Copilot due to incorrect version reporting


MCJack123:
## Describe the bug

The exo ollama API reports "exo v1.0" as the version, but VS Code's Copilot extension checks the version number and fails to read the extra text in the version number. This causes it to erroneously report that the server isn't compatible.

<img width="556" height="30" alt="Image" src="https://github.com/user-attachments/assets/f67f8bc0-b8d3-4bea-b4dc-673153b9aa6e" />

## To Reproduce

1. Run an EXO server
2. Run VS Code with the Copilot extension enabled
3. In the Language Models settings dialog for Copilot, click "Add Models" -> "Ollama" and add the EXO server's ollama endpoint
4. Notice that the new endpoint errors out

## Expected behavior

VS Code is able to connect to the server successfully.

## Actual behavior

VS Code detects an incompatible ollama version and rejects connection.

## Environment

I'm not sure about the EXO environment - someone else is running it for me, but it's on a 48 GB Mac mini running macOS 26.3.

VS Code version: 1.116.0
Copilot extension version: 0.45.1

## Additional context

https://github.com/exo-explore/exo/blob/9c6ff4ce95a7c2a67149f20bf8392cdf3345dab3/src/exo/api/main.py#L1710

Heidar-An:
Hey, I've fixed this in this PR - https://github.com/exo-explore/exo/pull/2091!

Evanev7:
thanks for the fix heidar!

PR #2091, opened 2026-05-13
fix ollama API compatibility for VS Code Copilot

Heidar-An:
Ollama adapter fixes for VS Code Copilot (#2042):

  - /api/version: bare semver "1.0.0" - Copilot parseInts each segment.
  - /api/show: populate model_info + capabilities - Copilot crashes on null model_info and filters by `tools`.
  - Add POST /ollama/v1/chat/completions - ollama serves the OpenAI-compat route here, BYOK clients 405 without it.

Before:
image

After:
image
image

### DIFF
--- a/src/exo/api/main.py
+++ b/src/exo/api/main.py
@@ -104,6 +104,7 @@ from exo.api.types.claude_api import (
     ClaudeMessagesResponse,
 )
 from exo.api.types.ollama_api import (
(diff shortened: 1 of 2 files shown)


Evanev7 (approved):
nice - we don't have many opportunities to test this part of our api so the contribution is much appreciated!

PR #2093, opened 2026-05-14
Runner error handling

AndreiCravtov:
# Runner error handling

## Motivation

Runner failures were mostly surfaced as plain shutdown messages, which made root cause hard to spot from API errors or runner status.

This adds a MVP path for preserving runner crash context and attaching known stderr diagnostics to failure reports.

## Changes

- Added `RunnerTerminationError` for Python exceptions raised inside runner bootstrap
- Changed runner bootstrap to send `Event | RunnerTerminationError` over the private runner channel
- Moved public `RunnerFailed` emission back into supervisor
- Added stderr-only `RunnerDiagnosticCollector`
- + Added known diagnostics for Metal GPU timeout, ring socket receive errno, and ring transport abort
- Added diagnostics to `RunnerFailed` and `ErrorChunk`
- Tweaked async process termination to join briefly before terminate/kill
- Updated tests/fixtures for new failure payload shape
- Added Ruff VS Code formatter settings

## Why It Works

Runner child now reports raw-ish failure context to supervisor instead of publishing failed status directly.

Supervisor still owns process lifecycle, exit code/signal handling, in-flight task error chunks, and final runner status. Stderr diagnostics stay best effort and only known root-cause variants are surfaced.

### Manual Testing

Hardware: remote runner logs from e16/e11/e4/e2

What you did:
- inspected live runner stderr logs
- used observed Metal GPU timeout and ring socket errors as initial diagnostic targets

### Automated Testing

- `nix flake check`
- supervisor test covers error chunk + failed status emission
- plan lifecycle test updated for failed runner diagnostics
- type/lint checks cover new runner channel union


src/exo/shared/types/chunks.py
```diff
@@ -12,6 +12,7 @@
 from exo.shared.models.model_cards import ModelId
 from exo.utils.pydantic_ext import TaggedModel
 
+from ...worker.runner.diagnostics import KnownRunnerDiagnostic
```
Evanev7: non-relative here
```suggestion
from exo.worker.runner.diagnostics import KnownRunnerDiagnostic
```

src/exo/worker/runner/bootstrap.py
```diff
@@ -56,25 +81,20 @@ def entrypoint(
...
         logger.opt(exception=e).warning(
             f"Runner {bound_instance.bound_runner_id} crashed with critical exception {e}"
         )
-        event_sender.send(
-            RunnerStatusUpdated(
-                runner_id=bound_instance.bound_runner_id,
-                runner_status=RunnerFailed(error_message=str(e)),
-            )
-        )
+        event_sender.send(RunnerTerminationError.from_exception(e))
```
Evanev7, line 96: yeah actually this is significantly better than just stringifying the error huh

.codex
Evanev7, line 1: .gitignore

Evanev7 (approved):
really good! just need to get these in the dashboard now!

PR #2102, opened 2026-05-21
fix(bug): EventRouter lifetime-handling fixed, no more process crashes

### LINKED ISSUE #2101: BrokenResourceError in _forward_info crashes process during multi-node pipeline-parallel inference
### Environment
- **exo version**: 0.3.70 (commit 90f24bef)
- **Python**: 3.13.13
- **Platform**: Linux aarch64 (NVIDIA GB10) + macOS (Mac Studio M3 Ultra)
- **Cluster**: 2-node pipeline-parallel over 10GbE, using libp2p with `EXO_BOOTSTRAP_PEERS`
- **Models**: 3 models loaded simultaneously (Qwen3-Coder-Next-4bit, Qwen3.6-35B-A3B-4bit, Qwen3.5-9B-8bit)

### Description

When running 8 parallel inference requests (~16k tokens each) against a 2-node pipeline-parallel cluster, the entire exo process crashes due to an unhandled `anyio.BrokenResourceError` in the worker's `_forward_info` method. The error propagates up through nested `TaskGroup`s and kills the process.

The root cause: when a peer node disconnects (or the internal event channel's receiver side closes), `_forward_info` at `src/exo/worker/main.py:126` attempts to `send()` on a `MemoryObjectSendStream` whose channel is already closed (`open_send_channels=0, open_receive_channels=0`). This raises `BrokenResourceError` which is not caught, bubbles through the TaskGroup hierarchy, and terminates the process.

### Additional issue: API dies but process stays alive

In some cases, the `BrokenResourceError` kills only the API server task group while the libp2p ping/pong continues. This leaves the process alive (systemd doesn't restart it) but port 52415 stops listening — a zombie state requiring external health checks to detect.

### Traceback

```
ExceptionGroup: unhandled errors in a TaskGroup (1 sub-exception)
  | File "/home/${user}/exo/src/exo/main.py", line 153, in run
  | File "/home/${user}/exo/src/exo/utils/task_group.py", line 65, in __aexit__
    | File "/home/${user}/exo/src/exo/worker/main.py", line 105, in run
    | File "/home/${user}/exo/src/exo/utils/task_group.py", line 65, in __aexit__
      | File "/home/${user}/exo/src/exo/worker/main.py", line 126, in _forward_info
      | File "anyio/streams/memory.py", line 243, in send
      | File "anyio/streams/memory.py", line 214, in send_nowait
      |   raise BrokenResourceError
      | anyio.BrokenResourceError
```

### Steps to Reproduce

1. Set up 2-node cluster (Linux + macOS) with `EXO_BOOTSTRAP_PEERS`
2. Load 3 models with `POST /place_instance` (min_nodes=2)
3. Send 8 parallel chat completion requests (~16k tokens each)
4. Node disconnection/reconnection triggers `BrokenResourceError`
5. Process exits with status 1

### Expected Behavior

- `_forward_info` should catch `BrokenResourceError` and handle peer disconnection gracefully
- The API server should not die when a worker task fails
- Process should exit cleanly so systemd can restart it (not enter zombie state)

### Suggested Fix

In `src/exo/worker/main.py:126` (`_forward_info`), wrap the `send()` call:

```python
try:
    await sender.send(info)
except (BrokenResourceError, ClosedResourceError):
    logger.warning("Event channel closed, peer may have disconnected")
    break
```

### Workaround

Added a systemd health check watchdog that polls `/node_id` every 30s and force-restarts exo if the API becomes unresponsive.

AndreiCravtov:
## Motivation

Trying to (partially) fix [this](https://github.com/exo-explore/exo/issues/2101) issue. 

## Changes

Changed channels (in channels.py) to support exception overriding.

Made EventRouter channels throw a subclass of the resource closed/broken errors.

The current lifetime logic of EventRouter in event loop no longer blows up because components that use channels from EventRouter now catch the subclass exceptions in the run method: Worker, Master, DownloadCoordinator, RunnerSupervisor.

Added logic to throw when API server exits without being asked to shut down - this kill the sleep-forever in the task-group.


src/exo/download/coordinator.py
```diff
@@ -140,6 +151,10 @@ async def run(self) -> None:
                 tg.start_soon(self._command_processor)
                 tg.start_soon(self._emit_existing_download_progress)
         finally:
+            # don't forget to clean up resources (does this need a cancel-shield??)
```
Evanev7: no - its sync.

src/exo/download/coordinator.py
```diff
@@ -63,6 +68,14 @@ class DownloadCoordinator:
     def __post_init__(self) -> None:
         self.shard_downloader.on_progress(self._download_progress_callback)
 
+    async def _send_event(self, e: Event):
+        try:
+            await self.event_sender.send(e)
+        except (BrokenResourceError, ClosedResourceError):
+            # Event router has been closed - stop download coordinator & yield
+            self._tg.cancel_tasks()
+            await anyio_checkpoint()
```
Evanev7: why include the checkpoint?

src/exo/master/main.py
```diff
@@ -160,6 +174,7 @@ async def run(self):
     async def shutdown(self):
         logger.info("Stopping Master")
         self._tg.cancel_tasks()
+        # shouldn't there be a wait-stopped signal/event like in worker?
```
Evanev7: no, the run() function returning is that event.

src/exo/download/coordinator.py
AndreiCravtov: to yield to the runtime? I saw someone else in our codebase use anyio.sleep(0) for the same purpose presumably? but apparently theres a function that does that

src/exo/worker/main.py
```diff
@@ -113,6 +118,8 @@ async def run(self):
         finally:
             # Actual shutdown code - waits for all tasks to complete before executing.
             logger.info("Stopping Worker")
+
+            # (don't you need a cancel shield here??)
```
Evanev7: also sync

src/exo/download/coordinator.py
AndreiCravtov: gone

Evanev7 (approved):
tentative approve. the aclose, __enter__, __exit__ and __del__ methods aren't relevant (you can't throw in __del__, __exit__ and aclose just call close() afaik, __enter__ is just return self). id prefer if you made a patch/patch_async method rather than iscoroutinefunction. the typing is cute.

it's some python nonsense still. we need to remove this restarting behaviour asap, but its going to have to wait until post event sourcing which is still a ways off.

AndreiCravtov:
> tentative approve. the aclose, **enter**, **exit** and **del** methods aren't relevant (you can't throw in **del**, **exit** and aclose just call close() afaik, **enter** is just return self). id prefer if you made a patch/patch_async method rather than iscoroutinefunction. the typing is cute.
> 
> it's some python nonsense still. we need to remove this restarting behaviour asap, but its going to have to wait until post event sourcing which is still a ways off.

If I make the patch/patch_async methods typing it wrong would be an easy silent failure, specifically, the sync version's typing is compatible with async too, in the wrong way. you can do .patch(async_method) and that will work and not yell at you at all, but that will wrap the creation of the coroutine in the try-catch. I'd rather do this one instead dince I would need to use iscoroutinefunction anyways

Ofcourese I still could have patch and patch_async where patch has something like:

assert not iscoroutinefunction(fn)

but at that point why not stick them into a single method?

Issue #1931, opened 2026-04-19
running in background


Smiril:
run as a daemon in background
on a linux box

AndreiCravtov:
Implemented in my PR branch

AndreiCravtov:
done

PR #2106, opened 2026-05-22
feat: added background/daemon support

AndreiCravtov:
## Motivation

Addresses [this](https://github.com/exo-explore/exo/issues/1931) issue.

## Changes

You can now launch Exo as a legacy SysV-style daemin (in the background) with `--legacy-daemon` flag.
NOTE: don't use it if you're managing Exo with systemd or launchd

SIDE FIX: the macmon process not found trace is no longer displayed on process shutdown via ctrl+c, that error is supressed.

## Why It Works

Because I used a daemonization library and tweaked it not to break multiprocessing. 

## Test Plan

I ran it in daemon mode, non daemon mode, etc., and pid locking + inference + everything else works just fine.

Also ran it `ssh user@host -t 'cd exo && nohup nix run .#exo -- --legacy-daemon'` on a 4-node TB mac-mini cluster and the mDNS didn't die


src/exo/utils/info_gatherer/info_gatherer.py
```diff
@@ -612,6 +614,11 @@ async def _monitor_macmon(self, macmon_interval: float):
                             text = data.decode("utf-8", errors="replace").strip()
                             metrics = MacmonMetrics.from_raw_json(text)
                         await self.info_sender.send(metrics)
+                finally:
+                    # we need to manually close to prevent process lookup error
+                    # from bubbling up to log when shutting down with ctrl+c
+                    with contextlib.suppress(ProcessLookupError):
+                        await p.aclose()
             except TimeoutError:
```
Evanev7: nit but this whole thing is already wrapped in a try except

src/exo/main.py
```diff
@@ -274,14 +275,60 @@ async def _elect_loop(self):
...
+            # 2) otherwise      => just write PID
+            try:
+                pidfile.write()
+            except PidfileError as e:
+                print(e, file=sys.stderr)
+                raise SystemExit(1) from e
+            main_inner(args)
+    finally:
+        del pidfile  # ensure this object is GC-d
+
```
Evanev7, line 329: its a gcd language, whats the concern here?

src/exo/main.py
```diff
@@ -274,14 +275,60 @@ async def _elect_loop(self):
...
+                detach_process=True,
+                files_preserve=[pidfile.as_raw_fd()],
+                stdin=stdin,
+                stdout=stdout,
+                stderr=stderr,
+            ):
+                # cleanup loose file descriptors (as long as they aren't stdio)
+                for f in (
+                    f for f in (stdin, stdout, stderr) if f.fileno() not in STDIO_FDS
+                ):
```
Evanev7, line 309: is this possible? and even if it is, if STDIN is actually closed, why do we want to fail to clean up fds here?

rust/exo_pyo3_bindings/src/pidfile.rs
```diff
@@ -77,6 +85,35 @@ impl PyPidfile {
...
+        self.0.will_cleanup()
+    }
+
+    /// Dangerously configures if destructor cleanup logic should run.
+    ///
+    /// Disabling cleanup can be used to prevent removal of the file,
+    /// for the purposes of e.g. double-fork demonization.
+    fn _dangerously_set_cleanup(&mut self, cleanup: bool) {
+        unsafe { self.0.config_cleanup(cleanup) }
+    }
```
Evanev7: is this used anywhere?

Evanev7 (approved):
just some suggestions for cleanup! looks good

src/exo/utils/info_gatherer/info_gatherer.py
AndreiCravtov: OHH yes, I can literally just catch ProcessLookupError: pass, how did I not think of that

src/exo/main.py
AndreiCravtov: I don't know I'm paranoid that the runtime will die before running the destructors associated lol

rust/exo_pyo3_bindings/src/pidfile.rs
AndreiCravtov: I thought it would be - its not

src/exo/main.py
AndreiCravtov: It [might be](https://stackoverflow.com/a/22370524) possible in the sense that its not forbidden by implementation. 

IF it is possible, we want STDIO always open to something, otherwise multiprocessing fails, as you yourself confirmed, and I also ran into the issue where closing STDIO makes multiprocessing not work.

So yeah, if by some miracle stdio is closed and these files are allocated 0,1,2 FDs then we want to keep them open for the lifetime of the program.

src/exo/main.py
AndreiCravtov: Actually, I'll just turn it into a context manager to make sure its actually cleaned up

PR #2114, opened 2026-05-25
fix(bug): no longer repeated _trigger_notify_user_to_download_model

### LINKED ISSUE #2098: EXO spontaneously shuts down and repeatedly prompts model download despite model existing
## Problem Description

EXO exhibits two related issues:

1. **Spontaneous shutdown**: EXO process terminates unexpectedly without user intervention
2. **Repeated download prompts**: EXO repeatedly calls `_trigger_notify_user_to_download_model` even when the model may already exist locally

## Observed Symptoms

From EXO logs (2026-05-18):
```
[ 2026-05-18 01:34:52.140 | WARNING  | exo.api.main:_trigger_notify_user_to_download_model:741 ] TODO: we should send a notification to the user to download the model
[ 2026-05-18 01:34:52.142 | WARNING  | exo.api.main:_trigger_notify_user_to_download_model:741 ] TODO: we should send a notification to the user to download the model
[ 2026-05-18 01:35:16.171 | WARNING  | exo.api.main:_trigger_notify_user_to_download_model:741 ] TODO: we should send a notification to the user to download the model
[ 2026-05-18 01:35:16.174 | WARNING  | exo.api.main:_trigger_notify_user_to_download_model:741 ] TODO: we should send a notification to the user to download the model
```

These warnings repeat multiple times in short succession.

## Related Issue

This issue is related to (and was escalated from) **OpenClaw Issue #82453** which has been **CLOSED** with the following determination from maintainer clawsweeper:

> \"The RequestEventLog messages in the OpenClaw logs appear to originate from within Exo itself, not from OpenClaw code. The OpenClaw codebase does not contain any references to RequestEventLog.\"

**OpenClaw Issue #82453**: https://github.com/openclaw/openclaw/issues/82453 (CLOSED)

The clawsweeper identified that the following log entries are from Exo internal Python modules:
- `exo.master.main:_command_processor`
- `exo.routing.event_router`

This suggests the root cause lies within EXO, not OpenClaw.

## Request

1. Investigate why EXO spontaneously shuts down
2. Fix the repeated download notification logic - EXO should not repeatedly prompt for download if the model already exists
3. Review the `_trigger_notify_user_to_download_model` implementation for proper state checking

## Environment

- EXO version: (please fill in)
- Platform: macOS (Apple Silicon)
- Python version: 3.x

AndreiCravtov:
## Motivation

Partially fixes [this](https://github.com/exo-explore/exo/issues/2098) issue. Removed erroneous logic for telling user to download when they already downloaded.

Could not figure out about the "spontaneous crashes" in that issue, author should consolidate more logs and open a new issue dedicated to that. I believe [this](https://github.com/exo-explore/exo/commit/74e9fe15e62fe189dc7e019db86e75c83eca2721) commit solved some EventRouter-related crashes, which was mentioned in [this](https://github.com/exo-explore/exo/issues/2098) issue, so it may have already been solved. If not, should be re-submitted as a new issue.

## Changes

- Consolidated _resolve_and_validate_text_model and _validate_image_model into one function: _validate_model_has_instance;
- + They already had virtually identical logic, it being different seems to be an artifact of history
- + Added logic to ensure that _trigger_notify_user_to_download_model is only called when no such model is downloaded, not just if there is no instance of it
- Added a new `/instance/await` SSE streaming endpoint to wait for when a model has an instance available. Complements instance-placement API, so we can wait till that is done without client-side polling.
- Updated docs and a /tmp script to reflect some of the changes
- Updated dashboard `getModelForRequest` to only return model ID if an instance exists for it, and updated bits to use `handleChatSend` instead of `sendMessage` because that checks for if a model instance exists first. 

## Why It Works

The problem was that there was erroneous logging for model not downloaded. I fixed that logic. The rest is extra.


src/exo/api/main.py
```diff
@@ -939,38 +982,31 @@ async def bench_chat_completions(
...
-        return model_id
+            if not model_is_downloaded:
+                await self._trigger_notify_user_to_download_model(model_id)
 
-    async def _validate_image_model(self, model: ModelId) -> ModelId:
-        """Validate model exists and return resolved model ID.
-
-        Raises HTTPException 404 if no instance is found for the model.
-        """
-        model_card = await ModelCard.load(model)
```
rltakashige, line 963: I feel like ModelCard.load(model_id).model_id was there for a reason. @ciaranbor can you elaborate

src/exo/master/main.py
```diff
@@ -181,18 +181,21 @@ async def _command_processor(self) -> None:
                         case TestCommand():
                             pass
                         case TextGeneration():
+                            # set-difference => prefill-only nodes
```
rltakashige, line 184: This logic will be fixed, but thanks for the comments

rltakashige (approved):
Looks good, thanks! Just check that one image model functionality with Ciaran. 

Also, the naming of the endpoint did make me think that it was for awaiting an instance to become ready, rather than for it to exist, but I can't think of a better name so it's fine :)

src/exo/api/main.py
AndreiCravtov: I agree, im curious to know. 

It looks like a vestige of when model IDs used to have aliases? so this may have been a way to normalize these into the proper normal ID.

Or maybe to trigger some kind of side-effect unintentionally. But ATM it kinda looks like it does nothing. I may be wrong ofcourse

src/exo/master/main.py
AndreiCravtov: nws

src/exo/api/main.py
ciaranbor: You're right, no need for this anymore

PR #2118, opened 2026-05-26
bugfix: prevent libp2p swarm starvation at startup on macOS

### LINKED ISSUE #1919: Two full Exo instances fail to establish libp2p connection (macOS Apple Silicon)
## Description

When two full Exo instances are running on separate Apple Silicon Macs connected via Thunderbolt bridge (link-local network), they fail to establish a libp2p connection. However, simpler configurations using the same networking stack work correctly.

## Environment

- macOS (Darwin), Apple Silicon (M4 Pro + M4)
- Exo commit: fd5b2328
- Python 3.13, libp2p with gossipsub, noise, yamux, pnet
- Two machines connected via Thunderbolt bridge (169.254.x.x link-local)

### Works ✅
- **Simple test ↔ Simple test**: Two minimal scripts that create `NetworkingHandle`, subscribe to topics, and call `recv()` — connection established, messages exchanged.
- **Simple test ↔ Full Exo**: One side runs a minimal test script, the other runs full `exo` — connection established.
- **Router + Election only ↔ Router + Election only**: Using `Router` and `Election` without other components — connection established.

### Fails ❌
- **Full Exo ↔ Full Exo**: Both sides running `exo` main entry point — connection never established. No `PyFromSwarm_Connection` event is received on either side.

## Observations

1. The `NetworkingHandle` and libp2p swarm work correctly in isolation — the Rust networking layer successfully discovers peers, dials, and exchanges gossipsub messages in simpler configurations.

2. When running full Exo on both sides, the swarm event loop (using `tokio::select!` between `from_client.recv()` and `swarm.next()`) appears to not yield connection events, despite the same bootstrap peer configuration that works in simpler tests.

3. The issue may be related to how the swarm event loop handles concurrent subscriptions and message processing when many topics are registered simultaneously during startup. In the full Exo startup, 6 topics are subscribed sequentially, and multiple components (`Router`, `EventRouter`, `Election`, `Master`, `Worker`, `DownloadCoordinator`, `API`) are started concurrently.

4. Both machines use identical Rust binaries (same `.so` file).

## Suspected Root Cause

The `tokio::select!` loop in `swarm.rs` may experience starvation or event processing issues when the full set of Exo components generates high channel traffic during startup. The `from_client.recv()` arm may be consuming events (topic subscriptions, publish commands) at a rate that prevents the `swarm.next()` arm from processing connection-related swarm events.

## Additional Context

An election race condition was also identified and fixed separately: when two connection events arrive in quick succession, the second event's campaign cancels the first, causing both nodes to elect themselves as master. This was fixed by skipping connection-triggered elections when a campaign is already in progress. However, this fix is irrelevant to the core connection failure described above, since connections are never established in the first place.

simongonzalezdc:
## Motivation

When two full EXO instances run on separate Apple Silicon Macs connected via
Thunderbolt bridge, they fail to establish libp2p connections. The simpler
test configurations (NetworkingHandle alone, or full EXO against a simple test)
work correctly.

Root cause: the `tokio::select!` loop in `rust/networking/src/swarm.rs` has
no `biased;` directive. During EXO startup, many components (Router,
EventRouter, Election, Master, Worker, DownloadCoordinator, API) subscribe
to topics simultaneously, flooding `from_client.recv()`. Without biased
ordering, `swarm.next()` is starved and `ConnectionEstablished` events are
never processed.

Fixes: exo-explore/exo#1919

## Changes

**File**: `rust/networking/src/swarm.rs`

Add `biased;` to `tokio::select!` and reorder arms so `swarm.next()` is
checked before `from_client.recv()`.

## Why It Works

`tokio::select!` without `biased` randomly picks an arm when multiple
are ready. With `biased;`, the loop checks arms in declaration order.
By placing `swarm.next()` first, connection events are processed immediately
even when `from_client` has queued messages.

### Manual Testing

Hardware: Mac mini M4 + MacBook Pro M1 Pro connected via Thunderbolt bridge
`10.42.0.1` / `10.42.0.2`

1. Pull this branch on both machines
2. Rebuild: `cargo build --release -p exo_pyo3_bindings`
3. Start MBP (worker): `EXO_MASTER_CANDIDATE=false uv run exo -v -m
   --api-port 52415 --libp2p-port 52417
   --bootstrap-peers /ip4/10.42.0.1/tcp/52416`
4. Start Mini (head): `uv run exo -v -m --api-port 52415 --libp2p-port 52416
   --bootstrap-peers /ip4/10.42.0.2/tcp/52417`
5. Wait 30s, check `curl -s http://127.0.0.1:52415/state` → expect 2 nodes

**Expected**: 2 nodes visible ✅  
**Before fix**: 0 nodes (libp2p connections never established)

6. Regression: verify Tailscale bootstrap still works (2 nodes with
   `100.x.x.x` addresses)

### DIFF
--- a/rust/networking/src/swarm.rs
+++ b/rust/networking/src/swarm.rs
@@ -1,4 +1,5 @@
 use std::pin::Pin;
+use std::time::Duration;
 
 use crate::swarm::transport::tcp_transport;
 use crate::{alias, discovery};
@@ -62,15 +63,21 @@ impl Swarm {
         let stream = async_stream::stream! {
             loop {
                 tokio::select! {
-                    msg = from_client.recv() => {
-                        let Some(msg) = msg else { break };
-                        on_message(&mut swarm, msg);
-                    }
+                    // Process swarm events first to prevent starvation during startup.
+                    // At startup, many components subscribe to topics simultaneously, which
+                    // continuously feeds from_client. Without biased ordering, the
+                    // swarm.next() arm is starved and ConnectionEstablished events are
+                    // never processed, causing two-node EXO clusters to fail to form.
+                    biased;
                     event = swarm.next() => {
                         let Some(event) = event else { break };
                         if let Some(item) = filter_swarm_event(event) {
                             yield item;
                         }
+                    }
+                    msg = from_client.recv() => {
+                        let Some(msg) = msg else { break };
+                        on_message(&mut swarm, msg);
                     }
                 }
             }
@@ -162,6 +169,7 @@ pub fn create_swarm(
         .with_tokio()
         .with_other_transport(tcp_transport)?
         .with_behaviour(|keypair| Behaviour::new(keypair, parsed_bootstrap_peers))?
+        .with_swarm_config(|config| config.with_idle_connection_timeout(Duration::from_secs(600)))
         .build();
 
     swarm.listen_on(format!("/ip4/0.0.0.0/tcp/{listen_port}").parse()?)?;


simongonzalezdc:
Testing environment had stale node ID state. Will reopen with clean test results.

PR #2126, opened 2026-05-28
feat(telemetry): MVP support for opt-in telemetry of runner crashes

AndreiCravtov:
## Motivation

Runner crashes currently leave useful diagnostics only on the local machine. This adds an opt-in path for collecting raw runner stderr logs so crash reports can be inspected centrally without making telemetry part of the critical execution path.

## Changes

- Added `--telemetry`, disabled by default.
- Added `EXO_TELEMETRY_API_URL`, defaulting to `https://telemetry.exolabs.net/`.
- Added `TelemetryService` and `TelemetrySink` for bounded, non-blocking telemetry submission.
- Wired telemetry from `Node` through `Worker` into `RunnerSupervisor`.
- Changed runner stderr logging to write per-runner timestamped files under:
  `runner_log/<instance_id>/<runner_id>/<utc_timestamp>.stderr.log`
- Upload runner stderr logs by:
  - reading the completed stderr file
  - skipping zero-byte files
  - hashing the raw file bytes with SHA-256
  - requesting a pre-signed upload URL
  - uploading the raw bytes to that URL
- Swallow telemetry upload failures after logging a warning.

## Why It Works

Telemetry is isolated behind a bounded channel and best-effort background service, so failures, slow uploads, or backpressure do not block runner supervision or node operation.

Runner stderr files are written per bound instance/runner with UTC timestamped names, avoiding the previous shared append-only log path. Hashing the final file contents before upload gives the backend a stable object identity and lets it verify that uploaded contents match the requested submission.

### Manual Testing

- Ran exo with `--telemetry` against the telemetry API.
- Forced a runner failure and confirmed a non-empty runner stderr log was submitted through the pre-signed upload flow.

### Automated Testing

- Added unit coverage for runner stderr telemetry upload:
  - SHA-256 presign payload generation
  - raw byte upload to the returned URL
  - upload failure handling without propagating exceptions
- Updated runner supervisor tests to use isolated per-test log paths and a dummy telemetry service.


src/exo/worker/runner/supervisor.py
```diff
@@ -74,58 +81,68 @@ class RunnerStdioHandler:
...
+
+    async def _handle_stdout(self):
+        # We don't expect anything in stdout so even reading this at all is going
+        # to be quite weird; hence handle it by logging error and the received chunk
+
+        rx = TextReceiveStream(self._stdout_rx, encoding="utf-8", errors="replace")
+        try:
+            async with rx:
+                async for chunk in rx:
+                    logger.error(f"Unexpected runner stdout chunk: {chunk}")
```
Evanev7: definitely shouldnt be an error - more complex runners may write to stdout

src/exo/main.py
```diff
@@ -445,6 +453,11 @@ def parse(cls) -> Self:
             action="store_true",
             help="Disable continuous batching, use sequential generation",
         )
+        parser.add_argument(
+            "--telemetry",
+            action="store_true",
+            help="Enable telemetry uploads. Disabled by default; disabled mode keeps telemetry in dry-run.",
```
Evanev7, line 459: cool. this needs to be persisted somewhere because as-is noone will enable it, but thats kinda fine actually.

src/exo/shared/telemetry.py
```diff
@@ -0,0 +1,176 @@
...
+    api_url: str
+    _send: Sender[TelemetrySubmission]
+    _recv: Receiver[TelemetrySubmission]
+    _http_transport: httpx.AsyncBaseTransport | None
+    _tg: TaskGroup = field(default_factory=TaskGroup, init=False)
+
+    @classmethod
+    def create(
+        cls,
+        dry_run: bool,
```
Evanev7: can i ask you to rename dry_run to "telemetry_disabled" or similar? it makes it sound like its for testing when it's the main toggle entrypoint

src/exo/main.py
AndreiCravtov: thats gonna be for a future commit :)

PR #2128, opened 2026-05-29
fix: drop undeserializable gossip messages instead of crashing the node

jasonpaulso:
## Problem

A single gossip message that fails to deserialize crashes the **entire receiving node**.

`Router._networking_recv` calls `router.publish_bytes(data)` → `topic.deserialize(data)` → `model_validate_json`. A `pydantic.ValidationError` there propagates out of the `while True` loop, tears down the receive loop's task group, and brings the whole node down (observed as a `BaseExceptionGroup` cascade, with a collateral `BrokenResourceError` in `event_router._ingest`).

In effect this is a **remote node-kill**: any peer publishing a payload this node can't deserialize takes the node offline.

## Why it isn't widely reported

It only triggers on **schema/version skew between nodes on the same network**. Same-version peers serialize identically and round-trip fine (`FrozenModel` uses `alias_generator=to_camel` + `validate_by_name=True`). The failure appears when two nodes run different commits — exactly what happens during a **rolling upgrade** (update one node, the others are briefly still on the old build) or when separate machines are updated at different times. The event/model-card schema changes often enough that any mismatched pair on a LAN will hit it.

Reproduced in the wild with two checkouts ~4 weeks apart auto-discovering each other: each node floods the other with events it rejects (`NodeDownloadProgress`, `NodeGatheredInfo`), and both die — one crash-loops under its launchd `KeepAlive`.

## Fix

Catch `ValidationError` around the per-message `publish_bytes` call, log a warning naming the origin peer, and `continue`.

The outer `except Exception: … raise` added in #1621 (to surface silent shutdowns from e.g. a Rust-layer `ConnectionError` on a closed channel) is **preserved**. This change only makes *untrusted peer input* non-fatal — it narrows the fatal set rather than silencing genuine loop failures.

## Test plan

- New unit test `test_router_recv.py::test_undeserializable_message_is_dropped_not_fatal`: drives a bad message followed by a valid one through `_networking_recv`, asserts the valid one is still delivered (i.e. the loop survived the bad one). Fails without the fix.
- `uv run pytest src/exo/routing/` — 9 passed
- `uv run ruff check` / `ruff format --check` — clean
- Verified live: a patched node stays up and serving (API 200) while logging `Dropping undeserializable message on local_events from <peer>: 83 validation errors`; bringing the stale node up to the same commit eliminated the errors entirely in both directions.

### DIFF
--- a/src/exo/routing/router.py
+++ b/src/exo/routing/router.py
@@ -22,6 +22,7 @@ from exo_pyo3_bindings import (
 )
 from filelock import FileLock
 from loguru import logger
+from pydantic import ValidationError
 
 from exo.shared.constants import EXO_NODE_ID_KEYPAIR
 from exo.utils.channels import Receiver, Sender, channel
@@ -201,7 +202,17 @@ class Router:
                             )
                             continue
                         router = self.topic_routers[topic]
-                        await router.publish_bytes(data)
+                        try:
+                            await router.publish_bytes(data)
+                        except ValidationError as exception:
+                            # A peer on an incompatible version (or a corrupt
+                            # packet) can produce undeserializable payloads.
+                            # Drop the message rather than killing the node.
+                            logger.warning(
+                                f"Dropping undeserializable message on {topic} "
+                                f"from {origin}: {exception.error_count()} validation errors"
+                            )
+                            continue
                     case PyFromSwarm.Connection():
                         message = ConnectionMessage.from_update(from_swarm)
                         logger.trace(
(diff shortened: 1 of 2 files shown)


Evanev7:
we do not currently intended to support any cross-version compatibility. we will in future, but exo is currently in a beta phase as we work out some core features like p/d dissaggregation. the noisy failure is actually quite helpful to this end, as our network is about to change significantly (#2076)

appreciate the pr but im closing this for now

PR #2133, opened 2026-06-01
feat(api): GET /events accepts since and limit query params with X-EXO-Last-Idx header

5F0jd2vLq54RerYW:
## Summary

Extends `GET /events` to support cursor-based tailing without polling the full `/state` snapshot.

**Changes:**
- `src/exo/api/main.py`: Add optional `since` (default 0) and `limit` (default None) query params to `stream_events()`. Takes a fast-path to `read_all()` when both are absent — identical to pre-patch behavior.
- `src/exo/api/tests/test_stream_events.py`: 7 test cases covering full-dump backward compat, `since`+`limit`, `since`-only, `since`-beyond-count (empty), limit-clamped, negative `since` rejected (FastAPI `ge=0`), and chained cursor reads with no overlap.
- `src/exo/api/tests/conftest.py`: `sys.modules` stub for `exo_rs` so API tests run without requiring a compiled Rust extension.

**Response header:** `X-EXO-Last-Idx: <N>` is set to the upper exclusive bound of the range consumed, allowing clients to chain reads with no gap and no overlap.

## Motivation

Downstream tooling needs to detect state changes efficiently. The previous approach required polling the full `/state` snapshot (~150KB). With this change, a client can tail events incrementally:

```python
cursor = 0
while True:
    resp = requests.get(f"http://localhost:52415/events?since={cursor}")
    events = resp.json()
    cursor = int(resp.headers["X-EXO-Last-Idx"])
    process(events)
    sleep(1)
```

## Backward Compatibility

`GET /events` with no query params takes the same `read_all()` fast-path as before. No existing client behavior changes.

### DIFF
--- a/src/exo/api/main.py
+++ b/src/exo/api/main.py
@@ -1010,7 +1010,28 @@ class API:
             )
         return model_id
 
-    def stream_events(self) -> StreamingResponse:
+    async def _validate_image_model(self, model: ModelId) -> ModelId:
+        """Validate model exists and return resolved model ID.
+
+        Raises HTTPException 404 if no instance is found for the model.
+        """
+        model_card = await ModelCard.load(model)
+        resolved_model = model_card.model_id
+        if not any(
+            instance.shard_assignments.model_id == resolved_model
+            for instance in self.state.instances.values()
+        ):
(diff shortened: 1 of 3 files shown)


5F0jd2vLq54RerYW:

PR #2134, opened 2026-06-01
fix(security): block SSRF in fetch_image_url via scheme allowlist and private IP rejection

5F0jd2vLq54RerYW:
## Summary

`fetch_image_url()` in `chat_completions.py` fetches arbitrary URLs provided by API clients with no validation. This allows SSRF attacks:

- `http://169.254.169.254/latest/meta-data/iam/security-credentials/` leaks AWS IAM credentials
- `http://192.168.1.1/admin` scans the internal network
- `file:///etc/passwd` reads local files (if aiohttp supports the scheme)

## Fix

Adds three validation checks before `session.get()` is ever called:

1. **Scheme allowlist** — only `http`/`https` permitted
2. **Metadata host blocklist** — rejects AWS IMDSv1, GCP, and Azure IMDS endpoints
3. **Literal IP rejection** — `ipaddress.ip_address(hostname)` detects private, loopback, and link-local IPs; DNS hostnames are not blocked (DNS rebinding protection is deferred)

No new package dependencies — uses only stdlib `ipaddress` and `urllib.parse`.

## Tests

12 test cases in `src/exo/api/adapters/tests/test_fetch_image_url.py`:
- Every rejection case asserts `session.get()` is **never called** (mock + `assert_not_called()`)
- Coverage: scheme, all 3 metadata hosts, RFC 1918, loopback, link-local, valid HTTPS, public IP literal, hostname passthrough

## Scope

**In scope:** Literal IP validation at parse time.
**Out of scope:** DNS rebinding (hostname → private IP after resolution). That requires resolving hostnames before connecting, which is a larger change best handled separately.

### DIFF
--- a/src/exo/api/adapters/chat_completions.py
+++ b/src/exo/api/adapters/chat_completions.py
@@ -1,10 +1,12 @@
 """OpenAI Chat Completions API adapter for converting requests/responses."""
 
 import base64
+import ipaddress
 import re
 import time
 from collections.abc import AsyncGenerator
 from typing import Any
+from urllib.parse import urlparse
 
 from exo.api.types import (
     ChatCompletionChoice,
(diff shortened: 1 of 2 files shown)


5F0jd2vLq54RerYW:

PR #2135, opened 2026-06-01
fix(master): add --no-master flag and prevent --no-api nodes from self-electing

5F0jd2vLq54RerYW:
## Summary

Two related fixes for node role management in multi-node clusters:

### 1. `--no-master` flag

New CLI flag that skips `Master` instantiation entirely. Previously every node always started a Master even when it was intended to run as a pure worker (e.g. a GPU worker with no API surface). With `--no-master`, the Master event loop, command processor, and download coordinator are not started.

### 2. `--no-api` nodes must not self-elect

When a node is started with `--no-api` (`spawn_api=False`), it should not be eligible to become master — an API-less node winning an election means no API is reachable in the cluster. Previously, a `--no-api` node in a solo partition would self-elect (the bully algorithm has no other candidates). This PR sets `is_candidate = args.spawn_api and not args.no_master`, which gives non-candidate nodes a seniority of `-1` so any API-bearing node beats them.

A complementary guard in `election.py` ensures non-candidate nodes re-propose the last known master during solo partitions instead of winning by default.

### Flag precedence

| Flags | Result |
|-------|--------|
| `--no-master` | master=None, is_candidate=False |
| `--no-api` | is_candidate=False, seniority=-1 |
| `--force-master` | seniority=1_000_000 (only applies when is_candidate=True) |
| `--no-master --force-master` | master=None (--no-master wins) |
| `--no-api --force-master` | seniority=-1 (--no-api wins) |

## Tests

- 14 existing election tests (`src/exo/shared/tests/test_election.py`) all pass
- Added `exo_rs` stub to `src/exo/shared/tests/conftest.py` so tests run without a compiled Rust binary

### DIFF
--- a/src/exo/main.py
+++ b/src/exo/main.py
@@ -46,6 +46,7 @@ class Node:
     node_id: NodeId
     offline: bool
     _api_port: int
+    _no_master: bool
     _tg: TaskGroup = field(init=False, default_factory=TaskGroup)
 
     @classmethod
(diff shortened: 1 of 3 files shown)


5F0jd2vLq54RerYW:

Evanev7:
"an API-less node winning an election means no API is reachable in the cluster" - this is simply not the case?

PR #2174, opened 2026-06-24
fix: pre-populate GPT-OSS tiktoken vocab cache to fix runner crash

sher-gruve:
## Problem
  The runner crashes on the first generation step when using the gpt-oss model with:
  ```
  openai_harmony.HarmonyError: error downloading or loading vocab file: failed to download or load vocab file
  ```
  
  The `openai_harmony` Rust library uses `reqwest` (with `rustls`) to download the `o200k_base.tiktoken` vocab file on first use. This download fails inside Python `spawn` subprocesses on some 
  platforms (confirmed on Linux aarch64), even though the URL is reachable via `curl` and Python's own `urllib`. The encoding is loaded lazily on the first generation step, so the crash surfaces 
  mid-inference.
  
  ## Fix
  Added `_ensure_gpt_oss_vocab_cached()` in `model_output_parsers.py`, called inside `get_gpt_oss_encoding()` before delegating to the Rust `load_harmony_encoding()`. It uses Python's `urllib` 
  to download and cache the file with the correct tiktoken-rs filename (SHA-1 of the URL), respects `TIKTOKEN_RS_CACHE_DIR`, verifies the SHA-256 of the downloaded file, and uses an atomic write to 
  avoid partial files.

### DIFF
--- a/src/exo/worker/runner/llm_inference/model_output_parsers.py
+++ b/src/exo/worker/runner/llm_inference/model_output_parsers.py
@@ -1,3 +1,7 @@
+import hashlib
+import os
+import tempfile
+import urllib.request
 from collections.abc import Callable, Generator, Iterator
 from functools import cache
 from typing import Any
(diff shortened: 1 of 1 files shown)


src/exo/worker/runner/llm_inference/model_output_parsers.py
```diff
@@ -31,9 +35,45 @@
...
+    Python's urllib as a fallback to pre-populate the tiktoken-rs cache so that
+    the Rust library can load from disk instead.
+    """
+    cache_dir = os.environ.get("TIKTOKEN_RS_CACHE_DIR", os.path.join(tempfile.gettempdir(), "tiktoken-rs-cache"))
+    # tiktoken-rs keys cached files by the SHA-1 of the download URL
+    cache_key = hashlib.sha1(_GPT_OSS_VOCAB_URL.encode()).hexdigest()
+    cache_path = os.path.join(cache_dir, cache_key)
+    if os.path.exists(cache_path):
+        return
+    os.makedirs(cache_dir, exist_ok=True)
```
rltakashige, line 63: You could catch the exception here (OSError) in something that tells us that making the GPT OSS cache failed (to be more descriptive), but this is perfectly fine too. 

Thanks for the fix!

rltakashige (approved):

sher-gruve:
hi @rltakashige, there was a ruff formatting issue which I've fixed, you can approve the workflow.

PR #2179, opened 2026-06-26
Add a use_prefix_cache request flag to bypass the KV prefix cache

mlpy0:
## Summary

Adds an optional `use_prefix_cache` flag to chat-completion requests. It defaults to `true` (current behavior). Set it to `false` to run a normal prefill+decode on a fresh per-request cache that is discarded, so the request never reads from or writes to the shared KV prefix cache.

## Motivation

Sometimes a caller wants to run a real generation without disturbing the prefix cache. For example:

- periodic warm-up / liveness probes that keep the model and generation path hot, but should not accumulate throwaway cache entries;
- one-off or automated requests that shouldn't evict other clients' cached conversation prefixes via LRU;
- ad-hoc calls where caching the prompt brings no benefit.

`BenchChatCompletionRequest` already has a `use_prefix_cache` field for exactly this bypass in bench mode. Following review feedback, this promotes that same flag to the base `ChatCompletionRequest` instead of adding a separate flag, so there's one knob doing one job.

## Implementation

- `ChatCompletionRequest.use_prefix_cache` defaults to `true`. `BenchChatCompletionRequest` keeps `use_prefix_cache = false` as an override, so bench behavior is unchanged.
- The chat-completions adapter maps it onto `TextGenerationTaskParams.use_prefix_cache`, whose default is now `true` so every existing caller keeps caching.
- The generators (`mlx_generate` and `ExoBatchGenerator`) skip the prefix cache whenever `use_prefix_cache` is `false`, decoupled from `bench`. `bench` now only governs benchmark measurement, not caching.

## Compatibility

Defaults to `true`, so existing callers are unaffected. The bench path still bypasses the cache by default via its `use_prefix_cache = false` override.

### DIFF
--- a/src/exo/api/adapters/chat_completions.py
+++ b/src/exo/api/adapters/chat_completions.py
@@ -175,6 +175,7 @@ async def chat_request_to_text_generation(
         presence_penalty=request.presence_penalty,
         frequency_penalty=request.frequency_penalty,
         images=images,
+        ephemeral=request.ephemeral or False,
     )
 
 
--- a/src/exo/api/adapters/chat_completions.py
+++ b/src/exo/api/adapters/chat_completions.py
@@ -175,7 +175,7 @@ async def chat_request_to_text_generation(
         presence_penalty=request.presence_penalty,
         frequency_penalty=request.frequency_penalty,
         images=images,
-        ephemeral=request.ephemeral or False,
+        use_prefix_cache=request.use_prefix_cache,
     )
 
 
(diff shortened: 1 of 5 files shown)


rltakashige (commented):
Thanks for the suggestion and PR!

I'm not sure how necessary this is, as BenchChatCompletionRequest contains a use_prefix_cache parameter that seems to be what this does. Perhaps it makes sense to move use_prefix_cache into the ChatCompletionRequest instead.

What do you think? Looks like a harmless improvement so I am also inclined to approve this.

mlpy0:
Yeah, agreed. I've pushed that version.

use_prefix_cache now lives on ChatCompletionRequest and defaults to true, with BenchChatCompletionRequest overriding it to false so bench stays the same. The generators only looked at it in bench mode before, so I changed them to skip the cache whenever it's false instead. Ephemeral flag's gone.

PR #2185, opened 2026-06-28
Fix/bench deepseek v4 encoding import

sher-gruve:
## Motivation

Two independent failures when running benchmarks and GPT-OSS (o-series) models:

  1. `exo_models_bench.py` imported `deepseek_v4_encoding` from `exo.worker.engines.mlx.deepseek_v4_encoding`, which fails because the `exo` package is not installed in the bench venv, and the module 
  lives under a `vendor/` subdirectory anyway.
  2. The `openai_harmony` Rust library (`reqwest`/`rustls`) fails to download the `o200k_base.tiktoken` vocab file when running inside a Python `spawn` subprocess on some platforms, causing GPT-OSS model 
  output parsing to break.

## Changes

  - Vendor `deepseek_v4_encoding.py` into `bench/vendor/` (pure stdlib, zero deps) and update the import to `from vendor.deepseek_v4_encoding`
  - Add `_ensure_gpt_oss_vocab_cached()` in `model_output_parsers.py` that uses Python's `urllib` to download and cache the `o200k_base.tiktoken` file before the Rust layer attempts to load it

## Why It Works

 - Vendoring the file removes the dependency on the `exo` package being importable from the bench venv entirely
  - Pre-populating the tiktoken cache via Python's `urllib` (which respects the process environment and TLS stack) means the Rust layer finds the file already on disk and skips the download that was failing.

### Manual Testing
  
  
  - Ran `exo_models_bench.py` against `mlx-community/DeepSeek-V4-Flash` — no `ImportError`
  - Ran GPT-OSS (`mlx-community/gpt-oss-20b-MXFP4-Q8`) end-to-end — tiktoken vocab loaded without error in spawned subprocess


bench/exo_models_bench.py
```diff
@@ -0,0 +1,1001 @@
+# type: ignore
```
rltakashige, line 1: Could you explain this file? The comment is copied from the tool call eval file, but the file seems to be repeating the existing speed benchmark but with meaningful text? Did you mean to edit exo_bench?

bench/gen_bench_prompts.py
```diff
@@ -0,0 +1,122 @@
+#!/usr/bin/env python3
```
rltakashige, line 1: I don't think this should be in this PR. It probably makes sense for the user to provide an actual text prompt to benchmark rather than building a prompt like this (when they want a real prompt that isn't just a's).

src/exo/worker/runner/llm_inference/model_output_parsers.py
```diff
@@ -1,3 +1,7 @@
+import hashlib
```
rltakashige, line 1: I think you need to rebase

rltakashige (changes requested):

PR #2195, opened 2026-07-02
feat(placement): bond parallel RDMA links in JACCL device matrix

Nottlespike:
## Motivation

`get_mlx_jaccl_devices_matrix` selected a **single** RDMA interface per device pair and discarded any additional parallel links via an early `break`:

```python
for conn in cycle_digraph.get_all_connections_between(node_i, node_j):
    if isinstance(conn, RDMAConnection):
        matrix[i][j] = conn.source_rdma_iface
        break  # <-- discards extra parallel links
```

This left extra Thunderbolt links unused even when two Macs are connected by more than one cable. The underlying **JACCL ring backend is already built to bond**: `RING_MAX_CONNS = 4`, `RingGroup` accepts `left_devices`/`right_devices` as vectors, and `RingImpl` stripes across `n_conns_` wires with per-wire buffers. The backend's JSON parser (`parse_devices_json`) also already accepts either a single string **or** an array of strings per cell. exo was the bottleneck — it threw away links the backend was designed to consume.

## Changes

- **`src/exo/master/placement_utils.py`** — `get_mlx_jaccl_devices_matrix` now:
  - Collects **every** `RDMAConnection` interface per pair instead of breaking on the first.
  - Emits matrix cells as `list[list[list[str] | None]]` (a list of interface names per cell, or `None` on the diagonal).
  - Normalises link counts down to the **minimum** observed across all pairs, so the backend's ring validator (which requires every ring edge to carry the same number of connections — `is_valid_ring` at `jaccl.cpp:130-141`) never rejects an asymmetric matrix.
- **`src/exo/shared/types/worker/instances.py`** — `MlxJacclInstance.jaccl_devices` widened from `list[list[str | None]]` to `list[list[list[str] | None]]`.
- **`src/exo/master/tests/test_placement.py`** — updated the existing tensor-parallel matrix assertions to expect list cells, plus two new tests:
  - `test_jaccl_devices_matrix_bonds_parallel_rdma_links` — multiple parallel links survive into the cell.
  - `test_jaccl_devices_matrix_normalises_asymmetric_link_counts` — asymmetric link counts are trimmed to a uniform minimum.

## Why It Works

The JACCL backend (the forked `mlx-jaccl-fix-small-recv` wheels exo pins) consumes `MLX_IBV_DEVICES` as a JSON matrix. Its parser handles both forms per cell:

```cpp
// jaccl.cpp:44-49
if (names.is_string()) {
  result[rank][dst].push_back(names);
} else if (names.is_array()) {
  for (auto name_it = names.begin(); name_it != names.end(); name_it++) {
    result[rank][dst].push_back(*name_it);
  }
}
```

`RingGroup` then receives `vector<string> left_devices/right_devices`, calls `create_connections` (plural), and `RingImpl` actively stripes sends/recvs across `n_conns_` wires. So emitting arrays was always supported downstream — exo just wasn't feeding the data through.

The uniform-count normalisation is required because `is_valid_ring` rejects a matrix where ring edges have differing connection counts (`jaccl.cpp:136`). Trimming to the minimum is the conservative choice — bonding only helps when it's consistent across the ring, and a partially-bonded ring would be rejected anyway.

**Known limitation:** the mesh path (`get_mesh_connectivity`, `jaccl.cpp:152`) takes only `devices_[rank][dst][0]`, so bonding is effective for ring/tensor-parallel all-reduce (where the heavy traffic is) but not for the mesh fallback. That's a JACCL limitation, not addressable from exo.

### Manual Testing
**Hardware available:** Two Thunderbolt 5-capable Macs, connected by two Thunderbolt 4 cables (TB4 cables between TB5 ports — links negotiate to TB4 signalling). This is exactly the multi-link scenario this change targets: two parallel cables between one pair, which the matrix builder now feeds to the backend as a two-element cell instead of discarding one.

End-to-end throughput benchmarking on this 2×TB4-cable setup is the planned validation. The change is already verified at the configuration/data level (both interface names reach the backend in the cell) and by reading the JACCL ring striping code. A bandwidth measurement on the dual-cable pair vs single-cable would confirm the gain and is the concrete next step I can run with the hardware above.

### Automated Testing
- `uv run ruff check` — clean
- `uv run basedpyright` — 0 errors, 0 warnings
- `uv run pytest src/exo/master/tests/test_placement.py src/exo/shared/tests/ src/exo/utils/info_gatherer/` — 79 passed, 1 skipped (macOS-only `test_tb_parsing`)
- New tests cover: multi-link bonding (2 links → both in cell), asymmetric normalisation (2-link pair trimmed to match 1-link pairs), and the existing single-link case still produces correct single-element list cells.

## Formatting note

`CONTRIBUTING.md` specifies `nix fmt`. I don't have Nix available in my environment, so I ran `ruff format` instead. If CI's `treefmt` config wants something different, a maintainer can re-run `nix fmt` on this branch.

---

Draft — opening for review, not ready to merge. Happy to adjust naming, test coverage, or the normalisation strategy based on feedback.


EternaPeptix:
Tried this out; here are my results; 

# Kimi K2.5 TB5 long-context benchmark

The long-context sweep confirms that Kimi K2.5 decode slows as the KV
cache grows. Four-link RDMA striping improves prefill and long-prompt
end-to-end latency, but it does not improve single-stream decode.

## Median results

| Actual prompt tokens | 1× prefill tok/s | 4× prefill tok/s | 4× change | 1× decode tok/s | 4× decode tok/s | 4× change |
|---:|---:|---:|---:|---:|---:|---:|
| 516 | 265.73 | 271.09 | +2.02% | 22.24 | 21.92 | -1.40% |
| 4,101 | 413.63 | 433.39 | +4.78% | 21.78 | 21.53 | -1.15% |
| 16,389 | 354.09 | 368.17 | +3.98% | 19.60 | 19.50 | -0.52% |
| 32,773 | 279.45 | 292.39 | +4.63% | 17.13 | 16.61 | -3.02% |

Relative to the 516-token point, decode throughput declined:

| Actual prompt tokens | 1× degradation | 4× degradation |
|---:|---:|---:|
| 516 | baseline | baseline |
| 4,101 | -2.07% | -1.82% |
| 16,389 | -11.83% | -11.05% |
| 32,773 | -22.98% | -24.24% |

Although four links did not help decode, the faster prefill reduced median
end-to-end request time for the longer prompts:

| Actual prompt tokens | 1× request time | 4× request time | 4× change |
|---:|---:|---:|---:|
| 516 | 10.37 s | 10.66 s | +2.85% |
| 4,101 | 16.00 s | 15.57 s | -2.74% |
| 16,389 | 55.75 s | 51.55 s | -7.52% |
| 32,773 | 127.55 s | 120.01 s | -5.91% |

## Method

- Model: `mlx-community/Kimi-K2.5`
- Clean exo commit: `9b23e6275d3ddc32b4a74cacefb81749a01fd59a`
- Two-node MLX/JACCL tensor parallel on 512S1 and 512S2
- One-link and four-link sweeps isolated by recycling both exo services
- Three timed repeats at each prompt length
- Exactly 128 completion tokens per timed request
- Temperature 0, thinking disabled, unique nonce, prefix cache requested off
- Fixed ascending context order within each link width
- Runner logs verified the exact device matrices and
  `MLX_JACCL_RING=1 (up to 4 links per peer)` on both ranks

Exo still reported a partial prefix-cache hit of 25–34 tokens, corresponding
to the fixed chat-template prefix. This was consistent across link widths and
is small relative to the long prompts. The 516-token prefill sets each had one
outlier; medians are reported throughout. Long-context prefill and decode
samples were substantially more stable.

## Interpretation

The context-length decode penalty is dominated by the increasing attention and
KV-cache work performed for every generated token. The additional TB5 rails do
not remove that local compute and memory-bandwidth cost. Striping instead helps
the communication-heavy prefill phase by roughly 4–5% at 4K–32K, which is
enough to improve total latency for long prompts despite decode remaining flat
or slightly worse.

Raw result SHA-256:
`40cde9e9c113ab8fa7142e9924bfdc707d2082d9b77be546d8dcb08e72ebc259`

[kimi-k2.5_tb5_context_sweep.json](https://github.com/user-attachments/files/30359742/kimi-k2.5_tb5_context_sweep.json)

PR #2216, opened 2026-07-10
feat: report CUDA GPU VRAM as node memory on Linux

darinlarimore:
## Problem

On Linux, exo reports psutil **system RAM** as a node's total/available memory regardless of the active MLX backend. For a discrete NVIDIA GPU, the memory that actually bounds MLX inference is the GPU's **VRAM**, not system RAM (unlike Apple Silicon's unified memory). Result: the dashboard over-reports memory, and the master can place shards too large for the GPU → OOM at model load.

## Fix

Report GPU VRAM (**total** → dashboard, **free** → placement) when MLX's default device is a CUDA GPU, via `nvidia-smi`. Falls back to psutil system RAM on CPU (`mlx-cpu`) builds, on Apple Silicon (handled by macmon), or when `nvidia-smi` is unavailable — so existing setups are unaffected.

- `profiling.py`: add `MemoryUsage.from_cuda()` + `_query_cuda_vram_bytes()`
- `info_gatherer.py`: add `_mlx_uses_cuda_gpu()` gate; the Linux memory monitor prefers VRAM when the CUDA backend is active

## Testing

Verified on an RTX 4070 (12 GB) node running the `cuda13` extra: node now advertises ~12.9 GB total / ~11.5 GB free instead of 32 GB system RAM. `ruff` and `basedpyright` pass.

### DIFF
--- a/src/exo/shared/types/profiling.py
+++ b/src/exo/shared/types/profiling.py
@@ -1,4 +1,5 @@
 import shutil
+import subprocess
 from collections.abc import Sequence
 from pathlib import Path
 from typing import Literal, Self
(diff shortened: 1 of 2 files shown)


jgawronek:
I've been running this PR on my cluster (2× DGX Spark 128 GB, Mac Studio 256 GB, RTX 3090 24 GB, model: Qwen3.5-397B-A17B-8bit) and it's a big improvement! before it, the Linux node advertised host RAM as if the GPU could use it, placement assigned it ~20 GB of weights, and the runner died with CUDA OOM on load. With this PR the reporting is truthful and that failure mode is gone.

One suggestion from debugging on real hardware: advertising 100% of memory free still causes runtime OOM. Placement fills the advertised number entirely with weights, but the runner additionally needs VRAM for its CUDA context, allocator overhead, and KV cache/activations. Measured on the RTX 3090 (24,576 MiB, display attached):

3 layers = 19.65 GB of weights against 23.3 GB free → 22.3 GB resident after load, runner aborts during warm-up / long-prompt prefill (cudaMallocAsync ... out of memory).
2 layers = 13.1 GB of weights → ~16 GB resident, stable through multi-turn and 15k-token prompts.


The fix that made it stable for me was advertising a usable fraction of free VRAM in from_cuda:

CUDA_VRAM_USABLE_FRACTION: Final = 0.75
usable_vram = int(free_vram * CUDA_VRAM_USABLE_FRACTION)
return cls.from_bytes(
    ram_total=total_vram,
    ram_available=usable_vram if override_memory is None else override_memory,
    ...
)


Two related observations that may be useful for this PR:

Unified-memory devices (DGX Spark GB10): nvidia-smi --query-gpu=memory.total,memory.free returns [N/A] there, so the code correctly falls back to psutil but the same headroom problem applies (weights are wired into system RAM; filling ~95% of available froze the machine hard during load). We cap the psutil fallback at 85% of available on CUDA-unified nodes.

A nice side effect of this PR's truthful reporting: when a crashed runner lingers holding VRAM, subsequent placements are cleanly refused instead of OOMing again.

Happy to send these as a follow-up PR on top of this one (or a patch to this branch) — whichever the author prefers. Related fixes that don't depend on this PR are in #2227 and #2228.

darinlarimore:
Hey @jgawronek Thanks so much for running this on real hardware! I'd love to have you send it as a commit to this branch so it lands under your name. 

I can add you as a collaborator on darinlarimore/exo and you push straight to linux-cuda-vram-node-memory, or maybe easier, open a PR from your fork targeting darinlarimore:linux-cuda-vram-node-memory (not exo-explore:main) and I'll merge it in.

Thank you! 
Yay local LLM's and not big corps!

jgawronek:
Hi @darinlarimore would love be to a collaborator. I've got a lot more stuff to add. :) 

127.0.0.1 is where the heart is, locally hosted all the way!

abendrothj:
Flagging a near-duplicate so it doesn't turn into a competing PR: while working on ring attention (#2255) I'd implemented the same idea — report dedicated accelerator memory and admit placements against it rather than system RAM. Yours predates mine, so I've dropped my version rather than split it out. #2255 now keeps only a ring-specific admission gate that consumes whatever memory figures a node reports, so it picks up your numbers automatically when this lands.

Two notes from that work, take or leave:

**Detection mechanism.** I used NVML (`pynvml`) rather than shelling out to `nvidia-smi`. It's already an indirect dependency (`src/exo/utils/info_gatherer/info_gatherer.py:359` imports it), and it keeps working in containers that expose the driver but not the `nvidia-smi` binary — which is a fairly common Docker/LXC setup, and the one I was testing on. Worth considering as a fallback if `nvidia-smi` isn't found.

**Free vs. total.** Using `free` for placement is the right call, but the number moves under you — on a node that's already hosting a runner, free VRAM reflects current load, so admission decisions aren't reproducible across attempts. I don't have a clean answer, just noting it as a wrinkle if placement starts flapping.

Happy to review if useful.

PR #2217, opened 2026-07-10
feat: sequence-parallel Ring Attention for MLX

### LINKED ISSUE #39: Ring Attention for coupling the data transfer with computation of attention block matrices
An idea from the community:

Screenshot 2024-07-19 at 02 23 49

abendrothj:
Related to/Closes #39 

# Sequence-parallel Ring Attention for MLX

Implements distributed Ring Attention prefill for verified MLX Llama and Qwen3 architectures, with replicated decode caches, capability-gated placement, per-node memory admission, failure-safe cleanup, dashboard selection, integration coverage, and reproducible benchmarks.

## What's included

- **`RingAttentionLayer`** (`src/exo/worker/engines/mlx/ring_attention.py`): wraps only the attention module (never a full decoder block), splits the sequence across ranks, and rotates KV blocks around the ring while accumulating attention with a numerically stable online-softmax merge.
- **Overlapped communication**: dedicated CPU send/receive streams (MLX distributed send/recv are CPU ops; unified memory lets them consume Metal-produced KV without copies). Receives are posted *before* the current block's attention is scheduled, and the lazy attention recurrence is explicitly `mx.async_eval`-ed to create the computation window the transfer overlaps with. Transport peer (always the ring neighbor) is correctly separated from KV origin rank.
- **Placement**: `Sharding.Ring` is gated on `MlxRing` transport, ≥2 nodes, model-card `supports_ring`, and per-node admission of the *full* model size (weights are replicated; only prefill sequence is sharded).
- **Model cards**: ring support declared for verified Llama 3.x cards; `supports_ring` defaults off for everything else.
- **Tests**: 33 unit/integration tests including scheduling-order regression (communication posted before compute is waited on), online-merge equivalence vs full attention, and real 2-rank *and* 3-rank distributed Metal tests over the MLX ring backend.
- **Docs/benchmarks**: README instructions (`MLX_METAL_FAST_SYNCH=1`), `exo-bench --sharding ring` support.

## Benchmarks

Collected with `exo-bench` against a live 2-node exo cluster running this branch (`--warmup 1 --repeat 3`, prefix caching disabled, `MLX_METAL_FAST_SYNCH=1 uv run exo --no-batch` on both nodes).

**Environment**: both nodes on a single MacBook Pro (M4 Pro, 14-core, 24 GB, macOS 26.5.2, MLX 0.32.0.dev20260709). Model: `mlx-community/Llama-3.2-1B-Instruct-4bit`.

> ⚠️ Because both ranks share one GPU, these numbers demonstrate *functional correctness and stability*, not multi-device scaling: ring decode replicates the full model on every rank (2× total GPU work on one chip), and prefill compute contends for the same Metal queue. Multi-machine numbers will differ qualitatively; treat this as a smoke-test baseline, and `prompt_tps` as the primary ring prefill metric.

### Ring (sequence-parallel prefill), 2 nodes

| pp | tg | prompt_tps (3 runs) | gen_tps (mean) | peak mem/node |
|---|---|---|---|---|
| 2048 | 64 | **2137 / 2114 / 2069** (mean 2114) | 125 | 1.12 GB |
| 4096 | 64 | **1857 / 1849 / 1712** (mean 1806) | 104 | 1.55 GB |
| 8192 | 64 | **1283 / 1281 / 1212** (mean 1258) | 58 | 2.36 GB |

### Pipeline baseline (same nodes, transport, model, params)

| pp | tg | prompt_tps (3 runs) | gen_tps (mean) | peak mem/node |
|---|---|---|---|---|
| 2048 | 64 | 240¹ / 2012 / 1993 (median 1993) | 165 | 1.24 GB |
| 4096 | 64 | 1998 / 2005 / 1993 (median 1998) | 159 | 1.56 GB |
| 8192 | 64 | 1990 / 2008 / 719² (median 1990) | 138 | 1.88 GB |

¹ cold first iteration (kernel compilation) · ² transient slowdown, likely thermal/contention on the shared machine

**Reading the data**: ring prefill matches or beats pipeline at 2048 on this setup and falls behind as sequence length grows — expected on a shared GPU, where ring's replicated per-rank compute and KV rotation contend for the same chip while pipeline splits layers across ranks. The value proposition of ring (splitting long-context prefill *compute* across machines) can only show up on real multi-node hardware; these runs verify the mechanism works end-to-end through placement → distributed prefill → correct generations → clean instance teardown. Generated text was coherent and consistent across all runs and placements.

Reproduce with:

```bash
uv run bench/exo_bench.py --model Llama-3.2-1B-Instruct-4bit \
  --pp 2048,4096,8192 --tg 64,64,64 --min-nodes 2 --max-nodes 2 \
  --instance-meta ring --sharding ring --warmup 1 --repeat 3
# baseline: same command with --sharding pipeline
```

## Known issues / follow-ups

1. ~~**Continuous batching is not supported**~~ **Fixed in 0a0a6c4f**: ring decode now forwards to the wrapped attention module instead of re-implementing it, so batched decode (`BatchKVCache`) works and `--no-batch` is no longer required. Ring prefill was already batch-compatible (the batch engine prefills each request into a per-request `KVCache` before merging). Verified against a live batching-enabled 2-node cluster — see PR comment for details. The `KVCache`-only requirement still applies to the prefill path itself, which is what the batch engine hands it anyway.
2. **Intermittent hang at long sequence length (same-host repro)**: with pp=16384 on 2 ranks, two consecutive iterations completed normally (~35 s each) and the third deadlocked. Stack samples: rank 0 blocked in `mx.eval` → `Event::wait` on a Metal shared event; rank 1 blocked inside `mx.async_eval` scheduling (`eval_impl` condition wait); the MLX ring `SocketThread` idle-spinning with all TCP queues drained — i.e., a cross-rank circular wait, not a network stall. This was observed only on the shared-GPU same-host setup and did not reproduce at ≤8192 (9/9 runs clean) or in the 2- and 3-rank distributed test suite; it needs investigation on genuine multi-node hardware before ring placements are exposed beyond experimental use.

## Testing

- `uv run ruff check` — clean
- `uv run pytest` — 529 passed, 3 skipped (includes 2-rank and 3-rank distributed Metal ring tests, plus BatchKVCache decode passthrough regression tests)
- Live 2-node cluster: 18 successful benchmark completions across ring and pipeline placements with coherent generations and clean instance lifecycle
- Live 2-node cluster with batching enabled (`BatchGenerator`): concurrent requests on a Ring instance batch and answer correctly; 3570-token ring prefill output is identical to the pipeline baseline at temperature 0

### DIFF
--- a/README.md
+++ b/README.md
@@ -538,6 +538,13 @@ The `exo-bench` tool measures model prefill and token generation speed across di
 - Nodes should be running with `uv run exo` before benchmarking
 - The tool uses the `/bench/chat/completions` endpoint
 
+For Ring Attention on Metal, start every node with MLX's low-latency CPU/GPU
+synchronization enabled:
+
+```bash
+MLX_METAL_FAST_SYNCH=1 uv run exo
+```
+
 **Basic usage:**
 
 ```bash
@@ -554,7 +561,7 @@ uv run bench/exo_bench.py \
 - `--tg`: Generation lengths (comma-separated integers)
 - `--max-nodes`: Limit placements to N nodes (default: 4)
 - `--instance-meta`: Filter by `ring`, `jaccl`, or `both` (default: both)
-- `--sharding`: Filter by `pipeline`, `tensor`, or `both` (default: both)
+- `--sharding`: Filter by `pipeline`, `tensor`, `ring`, or `both` (default: both; `both` retains the pipeline/tensor comparison)
 - `--repeat`: Number of repetitions per configuration (default: 1)
 - `--warmup`: Warmup runs per placement (default: 0)
 - `--json-out`: Output file for results (default: bench/results.json)
@@ -571,6 +578,29 @@ uv run bench/exo_bench.py \
   --repeat 3 \
   --json-out my-results.json
 ```
+
+To compare Ring Attention against replicated pipeline prefill on the same
+Ring-compatible model and node count, run both placements with identical prompt
+lengths, generation lengths, repetitions, and warmups:
+
+```bash
+uv run bench/exo_bench.py \
+  --model Llama-3.2-1B-Instruct-4bit \
+  --pp 4096,16384,65536 --tg 128,128,128 \
+  --min-nodes 2 --max-nodes 2 --instance-meta ring \
+  --sharding ring --warmup 1 --repeat 3 \
+  --json-out bench/ring-attention.json
+
+uv run bench/exo_bench.py \
+  --model Llama-3.2-1B-Instruct-4bit \
+  --pp 4096,16384,65536 --tg 128,128,128 \
+  --min-nodes 2 --max-nodes 2 --instance-meta ring \
+  --sharding pipeline --warmup 1 --repeat 3 \
+  --json-out bench/replicated-pipeline.json
+```
+
+Use `prompt_tps` as the primary Ring prefill metric. Keep prefix caching disabled
+(the default), and record node hardware and network topology alongside results.
 
 The tool outputs performance metrics including prompt tokens per second (prompt_tps), generation tokens per second (generation_tps), and peak memory usage for each configuration.
 
--- a/dashboard/src/lib/components/ModelCard.svelte
+++ b/dashboard/src/lib/components/ModelCard.svelte
@@ -22,7 +22,7 @@
       }>;
     } | null;
     nodes?: Record<string, NodeInfo>;
-    sharding?: "Pipeline" | "Tensor";
+    sharding?: "Pipeline" | "Tensor" | "Ring";
     runtime?: "MlxRing" | "MlxJaccl";
     onLaunch?: () => void;
     tags?: string[];
--- a/dashboard/src/lib/stores/app.svelte.ts
+++ b/dashboard/src/lib/stores/app.svelte.ts
@@ -173,7 +173,7 @@ export interface ModelDownloadStatus {
 // Placement preview from the API
 export interface PlacementPreview {
   model_id: string;
-  sharding: "Pipeline" | "Tensor";
+  sharding: "Pipeline" | "Tensor" | "Ring";
   instance_meta: "MlxRing" | "MlxJaccl";
   instance: unknown | null;
   memory_delta_by_node: Record<string, number> | null;
--- a/dashboard/src/routes/+page.svelte
+++ b/dashboard/src/routes/+page.svelte
@@ -783,6 +783,7 @@
       quantization?: string;
       base_model?: string;
       capabilities?: string[];
+      supports_ring?: boolean;
     }>
   >([]);
   type ModelMemoryFitStatus =
@@ -885,14 +886,14 @@
     sendMessage(content, files, thinkingEnabled());
   }
 
-  let selectedSharding = $state<"Pipeline" | "Tensor">("Pipeline");
+  let selectedSharding = $state<"Pipeline" | "Tensor" | "Ring">("Pipeline");
   type InstanceMeta = "MlxRing" | "MlxJaccl";
 
   // Launch defaults persistence
   const LAUNCH_DEFAULTS_KEY = "exo-launch-defaults-v2";
   interface LaunchDefaults {
     modelId: string | null;
-    sharding: "Pipeline" | "Tensor";
+    sharding: "Pipeline" | "Tensor" | "Ring";
     instanceType: InstanceMeta;
     minNodes: number;
   }
@@ -932,7 +933,9 @@
     // Apply sharding and instance type unconditionally
     selectedSharding = defaults.sharding;
     selectedInstanceType =
-      defaults.instanceType === "MlxRing" ? "MlxRing" : "MlxJaccl";
+      defaults.sharding === "Ring" || defaults.instanceType === "MlxRing"
+        ? "MlxRing"
+        : "MlxJaccl";
 
     // Apply minNodes if valid (between 1 and maxNodes)
     if (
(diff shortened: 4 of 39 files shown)


abendrothj:
**Update: continuous batching now works on Ring instances (0a0a6c4f) — `--no-batch` no longer required.**

The incompatibility turned out to be narrower than the original error suggested. The batch engine already prefills each request individually into a per-request `KVCache` before merging it into the batch, so ring's sequence-parallel prefill was batch-compatible all along. The failure was decode-only: `_decode_step` re-implemented attention by hand and hard-gated on `isinstance(cache, KVCache)`, rejecting the merged `BatchKVCache`.

Since ring decode is fully replicated with no communication, the fix is to forward decode to the wrapped attention module (`return self.original_layer(x, mask=mask, cache=cache)`) — the same passthrough the wrapper already used for `world_size == 1`. Decode is now byte-for-byte the non-ring code path and inherits support for every cache type the model handles (batched, quantized, rotating). Cross-rank batch determinism rides on the batch engine's existing rank task-ordering agreement, the same mechanism pipeline instances use with batching today.

**Verification** (same-host 2-node M4 Pro cluster, `MLX_METAL_FAST_SYNCH=1`, *without* `--no-batch`, runners confirmed on `BatchGenerator`):

- 3 concurrent short requests on a Ring instance completed in overlapping ~0.6 s wall time with correct answers (previously: immediate `TypeError` on first request).
- A 3570-token prompt (multi-chunk ring prefill: 4 chunks at block size 512 × 2 ranks) followed by batched decode produced output **identical to a Pipeline placement of the same model at temperature 0** — evidence the ring-prefilled KV matches the standard path.
- New regression tests: wrapped vs unwrapped attention outputs are `array_equal` across multi-step decode with both `BatchKVCache` (left-padded, batch of 2) and plain `KVCache`.
- `basedpyright` 0 errors, `ruff` clean, full suite 529 passed / 3 skipped including the 2- and 3-rank distributed Metal tests.

Net diff: −40 lines of duplicated attention code, +2 tests.

abendrothj:
**Real two-node validation: heterogeneous Metal + CUDA ring over a physical network**

Follow-up to the same-host caveat in the benchmarks above — ring attention has now been validated on a genuine two-machine cluster:

- **Node A**: MacBook Pro M4 Pro, 24 GB (Metal, `MLX_METAL_FAST_SYNCH=1`)
- **Node B**: Debian 13 LXC with an RTX 3050 Laptop GPU, 4 GB VRAM (CUDA, `mlx-cuda-12` 0.32.0, NVIDIA driver 610.43.02), connected over a Tailscale-routed LAN (~6 ms RTT)

**Result**: `Ring / MlxRing / nodes=2` places, warms up, and serves correctly across Metal and CUDA ranks. 20/20 consecutive 4096-token ring prefills completed with zero hangs or errors — median 468 prompt tok/s (428–508), identical deterministic output on every iteration, coherent generations throughout. Decode runs at ~42 tok/s, bottlenecked by the replicated-decode design on the slower CUDA rank, as expected.

**8K/16K on this cluster OOM the 4 GB rank** (`cudaMallocAsync failed: out of memory`; measured VRAM peak 3608/4096 MiB at pp=8192 before the failure). So the intermittent 16K deadlock from the same-host runs could not be probed on this hardware — it remains open, pending a multi-node cluster with more headroom. Notably, the OOM is surfaced cleanly: instance deleted, HTTP 500 with the CUDA error message, no wedged runners.

**Findings from bringing up the CUDA rank** (all pre-existing conditions, not regressions from this PR — filing as candidate follow-ups):

1. **Ring memory admission only counts model size.** The 4 GB rank was happily admitted for a 16K-context placement it cannot serve; admission should model context-length working memory (KV + activations) per rank, not just weights.
2. **MLX CUDA needs `CUDA_HOME` at runtime.** The first distributed `send` in ring prefill JIT-compiles via NVRTC and dies with `Can not find locations of CUDA headers` unless `CUDA_HOME` points at CUDA headers (the pip `nvidia-cuda-runtime-cu12` package suffices: `CUDA_HOME=<venv>/site-packages/nvidia/cuda_runtime`). exo could set this automatically when the venv's nvidia packages are present.
3. **loguru `diagnose=True` can segfault while logging runner exceptions.** The original CUDA failure surfaced as an opaque `signal 11` because loguru crashed repr'ing MLX CUDA locals during traceback formatting (`_get_relevant_values` → `_format_value`). Running with `LOGURU_DIAGNOSE=NO` produced the real traceback immediately. The runner's exception handler should disable variable inspection so root causes aren't masked.
4. **Current `mlx-cuda-12` wheels require NVIDIA driver ≥ 570.** On driver 535 every kernel launch fails with `device kernel image is invalid` (no sm_86 SASS; PTX from CUDA 12.8/12.9 toolkits can't be JIT'd by a 12.2-era driver). Worth a note in the docs' Linux requirements.

The same-host bridging scripts (multicast discovery relay + reachability-port proxies) were only needed because both discovery and the reachability probe assume single-node-per-host/single-port topology; on this two-machine cluster the only helper needed was a discovery bridge across the VPN boundary (Tailscale doesn't forward multicast).

abendrothj:
I followed up on all four issues from the CUDA validation and they are now addressed.

Runner logging no longer risks masking CUDA exceptions, `CUDA_HOME` is detected automatically from the bundled NVIDIA packages, Ring placement now accounts for VRAM and the long-context working set, and the prefill communication path no longer depends on unfinished GPU work. I also added a progress watchdog so a future stall becomes a clean runner failure instead of an indefinitely stuck request.

The updated branch passes 543 tests, including the new watchdog, CUDA environment, accelerator admission, and memory-filtering coverage. On the same-host Metal stress setup, 12/12 consecutive 8,192-token prefills completed cleanly, followed by a clean 12,288-token run. The original 16K deadlock still could not be reproduced in an A/B run because the machine was hitting Metal command-buffer OOM pressure even with the old code.

I also tried the 16K case again on a clean M4 Pro + RTX 3050 setup. The nodes connected successfully, but the 4 GiB CUDA rank had only about 3.68 GiB available and the new admission check rejected the placement before starting a runner. That confirms the intended safety behavior, although a successful 16K run still needs a multi-node setup with more memory headroom.

PR #2234, opened 2026-07-23
feat: add macOS MLX setup script and expand Xcode prerequisites in README

chronicgiardia:
## Summary

Adds a self-diagnosing setup script for the MLX inference backend on macOS and expands the README to cover the Xcode/Metal Toolchain edge cases that trip up first-time contributors.

## Motivation

Setting up MLX requires the full **Xcode.app** (not just Command Line Tools), an accepted Xcode license, and the Metal Toolchain component — none of which are obvious from the current README. Missing any one of these produces cryptic errors like:

- `xcrun: error: unable to find utility "metal"`
- `Agreeing to the Xcode license requires admin privileges`
- `cannot execute tool 'metal' due to missing Metal Toolchain`
- `ModuleNotFoundError: No module named 'mlx'`

### `scripts/setup-mlx-macos.sh` (new)

A single script that checks each prerequisite in order and prints the exact remediation command when something is missing:

1. Confirms platform is macOS
2. Confirms `Xcode.app` is installed (CLT alone are insufficient)
3. Checks `xcode-select` points to `Xcode.app/Contents/Developer`
4. Verifies the Xcode license has been accepted
5. Auto-downloads the Metal Toolchain component via `xcodebuild -downloadComponent MetalToolchain` if absent
6. Runs `uv sync --extra mlx`
7. Imports `mlx` and prints the detected GPU device as a final sanity check

### `README.md`

- Expands the Xcode prerequisite with the three setup commands needed after installing Xcode.app
- Adds the setup script to the quick-start clone sequence
- Adds a **Troubleshooting MLX build failures** table covering the four most common errors

## Testing

All 27 MLX unit tests pass after running the script on a fresh setup:

```
27 passed, 184 deselected in 4.95s
```

Co-Authored-By: Oz <oz-agent@warp.dev>

PR #2245, opened 2026-07-30
docs: request the mlx extra in the documented setup commands

mlpy0:
Since #2087 moved mlx, mlx-lm, mlx-vlm and mflux out of `dependencies` into the `mlx` extra, the documented setup path never installs them, and `[tool.uv]` sets no default extras. `uv run exo` starts the API, then every runner crashes with `ModuleNotFoundError: No module named 'mlx'` (#2156).

Adds the extra to the commands documented as the way to set up:

- README macOS: `uv sync --extra mlx`
- README Linux: `uv sync --extra mlx-cpu`, with the mlx-cuda12 / mlx-cuda13 alternatives
- CONTRIBUTING.md quick start
- `just sync` / `just sync-clean`, which `just build-app` runs before pyinstaller — the spec aborts when the mlx Metal libraries are missing

Docs plus two justfile recipes; no dependency or lock changes. Backend choice stays explicit, since mlx-cpu / mlx-cuda12 / mlx-cuda13 / mlx-none are declared as conflicting extras.

#2234 overlaps on the macOS README block only, via a setup script that runs `uv sync --extra mlx`; the Linux block, CONTRIBUTING.md and the justfile are not covered there.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -29,6 +29,7 @@ To run EXO from source:
 git clone https://github.com/exo-explore/exo.git
 cd exo/dashboard
 npm install && npm run build && cd ..
+uv sync --extra mlx
 uv run exo
 ```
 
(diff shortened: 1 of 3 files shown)


rltakashige (approved):
Thank you for this!

PR #2294, opened 2026-09-02
fix(utils): stop retrying SIGKILL forever on an unkillable child process

zhast:
## Bug

`AsyncProcess._terminate_if_still_alive()` escalates join → SIGTERM ×10 → SIGKILL, and the SIGKILL phase is an unbounded loop:

```python
while True:
    process.kill()
    with move_on_after(_KILL_GRACE_SECONDS):
        await self.wait()
    if self.exitcode is not None or not process.is_alive():
        break
```

A child blocked in an uninterruptible kernel wait (`ps` state `U`) does not receive SIGKILL until that wait returns — which may be never. `is_alive()` then stays true and the loop has no exit, so whatever awaits the owner's shutdown never completes.

## How it was hit

On a 4-node pipeline deployment (v1.0.71), deleting two `MlxJaccl` instances of the same model back-to-back left every node's runner in state `U`. The supervisor logged

```
Runner process didn't respond to SIGTERM, killing
That took 1 attempts :( … That took 26 attempts :(     (one line every 5 s, indefinitely)
```

and because that loop ran inside the worker's `Shutdown` plan, the node's event loop stalled and the HTTP API stopped accepting connections until the daemons were restarted by hand. The equivalent code on `main` is async (the API would survive) but still never returns, so the owner's shutdown never finishes.

## Fix

Bound the SIGKILL phase to `_KILL_ATTEMPTS = 5`; on success log as before, otherwise log `CRITICAL` with the pid and return, leaving the child to the kernel. `run()` already wraps `process.close()` in `suppress(ValueError)`, so a still-alive child is tolerated on the way out.

## Test

`test_unkillable_child.py` drives the full escalation against a fake child that ignores every signal, with the grace periods shortened via `monkeypatch`, and asserts the call returns after exactly `_TERMINATE_ATTEMPTS` terminates and `_KILL_ATTEMPTS` kills. Existing `test_async_process.py` unchanged and passing; `ruff`, `ruff format --check`, `basedpyright` clean.

## Not addressed here

Why the runners became unkillable (most likely two JACCL/RDMA groups tearing down on the same devices at once) — I could not reproduce that on demand, only the supervisor's behaviour once it happens.

### DIFF
--- a/src/exo/utils/async_process.py
+++ b/src/exo/utils/async_process.py
@@ -34,6 +34,10 @@ _TERMINATE_GRACE_SECONDS = 5.0
 _TERMINATE_RETRY_GRACE_SECONDS = 2.0
 _TERMINATE_ATTEMPTS = 10
 _KILL_GRACE_SECONDS = 2.0
+# A child stuck in an uninterruptible kernel wait ignores SIGKILL until the wait
+# returns, which may be never. Give up after this many attempts rather than
+# holding the owner's shutdown open indefinitely.
+_KILL_ATTEMPTS = 5
 
 
 @final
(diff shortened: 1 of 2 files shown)


zhast:
Worth flagging for anyone hitting this on the current release: **v1.0.71 has the same bug, but not in this file.** `src/exo/utils/async_process.py` does not exist in that tree, so applying this patch to a v1.0.71 checkout is a no-op. The equivalent code is `RunnerSupervisor` in `src/exo/worker/runner/runner_supervisor.py`:

```python
# v1.0.71, src/exo/worker/runner/runner_supervisor.py
logger.critical("Runner process didn't respond to SIGTERM, killing")
j = 0
while self.runner_process.is_alive():
    j += 1
    self.runner_process.kill()
    self.runner_process.join(timeout=5)
    logger.warning(f"That took {j} attempts :(")
```

Same unbounded shape, and the `join(timeout=5)` is what produces the one-line-every-5s `That took N attempts :(` in the report above.

There is a second defect in the release version that the `main` fix does not have an analogue for. The escalation path uses **blocking** `join()` — three of them (`timeout=10` after the first `terminate()`, `timeout=2` in the `range(2, 11)` retry loop, and `timeout=5` in the kill loop) — even though the same file already uses `await to_thread.run_sync(self.runner_process.join, 5)` in two other places. So on v1.0.71 the unkillable child does not merely hold the owner's shutdown open; it pins the worker's event loop, which is why the HTTP API stopped accepting connections entirely rather than just stalling that one shutdown. That matches the note in the description that `main` is async and the API would survive there.

A v1.0.71 backport therefore needs both changes: bound the kill loop, and move those three joins onto `to_thread.run_sync` so the loop stays responsive while the child is stuck in state `U`.

I have been running exactly that locally since 2026-09-02 on the same 4-node pipeline deployment described above — `_KILL_ATTEMPTS = 5`, all five joins via `to_thread.run_sync`, and `close()` wrapped in `contextlib.suppress(ValueError)` since the orphaned child can no longer be closed. No recurrence since, including through a deliberate instance-delete sequence that previously reproduced the wedge.

Happy to open that as a separate PR against a release branch if one is maintained — it is a different file and a different diff, so it does not belong in this one.

PR #2297, opened 2026-09-08
feat: place pipeline shards by measured bandwidth and latency

### LINKED ISSUE #957: [MEDIUM] Better placement algorithm for pipeline parallelism using memory bandwidth + latency

abendrothj:
## Problem

Pipeline placement currently chooses the smallest feasible cycle and divides layers by available RAM. That can put too much work on a low-bandwidth node, ignore a faster larger cycle, or select a ring ordering with slower links.

## Changes

- Profile live memory bandwidth on each node and carry it into placement.
- Allocate layers in proportion to bandwidth while treating available RAM as a hard capacity.
- Estimate each candidate's token time from its actual capacity-adjusted layer allocation and the measured RTT of each ring hop.
- Compare fully profiled cycles across device counts, with leaf status, download progress, and RAM used only as tie-breakers.
- Preserve the existing smallest-cycle and download-aware behavior until profiling has produced a complete candidate.
- Split placement filtering, selection, and instance construction into focused helpers.

This replaces #1088 and incorporates the review feedback there.

## Stack

Depends on #2254 for measured per-link latency. Until #2254 merges, GitHub will include its commits in this PR's diff because both head branches are in a fork. The diff will narrow to the placement changes automatically after #2254 lands.

## Tests

- Added end-to-end coverage for measured-latency selection, bandwidth taking priority over download progress, comparison across cycle sizes, and capacity-adjusted scoring.
- `uv run basedpyright`
- `uv run ruff check`
- `uv run ruff format --check` on changed Python files
- `uv run pytest --ignore=src/exo/download/tests --ignore=rust/exo_rs/tests/test_python.py` — 425 passed, 3 skipped, 190 deselected

`nix fmt` was unavailable locally because Nix is not installed.

Fixes #957

### DIFF
--- a/src/exo/master/main.py
+++ b/src/exo/master/main.py
@@ -378,6 +378,7 @@ class Master:
                                 self.state.node_backends,
                                 download_status=self.state.downloads,
                                 node_rdma_ctl=self.state.node_rdma_ctl,
+                                node_bandwidth=self.state.node_bandwidth,
                             )
                             transition_events = get_transition_events(
                                 self.state.instances, placement, self.state.tasks
--- a/src/exo/master/placement.py
+++ b/src/exo/master/placement.py
@@ -4,6 +4,7 @@ from typing import Sequence
 
 from exo.master.placement_utils import (
     Cycle,
+    estimate_token_seconds,
     filter_cycles_by_memory,
     get_mlx_jaccl_coordinators,
     get_mlx_jaccl_devices_matrix,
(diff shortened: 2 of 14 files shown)


abendrothj:
Moving to draft: this branch still carries #2254's commit, so the diff here double-counts that work until #2254 lands. It also rewrites the placement selection and layer-allocation paths that #2252 (accelerator tiebreaker) and #2253 (manual `node_layers`) change, and the current revision does not preserve either behaviour. I will rebase onto whatever lands first and mark this ready once the diff is placement-only.

Issue #1908, opened 2026-04-16
Sequential prefill blocks active decode — request for chunked / mixed batching


mlpy0:
Prefill currently blocks decode for all active tasks whenever a new request arrives. For the multi-user case this is a real problem — unrelated users sharing a cluster stall each other's streams.

Where it happens in the code:

- `BatchGenerator.submit()` in `src/exo/worker/engines/mlx/generator/batch_generate.py` runs a task's full prefill before inserting it into the generation batch (around lines 202-212).
- `_start_task()` in `src/exo/worker/runner/llm_inference/batch_generator.py` awaits prefill completion before any token is produced for concurrent requests.
- `TODO.md` line 4 already flags this: *"Solve the problem of in continuous batching when a new prompt comes in, it will block decode of the current batch until the prefill is complete."*

The practical effect: if user A is midway through a long streaming response and user B sends a new 8k prompt, A's stream stalls for the duration of B's prefill. On a shared cluster with several users this shows up as visible stutter (or outright pauses) in inter-token latency for everyone currently streaming, caused by whoever happens to submit next. That is hard to explain to end users.

Single-user multi-agent is more tolerant because one user is orchestrating the whole thing, but multi-user is where the current scheduler behavior is clearly wrong.

Two directions worth discussing:

1. **Chunked prefill at the scheduler level** — break the new task's prefill into chunks (the pipeline prefill path already uses 4096-token chunks internally, see `pipeline_parallel_prefill` in `generate.py`) and step the decode batch between chunks.
2. **Mixed batches** — run one prefill chunk inside the same forward pass as the active decode batch (vLLM-style chunked prefill / SGLang-style mixed batching).

Either removes the hard blocking without changing the "one prefill at a time" invariant, so the existing prefill-side optimizations should still hold.

Happy to prototype (1) if there's agreement on the direction.

PR #2312, opened 2026-09-15
fix(mlx): interleave prefill with active decode

STRML:
Addresses #1908.

## Summary

- admit ordinary text prompts into mlx-lm continuous batching when another request is active
- run active decode first, then process one bounded prompt chunk per scheduler iteration
- serialize competing prefills so simultaneous arrivals cannot multiply the decode pause
- default to 256-token chunks, configurable with EXO_INTERLEAVED_PREFILL_STEP_SIZE
- preserve prefix-cache updates, hybrid SSM snapshots, prefill progress, and distributed cancellation
- retain the existing synchronous path for pipeline-parallel, vision, and remote-prefill requests

The first request on an idle engine still uses the high-throughput synchronous prefill path. Only newly arriving prompts compete with existing work. This implements the scheduler-level chunking direction proposed in #1908 without requiring mixed prefill/decode kernels.

## Self-review notes

- prefill_batch_size is deliberately 1, preserving the issue stated one-prefill-at-a-time invariant and making the chunk size a cluster-wide latency budget rather than a per-request budget
- interleaved scheduler chunks are 256 tokens, while hybrid-cache state snapshots retain a 4096-token cadence plus the final reusable state; this avoids a 16x snapshot-memory increase
- rotating caches are clamped before mlx-lm merges them into either path
- exact and partial prefix-cache metadata, progress callbacks, and cancellation agreement are carried across the asynchronous prompt lifecycle

## Automated testing

- ruff check .
- basedpyright
- pytest src -q: 470 passed, 3 skipped, 190 deselected
- the new regression is part of the normal CI set, uses a self-contained tiny tokenizer/model, and runs locally in about 2.4 seconds
- it verifies active decode continues, only one competing prefill advances, chunked prefill reaches decode, progress is monotonic, prefix-cache save and exact reuse work, and distributed cancellation removes queued prefill
- existing slow B=2 tests require the source-built EXO MLX fork; with the stock prebuilt MLX wheel they fail identically on untouched main and this branch

Nix is not installed on the local test machine, so nix fmt was not run. No Nix files changed.

## Scope and follow-ups

This patch targets single-node and tensor-parallel text generation. Pipeline-parallel and vision prefills need specialized incremental handling and remain on the existing synchronous path. Remote prefill retains its existing synchronous transfer/cache-ingestion lifecycle. Hardware tuning can adjust EXO_INTERLEAVED_PREFILL_STEP_SIZE without changing scheduler behavior.


STRML:
@rltakashige this is ready for review and implements the scheduler-level chunked-prefill direction from #1908. The regression now runs in the normal CI set and covers competing prefill serialization, active decode progress, prefix-cache reuse, and cancellation. GitHub is holding the fork workflow at action_required pending maintainer approval.

